PR #68, opened 2025-04-09
Running Single Cells Simulations using SONATA circuit 

darshanmandge:
Description:
- To provide the ability to run single-cell simulations using a SONATA circuit via bluecellulab in obi-one 
- Adding an example to generate sonata circuit files for these simulations.


james-isbister:
Hi @darshanmandge 

Could you remind us what needed to be done to wrap this up? I remember there were just a few small things

Perhaps @chr-pok can look at doing this if you're busy today?

darshanmandge:
There are a few things that need to be updated: the h5 file structure and the ability to create an h5 file automatically using a script. I wanted to discuss this with you yesterday evening. Let's discuss it today, along with other single-cell/pair-neuron issues.

PR #81, opened 2025-04-11
Initial morphology_metrics form and example. Code adapted from neuroagent code written by Nicolas Frank, Boris Bergsma, Jan Krepl

james-isbister:
See:
- examples/morphology_metrics_example.ipynb
- https://github.com/openbraininstitute/obi-one/blob/neurom_prototype/obi/modeling/morphology_metrics/morphology_metrics.py


jdcourcol:
@james-isbister I think the next step is to use entity core instead as we have support for morphologies + authentication + accounting

PR #86, opened 2025-04-14
1) Docker + CI, 2) Multiservice launch script, 3) Various fixes 4) Entity SDK integration + example begun, 5) Morphology metrics example begun, 6) multiple_value_parameters and coordinate_parameters fixes

james-isbister:


chr-pok (changes requested):
Installation and launch works nicely! Just noticed in the readme that the URL for viewing the endpoints should be `http://127.0.0.1:8100/docs`.

Also, I suppose the entitycore example is not ready yet, as I got an `EntitySDKError: Request error: [Errno 61] Connection refused`.

The morphology and BlutETL examples did run without error. Other than that, I haven't tested too much.

james-isbister:
Yes, still fixing some small things with the entitysdk example.

I've changed the url in the README.

chr-pok (approved):

PR #89, opened 2025-04-15
Make connectivity optional

GianlucaFicarelli:
connectivity is not installed in the Docker image, but it's added only to the local installation (make install).

Reason: the connectivity packages depend on pyflagsercount and bigrandomgraphs, that don't adhere to modern standards and cannot be installed cleanly.
Moreover, pyflagser doesn't provide a source dist on pypi (see https://pypi.org/project/pyflagser/#files), so it cannot be installed in docker on Mac, unless enforcing platform=linux/amd64, but this would cause the process in the container to run using qemu.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,9 @@ apt-get install -qyy \
     build-essential \
     ca-certificates \
     cmake \
+    libboost1.81-all-dev \
+    libhdf5-dev \
+    libopenmpi-dev \
     zlib1g-dev \
     git
 EOT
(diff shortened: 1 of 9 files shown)


GianlucaFicarelli:
Note that the exclusion from the Docker image is temporary, because the service would return an error if the connectivity functionalities are called.

To properly handle this, we should:
- completely exclude the connectivity endpoints from the API, when the packages aren't installed
- or install all the required dependencies

But this can be postponed after the first deployment.

danielaegassan:
This is great thanks.  Maybe it helps us figure out optional dependencies for us as well indeed.

PR #99, opened 2025-04-16
Use authentication with GitHub IDP for the AWS role which can publish a new container image in AWS ECR

DriesVerachtert:


DriesVerachtert:
@GianlucaFicarelli this one uses now github idp authentication so we don't need a secret anymore. Would it be ok to merge this?

I tested once with some different 'on: anything' rule to make sure it could push an image, but deleted that image in the repository afterwards.

GianlucaFicarelli (approved):

PR #106, opened 2025-04-16
Merge dev to main

james-isbister:


james-isbister:
Something strange happened with the commit history, which made the merge tricky. Dev and main had diverged but had copies of the same commit following their divergence.

I think everything is now resolved. I've spoke with @chr-pok and we can delete the dev branch as suggested once this PR is merged

GianlucaFicarelli (approved):

PR #114, opened 2025-04-25
Implementation of a NeuronSet (block) prototype

chr-pok:
- Abstract `NeuronSet` base class representing sets of neurons (i.e., neuron IDs) of a circuit that can be used by other blocks that require a neuron set as an input
- Specific implementations (sub-classes): `PredefinedNeuronSet` (wrapper for existing node sets), `CombinedNeuronSet` (combination of node set), `IDNeuronSet` (based on neuron IDs), `PropertyNeuronSet` (based on neuron properties)
- Example notebooks for how to use neurons sets and how to generate simulation campaigns based on them
- Important: In general, the validity of neuron set definitions is not checked during initialization, but when resolved within a specific circuit's node population (see examples)

Related issue: https://github.com/openbraininstitute/prod-circuit-simulation/issues/25


obi_one/modeling/circuit/neuron_sets.py
```diff
@@ -1,10 +1,235 @@
...
+            json.dump(node_sets, f)
+
+        return output_file
+
+
+class ExistingNeuronSet(NeuronSet):
+    """
+    Neuron set wrapper of an existing (named) node sets.
+    """
+    node_set: Annotated[str, Field(min_length=1)] | Annotated[list[Annotated[str, Field(min_length=1)]], Field(min_length=1)]
```
james-isbister, line 160: I'm a bit confused by this notation for defining the node_set parameter (and other paramaters in other classes in this file). Especially by the lists part. Gianluca told me that this annotation style is indeed preferred though for pydantic. Could you explain it to me here?

obi_one/modeling/simulation/simulations.py
```diff
@@ -77,11 +115,33 @@ def generate(self):
...
 
+        # Write SONATA node sets file (.json)
         os.makedirs(self.coordinate_output_root, exist_ok=True)
-        simulation_config_path = os.path.join(self.coordinate_output_root, f"simulation_config.json")
-        with open(simulation_config_path, 'w') as f:
+        c = self.initialize.circuit.sonata_circuit
+        for _name, _nset in self.neuron_sets.items():
+            # Resolve node set based on current coordinate circuit's default node population
+            # FIXME: Better handling of (default) node population in case there is more than one
+            nset_name, nset_expression = self._resolve_neuron_set(_nset, self.initialize.circuit, self.initialize.circuit.default_population_name)
```
james-isbister, line 128: Maybe nset_dictionary would be more explicit than nset_expression? I had to look into the code to understand what it was

obi_one/modeling/simulation/simulations.py
```diff
@@ -77,11 +115,33 @@ def generate(self):
...
-            self._sonata_config['reports'][recording_key] = recording.generate_config()
+            self._sonata_config["reports"][recording_key] = recording.generate_config()        
 
+        # Write SONATA node sets file (.json)
         os.makedirs(self.coordinate_output_root, exist_ok=True)
-        simulation_config_path = os.path.join(self.coordinate_output_root, f"simulation_config.json")
-        with open(simulation_config_path, 'w') as f:
+        c = self.initialize.circuit.sonata_circuit
+        for _name, _nset in self.neuron_sets.items():
+            # Resolve node set based on current coordinate circuit's default node population
```
james-isbister, line 126: Maybe for now we can assert that there is only one population? Just so we don't forget about this

obi_one/modeling/simulation/simulations.py
james-isbister, line 128: Maybe there is a more explicit name than resolve_neuron_set, like resolve_neuron_set_dictionary_if_new

obi_one/modeling/circuit/neuron_sets.py
```diff
@@ -1,10 +1,235 @@
...
+        else:  # File existing but no option chosen
+            assert False, f"Output file '{output_file}' already exists! Delete file or choose to append or overwrite."
+                
+        with open(output_file, "w") as f:
+            json.dump(node_sets, f)
+
+        return output_file
+
+
+class ExistingNeuronSet(NeuronSet):
```
james-isbister, line 156: "Existing" could possibly leave the user wondering whether this includes the NeuronSets they've already defined with Blocks already in the Simulation perhaps. Maybe PredefinedNeuronSet could be used.

james-isbister (commented):
Looks great! Just some tiny suggestions on possible names for a few classes, variables and functions

obi_one/modeling/circuit/neuron_sets.py
chr-pok: With this `Annotated[XXX, Field(min_length=1)]` notation you can enforce that a variable of a certain type XXX must have at least one element (because of `min_length=1`). This works for example for `str`, `list`, and `tuple` types, and you can also use it in a nested way as it is used here.
So this line means that `node_set` must be either a non-empty string or a non-empty list of non-empty strings. The validation will fail if you pass an empty string, and empty list, or a list of strings with (at least) one which is empty.

obi_one/modeling/simulation/simulations.py
chr-pok: Done

chr-pok: Renamed it to `_resolve_neuron_set_dictionary`. Will resolve now into a new name in any case for consistency (as discussed).

chr-pok:
Addressed all the suggestions. Last thing missing is allowing a `str` for existing node sets as discussed.

chr-pok:
@james-isbister We could also merge right now and create a new ticket for addressing the proper handing of existing node sets using strings separately.

james-isbister (approved):

james-isbister:
Looks good! I'll make the merge. If you could make the ticket you mentioned, that would be great!

PR #117, opened 2025-04-30
Add authentication and allow access to the auth token

GianlucaFicarelli:
- Read the authentication headers and optionally validate the token with keycloak (main logic reused from entitycore, we may decide to move it to an external package but it needs some adjustments to be really reusable)
- In particular, the class UserContext:
  - exposes the authentication token in `UserContext.token`, so it can be reused for calling entitycore or other services needing authentication
  - exposes `virtual_lab_id` and `project_id`, if the headers `virtual-lab-id` and `project-id` have been passed (they may be needed when calling entitycore)

Breaking changes:
- Move all the forms under /forms so they can be all authenticated (excluding /version and /health)
- Add authentication to all the /forms endpoints, unless APP_DISABLE_AUTH is set to false


GianlucaFicarelli:
> If I understand the breaking changes correctly, it is that the forms now all require authentication unless APP_DISABLE_AUTH is set to false, so any services calling these endpoints will not work as before?
> 
> Or is there something else that needs to be changed in obi-one?

Both the breaking changes require changes to obi-generative gui:
- add the auth header to each call to obi-one
- change any call from `/<model>` to `/forms/<model>`, but this might not be needed if the information is retrieved from openapi.json
- moreover, it could be avoided to call the `/forms` endpoint to get the list of endpoints, because this information can be obtained by openapi.json by filtering by tag=forms

james-isbister (approved):

PR #126, opened 2025-05-08
Base class fix for endpoint calling

james-isbister:
* get_class_recursive re-added to base.py. It only caused an error when calling Form endpoints from the GUI


chr-pok (approved):
Oh, good to add that back. Only now I realized that deserialization got broken without that.

james-isbister:
Yes, sorry about that!

PR #128, opened 2025-05-09
Further linting

james-isbister:


chr-pok (approved):
There are quite some funny things the linter is doing ;-)

Issue #122, opened 2025-05-05
SomaticStimuli should properly use a new NeuronSet


james-isbister:
**Description:**
- The different types of somatic `Stimulus` blocks used in simulation campaigns should properly use the newly implemented `NeuronSet` blocks
- In details, a somatic stimulus should be applied to all the neurons as specified by a `NeuronSet`

chr-pok:
This is being worked on in the obi-one branch [neuron_set_integration](https://github.com/openbraininstitute/obi-one/tree/neuron_set_integration).

chr-pok:
[PR #136](https://github.com/openbraininstitute/obi-one/pull/136) merged into main.

PR #136, opened 2025-05-09
Neuron set integration into stimulus

chr-pok:
- Integrated proper use of NeuronSets into (somatic) stimulus types
- Added working example of how to generate a simulation campaign that uses somatic stimuli based on timestamps and neuron sets
- Related issue: https://github.com/openbraininstitute/obi-one/issues/122

Notes:
- The name of a neuron set is given by the (unique) dict key in `neuron_sets={...}` when initializing a `SimulationsForm`
- The same is true for `timestamps={...}` and `stimuli={...}`
- When initializing a `SimulationsForm`, the respective names are passed to all objects that are part of these dicts.
- In this way it is possible to check later on (i.e., when generating a grid scan) that all the dependencies of a stimulus and other components (like timestamps, neuron sets, etc.) are actually part of the simulation!
- To be checked if this has impact in the GUI.


chr-pok:
@james-isbister we can discuss if this is a useful way to handle dependencies in the SimulationForm. If so, I will add it to recordings in a consistent way...

james-isbister:
Looks nice! check_sim_init and the "initialize_stimuli" validators look good

For me to check the gui, you'll need to merge main into this branch, as there was a fix earlier today

chr-pok:
> ...you'll need to merge main into this branch, as there was a fix earlier today

Done (luckily, without any conflicts) + automatic linting fixes applied

james-isbister:
I just checked the GUI and when selecting the neuron_set of a SomaticStimulus the entry is a text field rather than a dropdown for selecting a NeuronSet. 

I noticed that a neuron set of a SomaticStimulus is defined as having type: NeuronSetUnion | list[NeuronSetUnion]. It seems this was made previous to the work of the PR so it might have been me.

I think we discussed that parameters of type Block be Block | list[Block], but rather the multi-dimensional parameters should instead be the parameters of the NeuronSet i.e. random_seed: int | list[int].

If this makes sense, I think you would just need to change the definition of the SomaticStimulus parameter to neuron_set: NeuronSetUnion

obi_one/scientific/simulation/simulations.py
```diff
@@ -51,15 +51,28 @@ class Initialize(Block):
...
     @model_validator(mode="after")
-    def check_node_set(self) -> Self:
-        """Checks that given node set is member of the neuron_sets dict."""
-        assert self.initialize.node_set.name in self.neuron_sets, (
-            "Node set must be within neuron sets dictionary!"
-        )
-        assert self.initialize.node_set == self.neuron_sets[self.initialize.node_set.name], (
-            "Node set inconsistency!"
-        )
+    def initialize_timestamps(self) -> Self:
```
james-isbister, line 58: This seems like a good solution. In the long-run when we use such multi-level Forms for other use-cases, we might see if we can somehow abstract this, so that it doesn't need to be repeated for the different dictionaries. This is great for now though, and might be the best solution

obi_one/scientific/simulation/simulations.py
```diff
@@ -119,6 +123,12 @@ def generate(self):
             "ProbGABAAB_EMS": {"init_depleted": True, "minis_single_vesicle": True},
         }
 
+        # Generate stimulus input configs
+        self._sonata_config["inputs"] = {}
```
james-isbister, line 127: Just to check I understand correctly how the stimuli can be made before the neuron_sets: 

The names of neuron_sets in the end are always the names that are given as the dictionary key or the name of the predefined neuron set (which should match the name of predefined the node_set)?

obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -13,7 +13,9 @@
 class NeuronSet(Block, abc.ABC):
     """Base class representing a SONATA neuron set."""
```
james-isbister, line 14: Maybe a more detailed description of several lines could be added here to explain the general behaviour of NeuronSet (i.e. it should have the name the same as in the Simulations neuron_sets dictionary if that is general to other use cases, what it returns exactly etc.). See the final comment below

obi_one/scientific/simulation/simulations.py
chr-pok: Yes indeed! I agree for now I would like to keep it explicit here for better readability and testability. Later, this could perhaps be moved to the `Form` base class which just runs over all dict member variables and sets the `sim_init_name` whenever there is a `Block`.

obi_one/scientific/circuit/neuron_sets.py
james-isbister: Also the name "sim_init_name" is not totally clear. I understand that it is the name defined in the dictionary.

Could you maybe describe explicitly what it represents in this file, and perhaps also give it a different name as well, if there is maybe something better? Perhaps simulation_level_name?

Also I think it should be a private property with a setter used in simulations.py. Otherwise it appears in the serialization (I just checked), which is unnecessary because it is a copy of the key of the associated dictionary

obi_one/scientific/simulation/simulations.py
james-isbister: Thanks for the clarrification. Maybe you can add this in the neuron_sets.py as a comment somewhere so its documented

chr-pok: Good idea! I have added it to simulations.py though, since I think this is more specific to simulation campaigns than to actual neuron sets.

obi_one/scientific/circuit/neuron_sets.py
chr-pok: I added more description and renamed `sim_init_name` to `simulation_level_name`.

However, it cannot be private and it must be part of the serialization by design, otherwise there will be an error if you try to use a de-serialized grid scan object. Because at the moment for a given `node_set` of a simulation (i.e., simulation target) there is no other way to tell which neuron set from the `neuron_sets` dict it corresponds to. 

This could perhaps be addressed separately by introducing a proper way of dealing with references between objects when (de-)serializing a simulation campaign.

james-isbister (approved):

PR #138, opened 2025-05-12
Morphology locations added

MWolfR:
Implemented a family of blocks that generate locations on a morphology skeleton according to specification. Different blocks expose this concept at different levels of complexity / customizability.

Also implemented a Form that uses the block to generate plots of the morphology with locations indicated.


obi_one/scientific/morphology_locations/morphology_location_form.py
```diff
@@ -0,0 +1,91 @@
...
+    class Initialize(Block):
+        morphology: (
+            ReconstructionMorphologyFromID
+            | list[ReconstructionMorphologyFromID]
+            | Path
+            | list[Path]
+        ) = Field(title="Morphology", description="The morphology skeleton to place locations on")
+
+    initialize: Initialize
+    morph_locations: MorphologyLocationsBlock = Field(
```
james-isbister, line 36: We will need to define MorphologyLocationsUnion as a union of the different morphology location Blocks in scientific/unions/unions_morphology_locations.py and use that as the type here rather than MorphologyLocationsBlock. 

Unfortunately Pydantic and FastAPI don't really 'work' with polymorphism (i.e. using the parent class here) in the sense the subclass types are not explicitly made part of the schema.

Instead using a Union explicitly adds the different subclasses as options for the parameter, like in the stimulus dictionary here:
Screenshot 2025-05-12 at 14 00 47

That way, from the schema the GUI can offer the different subclass Blocks as options automatically, and the AI agent can also know the possible Block options. Also helps with deserilization from a json request body, otherwise the data get parsed into the parent class, and the subclass type and parameters are lost.

You can see an example in unions/unions_stimuli.py

obi_one/scientific/morphology_locations/morphology_location_block.py
```diff
@@ -0,0 +1,249 @@
...
+        return self._make_points(morphology)
+
+
+class RandomMorphologyLocations(MorphologyLocationsBlock):
+    """Completely random locations without constraint"""
+
+    def _make_points(self, morphology):
+        locs = generate_neurite_locations_on(
+            morphology,
+            1,
```
james-isbister, line 53: Maybe explicitly using the keywords would improve readibility and risk of bugs. Like: n_centers=1, ...

obi_one/scientific/morphology_locations/morphology_location_block.py
```diff
@@ -0,0 +1,249 @@
...
+    n_groups: int | list[int] = Field(
+        default=1, name="Number of groups", description="Number of groups of locations to generate"
+    )
+
+    def _make_points(self, morphology):
+        locs = generate_neurite_locations_on(
+            morphology,
+            1,
+            self.number_of_locations,
+            self.n_groups,
```
james-isbister, line 82: Should we add a validator that srcs_per_center <= n_per_center here and in other Blocks?

obi_one/scientific/morphology_locations/specified_morphology_locations.py
```diff
@@ -0,0 +1,351 @@
...
+    center_pd_mean,
+    center_pd_sd,
+    max_dist_from_center,
+    lst_section_types=None,
+    seed=None,
+):
+    """Generates a specified number of morphology locations according to complex specifications.
+    Locations are defined along morphology skeleton, i.e. they are 1d locations, not membrane locations.
+
+    Locations are 'clustered' around center locations on the morphology. If that is notr required, the
```
james-isbister, line 265: Typo "notr"

obi_one/scientific/morphology_locations/specified_morphology_locations.py
```diff
@@ -0,0 +1,351 @@
...
+    parameter 'max_dist_from_center' can be set to None and clustering is ignored.
```
james-isbister, line 266: In the case that max_dist_from_center=None and clustering is ignored, my understanding is that n_centers are created, but then locations associated with each centre are drawn uniformly from across the neuron, rather than from within max_dist_from_center?

obi_one/scientific/morphology_locations/morphology_location_form.py
james-isbister: This might also solve this warning I get when I run the example:
```
PydanticSerializationUnexpectedValue(Expected `BasicConnectivityPlots` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `CircuitExtractions` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `ConnectivityMatrixExtractions` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `FolderCompressions` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `MorphologyContainerizationsForm` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `MorphologyMetricsForm` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `SimulationsForm` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `SingleBlockGenerateTestForm` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `SingleBlockEntityTestForm` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `MultiBlockEntitySDKTestForm` - serialized value may not be as expected [input_value=MorphologyLocationsForm(t...n_groups_per_cluster=2)), input_type=MorphologyLocationsForm])
  PydanticSerializationUnexpectedValue(Expected `MorphologyLocationsBlock` - serialized value may not be as expected [input_value=ClusteredPathDistanceMorp... n_groups_per_cluster=2), input_type=ClusteredPathDistanceMorphologyLocations])
  return self.__pydantic_serializer__.to_python(
```

examples/morphology_locations_example.ipynb
james-isbister, line 1: Adding an additional commented out MorphologyLocation set that is more interesting than Random might be cool.

This one for example that I made when testing the example:

```
# morph_locations=obi.ClusteredPathDistanceMorphologyLocations(
                    #     number_of_locations=[100, 200],
                    #     section_types=(4, ),
                    #     n_clusters=10,
                    #     cluster_max_distance=20,
                    #     path_dist_mean=200,
                    #     path_dist_sd=80,
                    #     n_groups_per_cluster=2,
                    # )
```

james-isbister (commented):
The Blocks and key function: "generate_neurite_locations_on" and the functions it calls are very clearly documented such that I could understand how it was working!

I've just suggested some simple additions + changes

examples/morphology_locations_example.ipynb
james-isbister: Also the output path of the grid scan can be something like: "../../obi-output/morphology_locations" to be consistent with the other examples

obi_one/scientific/morphology_locations/morphology_location_block.py
MWolfR: No. It actually works with more src_per_center than n_per_center. Just some src ids will remain unused. There may be a use case for this.

james-isbister (dismissed):
Thanks!

james-isbister:
Looks good! 

For the test to pass @MWolfR  I think you just need to rerun `make install` to update uv.lock, and push a commit with it

james-isbister (approved):

Issue #30, opened 2025-04-01
Somatic spikes + voltage recording should properly use a new NeuronSet


james-isbister:
**Description:**
- The different types of `Recording` blocks used in simulation campaigns should properly use the newly implemented `NeuronSet` blocks
- In details, a recording should contain all the neurons as specified by a `NeuronSet`

**Note:**
Recording types have been restricted to `SomaVoltageRecording` only, other types may be added later if needed.

chr-pok:
[PR #145](https://github.com/openbraininstitute/obi-one/pull/145) merged into main.

PR #145, opened 2025-05-12
Neuron set integration into recording

chr-pok:
- Integrated proper use of NeuronSets into VoltageRecording
- Added a working example of how to generate a simulation campaign that uses a voltage recording based on a given neuron set
- Related issue: https://github.com/openbraininstitute/obi-one/issues/30

Notes:
- SpikeRecording has been removed; seems like spikes are always recorded by the simulator
- Recording has been made abstract


chr-pok:
Here is an example of the generated simulation config:
[simulation_config.json](https://github.com/user-attachments/files/20164128/simulation_config.json)
[node_sets.json](https://github.com/user-attachments/files/20164129/node_sets.json)

obi_one/scientific/simulation/recording.py
```diff
@@ -7,33 +10,76 @@
...
+    @abstractmethod
+    def _generate_config(self) -> dict:
+        pass
 
 
 class VoltageRecording(Recording):
-    recording_type: str = "voltage"
-    dt: float | list[float] = 0.1
+    neuron_set: NeuronSetUnion = Field(description="Neuron set to record from.")
+    section: Literal["soma", "axon", "dend", "apic", "all"] = Field(
```
james-isbister, line 58: It is not clear to me where exactly neurodamus makes recordings for section: "axon", "dend", "apic", "all". 

I had a look here: https://github.com/BlueBrain/neurodamus/blob/main/neurodamus/report.py

Perhaps we can ask Jim and company in Teams.

Does only a single location get recorded for "axon", "dend", "apic", for example, and does "all" record from one location in each of these three?

Given the ambiguity we might just want to have SomaVoltageRecording and AxonInitialVoltageRecording if they are both unambiguous. 

We could deal with the other types later in conjunction with MorphologyLocation

obi_one/scientific/simulation/recording.py
```diff
@@ -7,33 +10,76 @@
...
+            "variable_name": "v",
+            "unit": "mV",
+            "dt": self.dt,
+            "start_time": self.start_time,
+            "end_time": self.end_time,
+        }
+        return sonata_config
 
 
 class IntracellularLocationSetVoltageRecording(VoltageRecording):
```
james-isbister, line 78: Perhaps IntracellularLocationSetVoltageRecording and ExtraceullarLocationSetVoltageRecording can be removed to avoid confusion, until we make a proper integration with MorphologyLocation for the former

obi_one/scientific/simulation/recording.py
james-isbister: AxonInitialSegmentVoltageRecording would probably be better

obi_one/scientific/simulation/recording.py
chr-pok: Agreed, I removed them for now. Just putting the code snippets here so we can find them more easily in case we need them again later on:
~~~
from obi_one.scientific.unions.unions_extracellular_location_sets import (
    ExtracellularLocationSetUnion,
)
from obi_one.scientific.unions.unions_intracellular_location_sets import (
    IntracellularLocationSetUnion,
)

class IntracellularLocationSetVoltageRecording(VoltageRecording):
    intracellular_location_set: IntracellularLocationSetUnion

    def _generate_config(self) -> dict:
        sonata_config = {
            "type": "compartment",
            "sections": self.intracellular_location_set.section,
            "cells": self.intracellular_location_set.neuron_ids,
            "variable_name": "v",
            "dt": self.dt,
            "compartments": "center",
            "start_time": self.start_time,
            "end_time": self.end_time,
        }
        return sonata_config


class ExtraceullarLocationSetVoltageRecording(VoltageRecording):
    extracellular_location_set: ExtracellularLocationSetUnion
~~~

obi_one/scientific/simulation/recording.py
chr-pok: I was looking at the documentation here:
https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#reports

I think it is a bit more complex, since the actual recording location also depends on the parameter `type` which is set fixed to "compartment" here (which is the same as in the BlueCelluLab examples). According to the above doc, it means that each compartment will be reported separately in the report file for the specified section(s).

So I think restricting it to "soma" or "axon" only would introduce yet another level of complexity which may not be required. So if I understand correctly, "all" just means that all compartments in soma, axon, dendrite are recorded. But yes we can talk to Jim to confirm.

chr-pok: Ok, we now have `SomaVoltageRecording` only which should be more intuitive. Other types to come later...

james-isbister (approved):

PR #159, opened 2025-05-16
extract morphology_metrics out of the endpoint

jdcourcol:
- extract get_morphology_metrics out of the end point and put it in the scientific part
- isort import and fix ruff issues.


james-isbister (approved):
Looks good, thanks!

PR #172, opened 2025-05-20
Spike stimulus

romani79:
Base class for spike stimulus.
Class for poisson spike stimulus.
Notebook showing usage of spike stimulus.


james-isbister:
Looks good @romani79 

Maybe you could add the Poisson stimulus to the circuit_simulations_example.ipynb though? 

That uses the SSCx toy circuit which is available here:
https://zenodo.org/records/14358272

Maybe you could then turn your spike_stimulus_example.ipynb into hippocampus_cyclinder_example.ipynb and move it to a new directory projects/tiny_circuits ? Then you could use it as a basis for building a working example for CNS over the next month?

obi_one/scientific/simulation/stimulus.py
```diff
@@ -349,3 +356,85 @@ def _generate_config(self) -> dict:
...
+        spike_df = pd.DataFrame(np.array([time_list, gid_list]).T, columns=['t', 'gid'])
+        spike_df = spike_df.astype({'t': float, 'gid': int})
+        spike_df.sort_values(by=['t', 'gid'], inplace=True)  # Sort by time
+        with h5py.File(spike_file, 'w') as f:
+            pop = f.create_group(f"/spikes/{popul_name}")
+            ts = pop.create_dataset("timestamps", data=spike_df['t'].values, dtype=np.float64)
+            nodes = pop.create_dataset("node_ids", data=spike_df['gid'].values, dtype=np.uint64)
+            ts.attrs['units'] = 'ms'
+
+class PoissonSpikeStimulus(SpikeStimulus):
```
james-isbister, line 418: Perhaps we can find better names for SpikeStimulus (and PoissonSpikeStimulus), which better capture what synapse_replay does and will also avoid ambiguity with future stimulus types where we activate single synapses individually.

Perhaps EfferentSpikeStimulus and PoissonEfferentSpikeStimulus better capture spike replay which is a spike initiated at the start of the axon, which propagates with a delay to the efferent synapses.

james-isbister: Also am I right in thinking that the spikes of the stimulated neuron that are instead generated by recurrent activity are also propagated in this form of spike replay?

obi_one/scientific/simulation/stimulus.py
```diff
@@ -349,3 +356,85 @@ def _generate_config(self) -> dict:
...
+    _module: str = "synapse_replay"
+    _input_type: str = "spikes"
+    frequency: float  # Hz
```
james-isbister, line 421: This can be something like:

`
float | list[float] = Field(
        default=0.0, title="Frequency", description="[Add a description of the parameter here]"
    )
`

The `float | list[float]` allows the user to specify multiple values which are then picked up as a dimension of the grid scan

obi_one/scientific/simulation/stimulus.py
```diff
@@ -349,3 +356,85 @@ def _generate_config(self) -> dict:
 
 class SynchronousSingleSpikeStimulus(Stimulus):
     spike_probability: float | list[float]
+
+
+class SpikeStimulus(Stimulus):
+    _module: str = "synapse_replay"
+    _input_type: str = "spikes"
+    duration: float
```
chr-pok, line 364: Should accept lists as well:
```suggestion
    duration: float | list[float]
```

obi_one/scientific/simulation/stimulus.py
```diff
@@ -349,3 +356,85 @@ def _generate_config(self) -> dict:
...
+    gid_spike_map: dict = {}
+    spike_file: Path | None = None
```
chr-pok, line 366: I think this variable could be made private, since it is only set and used internally. Also, in this way, it won't be part of the (de-)serialization. (Will require changing all occurrences!)
```suggestion
    _spike_file: Path | None = None
```

obi_one/scientific/simulation/stimulus.py
```diff
@@ -349,3 +356,85 @@ def _generate_config(self) -> dict:
...
+class PoissonSpikeStimulus(SpikeStimulus):
+    _module: str = "synapse_replay"
+    _input_type: str = "spikes"
+    frequency: float  # Hz
+
+    def generate_spikes(self, circuit, population, spike_file_path):
+        gids = self.neuron_set.get_neuron_ids(circuit, population)
+        gid_spike_map = {}
+        start_time = self.timestamps.timestamps()[0]
+        end_time = start_time + self.duration
```
chr-pok, line 427: At the moment, we are only using the first timestamp here (which was just a temporary solution). How about turning this into a loop iterating over all timestamps and using each of them as start of a new Poisson spike strain with the given duration?
In this way it would be possible to generate, e.g., a series of Poisson stimuli with duration XX ms every YY ms etc. It would have to be checked that the intervals are not overlapping though.
```suggestion
        timestamps = self.timestamps.timestamps()
        for t_idx, start_time in enumerate(timestamps):
            end_time = start_time + self.duration
            if t_idx < len(timestamps) - 1:
                # Check that interval not overlapping with next stimulus onset
                assert end_time < timestamps[t_idx + 1], "Stimulus time intervals overlap!"
            ...
```

obi_one/scientific/simulation/stimulus.py
```diff
@@ -349,3 +356,85 @@ def _generate_config(self) -> dict:
...
+            ts = pop.create_dataset("timestamps", data=spike_df['t'].values, dtype=np.float64)
+            nodes = pop.create_dataset("node_ids", data=spike_df['gid'].values, dtype=np.uint64)
+            ts.attrs['units'] = 'ms'
+
+class PoissonSpikeStimulus(SpikeStimulus):
+    _module: str = "synapse_replay"
+    _input_type: str = "spikes"
+    frequency: float  # Hz
+
+    def generate_spikes(self, circuit, population, spike_file_path):
```
chr-pok, line 423: I think for reproducibility this function should set the seed of the numpy random generator before generating any random spikes, i.e., `np.random.seed(self.random_seed)`.

`random_seed` should be added as an (optional) member variable to the class (which could be even a list for generating multiple instances with different seeds) in the same way as in the `NeuronSet` class:
`random_seed: int | list[int] = 0`

chr-pok (commented):
Overall looks good to me! Just a few ideas and suggestions for improvement.

obi_one/scientific/simulation/stimulus.py
```diff
@@ -349,3 +356,85 @@ def _generate_config(self) -> dict:
...
+        gids = self.neuron_set.get_neuron_ids(circuit, population)
+        gid_spike_map = {}
+        start_time = self.timestamps.timestamps()[0]
+        end_time = start_time + self.duration
+        for gid in gids:
```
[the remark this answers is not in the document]
james-isbister: This could be done in a follow up pull request so we can merge this initial work sooner

obi_one/scientific/simulation/stimulus.py
james-isbister: This could also be done in a follow up pull request rather than this one

examples/circuit_simulations_example.ipynb
```diff
@@ -97,6 +99,11 @@
     "stim_times = obi.RegularTimestamps(start_time=0.0, number_of_repetitions=3, interval=1000.0)  # in ms!!\n",
     "current_stimulus = obi.ConstantCurrentClampSomaticStimulus(timestamps=stim_times, duration=5.0, neuron_set=stim_neuron_set, amplitude=[0.2, 0.5])\n",
     "sync_current_stimulus = obi.ConstantCurrentClampSomaticStimulus(timestamps=stim_times, duration=100.0, neuron_set=stim_neuron_set, amplitude=0.1)\n",
+    "poisson_input = obi.PoissonSpikeStimulus(timestamps=stim_times,\n",
+    "                                                                        duration=sim_duration,\n",
+    "                                                                        frequency=stim_frequency,\n",
+    "                                                                        neuron_set=stim_neuron_set,\n",
```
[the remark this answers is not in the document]
romani79: done

obi_one/scientific/simulation/stimulus.py
romani79: done

obi_one/scientific/simulation/stimulus.py
```diff
@@ -349,3 +356,85 @@ def _generate_config(self) -> dict:
 
 class SynchronousSingleSpikeStimulus(Stimulus):
     spike_probability: float | list[float]
+
+
+class SpikeStimulus(Stimulus):
+    _module: str = "synapse_replay"
+    _input_type: str = "spikes"
+    duration: float
+    gid_spike_map: dict = {}
```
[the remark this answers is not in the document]
romani79: done

obi_one/scientific/simulation/stimulus.py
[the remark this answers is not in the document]
romani79: I agree.

obi_one/scientific/simulation/stimulus.py
romani79: I agree.

romani79:
I integrated most of the feedback with the exception of deleting SynchronousSingleSpikeStimulus for which I would wait the confirmation from @james-isbister

projects/tiny_circuits/.ipynb_checkpoints/hippocampus_cylinder_example-checkpoint.ipynb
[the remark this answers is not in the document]
romani79: Removed.

obi_one/scientific/simulation/stimulus.py
```diff
@@ -349,3 +356,85 @@ def _generate_config(self) -> dict:
 
 class SynchronousSingleSpikeStimulus(Stimulus):
```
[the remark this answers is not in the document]
james-isbister: Yes, this can be removed! @romani79

james-isbister:
Thanks for the changes @romani79 

The error when recreating a campaign is the expected behaviour at the moment (this may change). For now you just need to delete the simulation campaign folder to re-run

I think once you delete the SynchronousSingleSpikeStimulus and solve the merge conflict in the circuit_simulations_example we can probably merge @chr-pok ?

chr-pok:
> Thanks for the changes @romani79
> 
> The error when recreating a campaign is the expected behaviour at the moment (this may change). For now you just need to delete the simulation campaign folder to re-run
> 
> I think once you delete the SynchronousSingleSpikeStimulus and solve the merge conflict in the circuit_simulations_example we can probably merge @chr-pok ?

Ran a few quick tests, looks very good to me! I could not reproduce the error about existing folders. The spike file is correctly written with stimulus onsets as given by the timestamps.

Only the `random_seed` did not seem to have any effect, not sure why. After this is fixed, and after rebasing to main, I think we are ready to merge. I can look into it on Monday...

james-isbister (approved):

PR #173, opened 2025-05-20
1. Linting of core/serialization.py. 2. ZERO_INDEX and NAME_EQUALS_VALUE coord directories working. 3. Several automatic linting fixes in 4 additional files

james-isbister:
* core/serialization.py all linting fixes
* ZERO_INDEX and NAME_EQUALS_VALUE coord directories working
* Remaining automatic linting in simulations.py, circuit_extraction.py, morphology_metrics.py

### DIFF
--- a/obi_one/core/serialization.py
+++ b/obi_one/core/serialization.py
@@ -24,5 +24,6 @@ def deserialize_obi_object_from_json_data(json_dict: dict) -> obi.OBIBaseModel:
 
 
 def deserialize_obi_object_from_json_file(json_path: Path) -> obi.OBIBaseModel:
-    json_dict = json.load(json_path)
+    with Path.open(json_path) as file:
+        json_dict = json.load(file)
     return deserialize_obi_object_from_json_data(json_dict)
--- a/obi_one/core/serialization.py
+++ b/obi_one/core/serialization.py
@@ -1,6 +1,10 @@
+import json
+from pathlib import Path
 from typing import Any
 
 from entitysdk.models.entity import Entity
+
+import obi_one as obi
 
 
 def entity_encoder(obj: Any) -> dict[str, str]:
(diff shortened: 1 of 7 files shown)


obi_one/core/serialization.py
```diff
@@ -10,20 +14,15 @@ def entity_encoder(obj: Any) -> dict[str, str]:
...
+    obi_object = getattr(obi, json_dict["type"]).model_validate(json_dict)
     return obi_object
 
 
-def deserialize_obi_object_from_json_file(json_path):
-    with open(json_path) as file:
-        data = json.load(file)
-        return deserialize_obi_object_from_json_data(data)
+def deserialize_obi_object_from_json_file(json_path: Path) -> obi.OBIBaseModel:
+    json_dict = json.load(json_path)
```
chr-pok, line 27: Deserialization is broken, I think you need to add back the `with open ...` code:
```suggestion
    with open(json_path) as file:
        json_dict = json.load(file)
```

chr-pok (commented):
ZERO_INDEX and VALUE options are working fine (again), thanks for fixing it!

However, I noticed that de-serialization is broken now.

Also, just wondering how to actually access the full paths of the individual instances of the grid scan? I tried `grid_scan.coordinate_instances()[0].coordinate_output_root` but got `PosixPath('.')`, not sure if this is intended?

james-isbister:
@chr-pok The deserialization is fixed now. Thanks for spotting that!

The question about "grid_scan.coordinate_instances()[0].coordinate_output_root showing an empty path" is slightly more complex, and is the expected basic prototype behaviour at the moment. Now seems a good time to review that though.

Particularly, the three functions: `multiple_value_parameters`, `coordinate_parameters` & `coordinate_instances` get run from scratch whenever they are called. This means that the coordinate_instances actually get remade when coordinate_instances() is called. That is totally new instances of "Simulation" are created. Moreover, it is only when `execute` is called that the each coordinate_output_root is set and created.

This is related to the behaviour we want when someone runs execute a second time on the same GridScan object.

Some suggestions for quick improvements could be:
- Set 'coordinate_output_root' when `coordinate_instances()` is called
- Never re-run `multiple_value_parameters`, `coordinate_parameters` & `coordinate_instances` fully if execute() has already successfully completed.

I think it might be useful to make a ticket for this and discuss the desired functionality in person, and then I can work on this in a separate PR if you agree?

chr-pok:
> I think it might be useful to make a ticket for this and discuss the desired functionality in person, and then I can work on this in a separate PR if you agree?

Sure, sounds good! Would just be useful to have a quick way to get the actual paths of all instances after `execute` has been called.

james-isbister:
Thanks, I made a ticket which I'll try to work on asap:
https://github.com/openbraininstitute/obi-one/issues/176

Issue #168, opened 2025-05-16
Volumetric NeuronSet Blocks


james-isbister:

MWolfR:
Implemented in https://github.com/openbraininstitute/obi-one/tree/volumetric_neuron_set

chr-pok:
> I implemented Volumetric neuron sets in this branch https://github.com/openbraininstitute/obi-one/tree/volumetric_neuron_set

Nice!

> Please provide feedback. Also, note how I implemented filtering by neuron properties, such as mtype, layer, or whatever. Does that implementation make sense? If so, let's use it consistently.

- I like the way these filters are implemented! I think it will better fit into the GUI than my current implementation of using a dict with property/value pairs so I think we could use filters consistently like that. Just one thing I tried was to keep the symbolic notation - whenever possible - (i.e., keeping `"layer": 1` instead of `"node_id": [<list-of-layer1-IDs>]`) which I find is way more intuitive so we will have to see if this still fits for all different use cases.
- Other than that, I think `VolumetricNeuronSetBlock` should be derived from `NeuronSet` which has quite some functionality (e.g., for adding the node set to a circuit etc.) and only needs to implement the `_get_expression()` method.
- On the other hand, I think it should not be derived from `abc.ABC` which is only for abstract classes if I am not mistaken.
- I also noticed that the node population is hard-coded as "S1nonbarrel_neurons" in L65 in the code.
- I think we talked about providing a radius as well instead of an absolute number of neurons. Could this be added as well?

MWolfR:
I fixed what @chr-pok complained about and integrated it with the existing NeuronSet blocks better.
PR created!

danielaegassan:
Is this one final?  If yes, I could try to use it as a template for the simplex one.

james-isbister:
I'll need to look into this @danielaegassan. See my comments on the PR: https://github.com/openbraininstitute/obi-one/pull/179

Hopefully I'll have something soon

PR #179, opened 2025-05-21
Volumetric neuron set

MWolfR:
Now compatible with existing NeuronSets.


obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -269,43 +304,159 @@ def _get_expression(self, circuit, population):
 
 class PropertyNeuronSet(NeuronSet):
     """Neuron set definition based on neuron properties, optionally combined with (named) node sets."""
-
-    property_specs: (
-        Annotated[dict, Field(min_length=1)]
-        | Annotated[list[Annotated[dict, Field(min_length=1)]], Field(min_length=1)]
+    property_filters: tuple[NeuronPropertyFilterBlock, ...] | list[tuple[NeuronPropertyFilterBlock, ...]] = Field(
```
james-isbister, line 307: Currently to fit with the Form logic it would be better if the parameters and logic of NeuronPropertyFilterBlock belong to PropertyNeuronSet directly, and we don't create another Block for two reasons:

1. We would need to add a dictionary for the NeuronPropertyFilterBlocks at the root of the SimulationsForms like for stimuli, timestamps etc. for the scan object to look for list parameters / dimensions for the parameter scan.

2. We should never have a parameter of type Block | list[Block], because if someone uses [BlockA, BlockB] and Block B has a list parameter with multiple values, then we created nested dimensionality again, which we don't support. I should better document this.

Maybe property_filters can be of type: dict[str, tuple[str, ...]] | list[dict[str, tuple[str, ...]] 

I may have to update the multi-dimensional scan logic, but I think this representation would be reasonable to build upon

obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -5,11 +5,46 @@
...
 
 
+class NeuronPropertyFilterBlock(Block, abc.ABC):
+
+    property_name: str | list[str] = Field(
+        name="Property name",
+        description="Property name"
+    )
+
+    property_values : tuple[str, ...] | list[tuple[str, ...]] = Field(
```
james-isbister, line 21: The property values are always strings?

james-isbister:
I'll take a look into this to try turning NeuronPropertyFilterBlock into an OBIBaseModel and to check it works correctly with the multi-dimensionality stuff if that sounds good @MWolfR? @chr-pok has assigned me to the ticket https://github.com/openbraininstitute/obi-one/issues/168

obi_one/scientific/circuit/neuron_sets.py
james-isbister: I think this is maybe the simplest option @MWolfR 
`property_filters: dict[str, tuple[str, ...]] | list[dict[str, tuple[str, ...]]`

On the frontend side there is not much way of determining whether an object is a Block or just an Object (the typing gets removed in FastAPI, we're already adding the type strings manually).

As there's not much difference between this dict representation and what a Filter object would do, I think its maybe better to use the former, to avoid the small chance that a the Filter object gets interpreted by some frontend code as a Block.

I think this filter dictionary should only ever be used in the context of a NeuronSet, so the front end can just have a check akin to:
if BlockType in NeuroSetUnion & variable_name == "property_filter":
  > Display property filter selector

Let me know if you agree, and I can make the change?

james-isbister:
Now added to the circuit_simulation_example and added a string representation so that the folder names are reasonable.

Also in the neuron_sets_example, this produces the exact same neuron_sets as before (I compared the ids), so the functionality is preserved

james-isbister:
By the way @chr-pok, the final block in the neuron_sets_example notebook "Writing a NeuronSet to a SONATA node set file" is broken.

https://github.com/openbraininstitute/obi-one/blob/volumetric_neuron_set/examples/neuron_set_example.ipynb

I think this might not be related to Michael and my changes

obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -5,9 +5,67 @@
...
+import pandas
 from pydantic import Field, model_validator
 
+from obi_one.core.base import OBIBaseModel
 from obi_one.core.block import Block
+from obi_one.scientific.circuit.circuit import Circuit
+
+
+class NeuronPropertyFilter(OBIBaseModel, abc.ABC):
+    filter_dict: dict[str, list[str]] = Field(
```
chr-pok, line 17: We should not restrict to list of strings. E.g., in some older circuit releases, layers are still encoded as integers, not strings.
```suggestion
    filter_dict: dict[str, list] = Field(
```

obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -5,9 +5,67 @@
...
+    def test_validity(self, circuit, node_population: str) -> None:
+        circuit_prop_names = circuit.sonata_circuit.nodes[node_population].property_names
+        filter_keys = list(self.filter_dict.keys())
+
+        assert all(_prop in circuit_prop_names for _prop in self.filter_keys), (
+            f"Invalid neuron properties! Available properties: {prop_names}"
+        )
+
+    def __repr__(self) -> str:
+        """Return a string representation of the NamedPath object."""
```
chr-pok, line 59:
```suggestion
        """Return a string representation of the NeuronPropertyFilter object."""
```

obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -255,57 +326,168 @@ class IDNeuronSet(NeuronSet):
...
                 f"Node set '{_nset}' not found in circuit '{circuit}'!"
             )
 
-    def _get_expression(self, circuit, population):
+    def _get_resolved_expression(self, circuit: Circuit, population: str) -> dict:
+        """A helper function used to make subclasses work."""
+        c = circuit.sonata_circuit
+        node_ids = np.array([]).astype(int)
+        for _nset in self.node_sets:  # Assumed that all (outer) lists have been resolved
+            node_ids = np.union1d(node_ids, c.nodes[population].ids(_nset))
```
chr-pok, line 370: Use of `self.node_sets` is broken. `node_ids` is computed here but is not combined with the results of the `property_filter` below.

examples/neuron_set_example.ipynb
```diff
@@ -356,32 +347,23 @@
...
+      "PropertyNeuronSet resolved in population 'S1nonbarrel_neurons' of circuit 'SSCX_O1':\n",
+      "> Neuron IDs (7786): [ 23620  23621  23622 ... 170851 170852 170853]\n",
+      "> Node set dict: {'layer': ['2', '3'], 'synapse_class': 'INH'}\n"
      ]
     }
    ],
    "source": [
-    "neuron_set = obi.PropertyNeuronSet(property_specs={\"synapse_class\": \"INH\"}, node_sets=(\"Layer2\", \"Layer3\"))\n",
+    "neuron_set = obi.PropertyNeuronSet(\n",
+    "    property_filter=obi.scientific.circuit.neuron_sets.NeuronPropertyFilter(filter_dict={\"layer\": [\"2\", \"3\"], \"synapse_class\": [\"INH\"]}),\n",
```
chr-pok, line 365: This example (h) was supposed to demonstrate the use of property filters in combination with `node_sets`, so e.g. `node_sets=("Layer2", "Layer3") should be added. But its functionality is currently broken, see other comment.

examples/circuit_simulations_example.ipynb
```diff
@@ -76,10 +76,16 @@
     "    simulation_length=sim_duration,\n",
     ")\n",
     "\n",
+    "\n",
+    "\n",
     "# Stimuli\n",
+    "property_neuron_set = obi.PropertyNeuronSet(\n",
+    "    property_filter=[obi.scientific.circuit.neuron_sets.NeuronPropertyFilter(filter_dict={\"layer\": [\"2\", \"3\"], \"synapse_class\": [\"INH\"]}),\n",
+    "                     obi.scientific.circuit.neuron_sets.NeuronPropertyFilter(filter_dict={\"layer\": [\"1\", \"2\", \"3\"], \"synapse_class\": [\"EXC\"]})],\n",
```
chr-pok, line 84: For this specific example using the toy circuit `ToyCircuit-S1-6k`, "layer" should be given as integer numbers, not strings.

examples/circuit_simulations_example.ipynb
```diff
@@ -76,10 +76,16 @@
     "    simulation_length=sim_duration,\n",
     ")\n",
     "\n",
+    "\n",
+    "\n",
     "# Stimuli\n",
+    "property_neuron_set = obi.PropertyNeuronSet(\n",
+    "    property_filter=[obi.scientific.circuit.neuron_sets.NeuronPropertyFilter(filter_dict={\"layer\": [\"2\", \"3\"], \"synapse_class\": [\"INH\"]}),\n",
```
chr-pok, line 83: Could we add the class `NeuronPropertyFilter` to \_\_init\_\_.py? So that we can access it with `obi.NeuronPropertyFilter`?

chr-pok (commented):
Nice integration of property filters! Just a few things I noticed that may need to be fixed...

james-isbister:
Looks good to me. I'll merge it!

PR #184, opened 2025-05-26
Partial linting of Morphology Locations + prints replaced by logging in some other files

james-isbister:
* Partial linting of Morphology Locations
* Prints replaced by logging in some other files
* Connectome utilities made required dependency


obi_one/scientific/morphology_locations/specified_morphology_locations.py
MWolfR, line 1: I really hope the re-namings don't break things....

MWolfR (dismissed):
Optimistic approval

james-isbister:
I reran the morphology example notebook and the generated dataframes were identical to before the linting so I think we're the linting has not broken anything.

I found a single stray "/" which was producing a warning but that is now removed.

@MWolfR had already approved the PR @danielaegassan, just the removal of the forward slash means we require re-approval. If you reapprove @danielaegassan I can merge this PR which also adds conn-utilities as a regular dependency

danielaegassan (approved):
Already reviewed by Michael.

PR #212, opened 2025-05-28
Adding a connectivity matrix asset option to the circuit class

danielaegassan:
Adding path to optional asset of a circuit connectivity matrix and a loader for it.

### DIFF
--- a/obi_one/scientific/circuit/circuit.py
+++ b/obi_one/scientific/circuit/circuit.py
@@ -55,15 +55,12 @@ class Circuit(OBIBaseModel):
     def _default_population_name(c):
         """Returns the default node population name of a SONATA circuit c."""
         popul_names = Circuit.get_node_population_names(c, incl_virtual=False)
-        #popul_names = ["A"]
         assert len(popul_names) == 1, "Default node population unknown!"
         return popul_names[0]
 
     @property
     def default_population_name(self):
         """Returns the default node population name."""
-        #popul_names = self.get_node_population_names(incl_virtual=False)
-        #assert len(popul_names) == 1, "Default node population unknown!"
         return self._default_population_name(self.sonata_circuit)
 
     def get_edge_population_names(self, incl_virtual=True):


obi_one/scientific/circuit/circuit.py
```diff
@@ -1,13 +1,15 @@
...
 from obi_one.core.base import OBIBaseModel
 
 
 class Circuit(OBIBaseModel):
-    """Class representing a circuit, i.e., pointing to a SONATA config."""
+    """Class representing a circuit, i.e., pointing to a SONATA config and possible additional assets."""
 
     name: str
     path: str
+    matrix_path: str = None
```
james-isbister, line 12: I think this could be: `matrix_path: Optional[str] = None ` to use the latest notation

james-isbister (commented):
Looks good to me

obi_one/scientific/circuit/circuit.py
```diff
@@ -1,13 +1,15 @@
...
 
     def __init__(self, name, path, **kwargs):
```
chr-pok, line 14: I would add in `__init__()` a check that the matrix file exists and is in the correct format, as it is done with the SNAP circuit, by trying to load it. And also check that the node IDs are consistent between the SNAP circuit and the matrix. Something like:
```
if self.matrix_path is not None:
    _cmat = ConnectivityMatrix.from_h5(self.matrix_path)  # Basic check: Try to load the connectivity matrix w/o error
    np.testing.assert_array_equal(_cmat.vertices["node_ids"], _c.nodes[circuit.default_population_name].ids())
```

Note: In this code snippet I am assuming that the connectivity matrix belongs to the (local) connectivity within the circuit's default node population (=non-virtual and only one existing) which I think is fine for now. Later we may need to extend this to handle multiple connectivity matrices for connectivity between different node populations as well.

chr-pok (commented):
Looks good to me! Just a few suggestions for improvement.

obi_one/scientific/circuit/circuit.py
danielaegassan: mmmmm if I put this in the init, then it will effectively load the matrix twice if there is not error no?  This can be costly if the circuit is big, but I add it.

james-isbister: You could make _cmat a private property which is only loaded once through something like this:

```
_cmat = None  # type: Optional[ConnectivityMatrix]

@property
def cmat(self) -> ConnectivityMatrix:
    """Return the ConnectivityMatrix loaded from the HDF5 file."""
    if self._cmat is None:
        self._cmat = ConnectivityMatrix.from_h5(self.matrix_path)
    return self._cmat
```

You can then access it with self.cmat of circuit.cmat

danielaegassan: This is already similar to what's there no?  The question what about the checking that the nodes in the matrix and the nodes in the circuit correspond to each other and how to do that avoiding a circular logic. 


I have implemente @chr-pok suggestion already, I let him check it when he comes back on Monday.

danielaegassan:
@chr-pok, @james-isbister  I don't need connalysis here but in the extraction function, so I just removed your comments and for me it's ready to merge.  The error that it didn't find connectome utilities has nos disappeared since it has been added to global dependencies. So please merge if you agree.

I will continue working on this branch after the merge.

Issue #37, opened 2025-04-02
Pull circuit extraction brainbuilder fixes into brainbuilder main


james-isbister:
It would be great to merge @chr-pok 's circuit extraction fix into main so that we can install brainbuilder without needing to reference a local copy. 

Pyproject.ml and README.md can also be updated once this is done

chr-pok:
A PR has been created here: https://github.com/openbraininstitute/brainbuilder/pull/2

@GianlucaFicarelli, it would be great if you could have a look please.

evakennyobi:
This is still being worked on @chr-pok ? thanks

chr-pok:
Well, there is nothing more I can do at the moment until someone will review my PR (it is not urgent though, since we are not blocked by this). So I would consider it as blocked.

mgeplf:
Sorry, that's on me; I haven't had time. I'll be back at 100% starting next week, so I will for have time either this week or next.

james-isbister:
Thanks @mgeplf I'll close this ticket @chr-pok

PR #216, opened 2025-06-02
Changed brainbuilder to main branch

chr-pok:
After brainbuilder fixes have been merged into main, the branch needs to be updated.

Related issue: https://github.com/openbraininstitute/obi-one/issues/37

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ include = ["obi_one*"]
 
 [tool.uv.sources]
 bluepysnap = { git = "https://github.com/openbraininstitute/snap.git" }
-brainbuilder = { git = "https://github.com/openbraininstitute/brainbuilder.git", branch = "split_popul_fix" }
+brainbuilder = { git = "https://github.com/openbraininstitute/brainbuilder.git" }
 
 [tool.ruff]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -68,7 +68,7 @@ include = ["obi_one*"]
 
 [tool.uv.sources]
 bluepysnap = { git = "https://github.com/openbraininstitute/snap.git" }
-brainbuilder = { git = "https://github.com/openbraininstitute/brainbuilder.git", branch = "split_popul_fix" }
+brainbuilder = { git = "https://github.com/openbraininstitute/brainbuilder.git" }
```
GianlucaFicarelli, line 71: It would be even better if brainbuilder can be released with the changes, remove this line, and set the minimum version in the dependencies above.

This can be done for snap as well, since the version 3.0.2 has been already released.

chr-pok: Yes, indeed, this should be done once the new brainbuilder version has been released. But there are more fixe in brainbuilder currently in progress, so not sure when a new release will be ready.

This is just meant to be a quick fix so that obi-one does not break because the `split_popul_fix` branch does not exist any more.

PR #217, opened 2025-06-05
Activity, Validation + ReconstructionMorphologyValidation classes + reconstruction_morphology_validation_example

james-isbister:
* Activity, Validation + ReconstructionMorphologyValidation classes
* examples/validation/reconstruction_morphology_validation_example.ipynb

I split each validation into run() and save() where the logic and database operations could go respectively. Then we are at liberty to run them in isolation, or collectively in a later generated endpoint, or executed by a workflow engine.

For the ReconstructionMorphologyValidation we use just a local morphology to begin. I think we can integrate with entitycore as a second step.

Once this is merged in main, you could then add some logic for filling the ReconstructionMorphologyValidationOutput, using the notebook to test @dkeller9 ?

Or classes and examples could be begun for other types of Entity also. @AurelienJaquier would like to add the code he has written for ME-Model validation eventually also

The code is very isolated from the rest of the repo, so this is more of a way for you to see which changes have been made (and would be needed for adding new Validations).

**Note a base Activity class has now been created which Validation inherits from. Both are in obi_one.core**

### DIFF
--- a/obi_one/scientific/validations/reconstruction_morphology_validation.py
+++ b/obi_one/scientific/validations/reconstruction_morphology_validation.py
@@ -1,17 +1,15 @@
-from obi_one.scientific.validations.validation import Validation
+import logging
+from pathlib import Path
+from typing import Annotated, ClassVar
 
-import neurom
 import morphio
-
-from typing import ClassVar, Annotated
-
+import neurom
 from pydantic import BaseModel, Field
 
-from pathlib import Path
-
-import logging
+from obi_one.scientific.validations.validation import Validation
 
 L = logging.getLogger(__name__)
+
 
 class ReconstructionMorphologyValidationOutput(BaseModel):
     validation_a: Annotated[
@@ -45,7 +43,6 @@ class ReconstructionMorphologyValidation(Validation):
 
     def run(self) -> None:
         """Run the validation logic."""
-
         L.info("Running Reconstruction Morphology Validation")
 
         if not self.morphology_file_path:
@@ -60,17 +57,12 @@ class ReconstructionMorphologyValidation(Validation):
         )
 
         # Implement the validation logic here
-        pass
 
     def save(self) -> None:
         """Save the result of the validation."""
-        
         L.info("Saving Reconstruction Morphology Validation Output")
 
         if self._validation_output is None:
             raise ValueError("Validation output must be set before saving.")
-        
+
         # Example: Save the validation output to a database or file
-
-        pass
-
--- a/obi_one/scientific/validations/reconstruction_morphology_validation.py
+++ b/obi_one/scientific/validations/reconstruction_morphology_validation.py
@@ -1,5 +1,35 @@
 from obi_one.scientific.validations.validation import Validation
 
+import neurom
+import morphio
+
+from typing import ClassVar, Annotated
+
+from pydantic import BaseModel, Field
+
+from pathlib import Path
+
+import logging
+
+L = logging.getLogger(__name__)
+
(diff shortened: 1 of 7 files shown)


james-isbister:
Just to say we have got started on this @jdcourcol

obi_one/scientific/validations/reconstruction_morphology_validation.py
```diff
@@ -0,0 +1,68 @@
...
+import morphio
+import neurom
+from pydantic import BaseModel, Field
+
+from obi_one.scientific.validations.validation import Validation
+
+L = logging.getLogger(__name__)
+
+
+class ReconstructionMorphologyValidationOutput(BaseModel):
```
AurelienJaquier, line 14: So ReconstructionMorphologyValidationOutput contains the output for each validation that is being run in ReconstructionMorphologyValidation, if I understand correctly. I think it would be better if validation_a, validation_b, etc. were not just the 'validation passed or not' bool, but were a dict or object containing everything that would be saved in the ValidationResult in entitycore.
So that would be indeed the 'validation passed' bool, but also the validation name, and eventually some log string or the figures produced during validation or anything else that might be relevant to morphology validation output.
Making them dict or obejct would give us more flexibility for what we want to pass as output.
For reference, [here](https://github.com/openbraininstitute/BlueCelluLab/blob/main/bluecellulab/validation/validation.py#L186) is a dict output of an memodel validation

james-isbister:
Note: a base Activity class has now been created which Validation inherits from. Both are in obi_one.core

obi_one/scientific/validations/reconstruction_morphology_validation.py
AurelienJaquier: @james-isbister I pushed a commit to this PR with a new class SingleValidationOutput, representing the base structure of one single validation output that would be part of e.g. ReconstructionMorphologyValidationOutput.
It has:
- `name` and `passed`, since they are needed in entitycore's ValidationResult
- `validation_details` and `assets` as optional, since I think we would like to have those for most validation outputs
This SingleValidationOutput can then be inherited in SingleReconstructionMorphologyValidationOutput in case we want to further customise it for the morphology.

obi_one/scientific/validations/reconstruction_morphology_validation.py
```diff
@@ -0,0 +1,72 @@
...
+        """Run the validation logic."""
+        L.info("Running Reconstruction Morphology Validation")
+
+        if not self.morphology_file_path:
+            raise ValueError("File path must be provided for validation.")
+
+        neurom_morphology = neurom.load_morphology(self.morphology_file_path)
+        morphio_morphology = morphio.Morphology(self.morphology_file_path)
+
+        self._validation_output = ReconstructionMorphologyValidationOutput(
```
james-isbister, line 58: Looks good to me @AurelienJaquier 

I think this line just needs updating to use SingleReconstructionMorphologyValidationOutput?

dkeller9:
I wonder if there is a way to make the structure more general, so for example ValidationOutput instead of ReconstructionMorphologyValidationOutput. Then the entity type is indicated as a field within the structure. That way we don't have to make a new class to support every type of entity.

james-isbister:
> I wonder if there is a way to make the structure more general, so for example ValidationOutput instead of ReconstructionMorphologyValidationOutput. Then the entity type is indicated as a field within the structure. That way we don't have to make a new class to support every type of entity.

Sounds good  @dkeller9 . Let's think if theres such a general format as we go?

PR #223, opened 2025-06-10
ReconstructionMorphologyFromID updated, Generated MorphologyMetrics endpoint tested, db_client parameter used consistantly

james-isbister:
* ReconstructionMorphologyFromID updated
* Generated MorphologyMetrics endpoint tested
* db_client parameter used consistently and no longer need for token


chr-pok (dismissed):
Not really looked into the details but seems to be ok.

PR #224, opened 2025-06-11
work around new docker-compose

mgeplf:
* docker-compose seems to have gotten picky about --profiles when using build, between version 2.35.1 and 2.36.0, the latter is used in the CI.
* probably due to [BUG] Explicitly building services that are not part of COMPOSE_PROFILES broken since 2.36.0 docker/compose#12825


danifr (approved):
LGTM

PR #225, opened 2025-06-11
BlockReferences

james-isbister:
* circuit_simulations_example.ipynb updated to use new add / set options. Still needs tidying up and implementation of previously used Blocks
* BlockReference class used and implemented for NeuronSets and Timestamps
* Shared Block functionality moved to Block class
* ref getter and setter added to Block base class
* Form base class fills block reference parameters
* Info Block
* EntityFromID inherits from OBIBaseModel
* Circuit fixes
* simulation_level_name made private variable


chr-pok (approved):
Some non-critical things left to be done for later, as discussed. Otherwise, looks good!

PR #230, opened 2025-06-13
Circuit simultaion example updated. block_mapping removes hardcoded add functionality, and need to specify name of Block dictionary

james-isbister:
* Circuit simultaion example updated. 
* block_mapping removes hardcoded add functionality, and need to specify name of Block dictionary


chr-pok (approved):
Looks good!

chr-pok (approved):

PR #231, opened 2025-06-13
Simplex subcircuit extraction

danielaegassan:
Adding simplex extraction class, simple version.  Generate a node set with source/target a given node_id.


obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -439,6 +439,101 @@ def _get_expression(self, circuit: Circuit, population: str) -> dict:
...
+        name="Dimension",
+        description="Dimension of the simplices to be extracted",
+    )
+    position: Literal['source', 'target'] = Field('source',
+        name="Position",
+        description="Position of the center node in the simplex, it can be either source or target",
+    )
+    subsample: bool = Field(False,
+        name="subsample",
+        description="Whether to subsample to set of nodes in the simplex lists or not",
```
james-isbister, line 460:
```suggestion
        description="Whether to subsample the set of nodes in the simplex lists or not",
```

obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -439,6 +439,101 @@ def _get_expression(self, circuit: Circuit, population: str) -> dict:
...
+    )
+    n_count_max: Optional[int] = Field(None,
+        name="Max node count",
+        description="Maximum number of nodes to be subsampled",
+    )
+    subsample_method: Literal['node_participation', 'random', 'sample_simplices'] = Field('node_participation',
+        name="Method to subsample nodes the extracted simplices",
```
james-isbister, line 467:
```suggestion
        name="Method to subsample nodes from the extracted simplices",
```

obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -439,6 +439,101 @@ def _get_expression(self, circuit: Circuit, population: str) -> dict:
         expression = {"population": population, "node_id": list(df["node_ids"].astype(int))}
         return expression
 
+class SimplexNeuronSet(PropertyNeuronSet):
+    """
+    Get neurons that form simplices of a given dimension with a chosen source or target neuron.
```
james-isbister, line 444: I'm just a little confused by the terminology so maybe it could be described here in more detail.

For example, this is a "SimplexNeuronSet" extractor which suggests it extracts a single simplex, but then there is the concept of simplex_lists and sampling simplices ("sample_simplices"), which suggests the neuron_set might be some combination of multiple simplices.

obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -439,6 +439,101 @@ def _get_expression(self, circuit: Circuit, population: str) -> dict:
...
+    """
+    center: int = Field(
```
james-isbister, line 446:
```suggestion
    centeral_neuron_id: int = Field(
```

obi_one/scientific/circuit/neuron_sets.py
```diff
@@ -439,6 +439,101 @@ def _get_expression(self, circuit: Circuit, population: str) -> dict:
...
+        name="Center node",
+        description="Node index that will be source or target of the simplices extracted",
+    )
+    dim: int = Field(
+        name="Dimension",
+        description="Dimension of the simplices to be extracted",
+    )
+    position: Literal['source', 'target'] = Field('source',
```
james-isbister, line 454:
```suggestion
    central_neuron_position: Literal['source', 'target'] = Field('source',
```

james-isbister (commented):
Looks good, although I just need some small clarifications to fully understand

obi_one/scientific/circuit/neuron_sets.py
MWolfR: Yes, I agree. And I would even say this is not just about naming things.
If I understand this correctly, there are two conceptual ways of using this: 

With subsample_method=random or subsample_method=node_participation there is no guarantee that the resulting selection forms any simplex of the specified dimension. While the selection is based on simplex participation, it is still possible that the random sample just happens to sample from non-overlapping simples.

With subsample_method="sample_simplices" you are guaranteed to get simples as specified.

I suggest they become two separate NeuronSets. Maybe SimplexMembershipBasedNeuronSet and SimplicesNeuronSet respectively.

For SimplicesNeuronSet it may be more intuitive to specify the number of simplices to sample rather then the number of nodes.

obi_one/scientific/circuit/neuron_sets.py
danielaegassan: It is not the position it's the id.  So I'll go with James' suggestion.

obi_one/scientific/circuit/neuron_sets.py
james-isbister: I'd suggest the following names (and to check my understanding):
- SimplexNeuronSet (A neuron set that contains all the neurons in a single simplex)
- SampleFromSimplicesNeuronSet (A neuron set which samples from multiple simplices)

MWolfR (changes requested):
Two comments:
1. What do you want to happen if the requested dimension cannot be reached? In simplex_extractors the code is lenient and uses the maximum dimension reached instead. In the NeuronSet itself it is stricter: If the number of neurons is < requested dimension, then an error is raised. Just want to make sure this is consistent and does what you intended it to do.
2. More important: When a NeuronSet is used for example as a source for a spike replay stimulus, its "_get_expression" method is called 3 or 4 times. In this case, this is a rather expensive function to run, right? So I suggest you store its output in a private variable during the first call, and after the first call it just returns the stored result.

PR #237, opened 2025-06-18
Circuit, Simulation, SimulationCampaign entity integration to SimulationsForm

james-isbister:
* Simulation used new CircuitFromID and registers campaign, simulation and simulationgeneration in staging with all assets and demonstrates staging of same files
* entitysdk_circuit_simulations.ipynb calls endpoint for testing and demonstrates staging of same files
* Some generated morphology endpoint fixes + adjustments to work with new changes
* spikingstimulus node sets no longer default to none

### DIFF
--- a/obi_one/scientific/basic_connectivity_plots/helpers.py
+++ b/obi_one/scientific/basic_connectivity_plots/helpers.py
@@ -489,7 +489,11 @@ def plot_smallMC_network_stats(conn, full_width,
     adj_plot=adj.toarray().astype(float)
     adj_plot[adj_plot == 0] = np.nan
     # Connectivity matrix 
-    plot=axs[0].imshow(adj_plot, cmap=cmap_adj, interpolation='nearest', aspect='auto', vmin=0, vmax=np.nanmax(adj_plot))
+    min_val = int(np.nanmin(adj_plot[adj_plot > 0]))
+    max_val = int(np.nanmax(adj_plot))
+    plot=axs[0].imshow(adj_plot, cmap=cmap_adj, interpolation='nearest', aspect='auto', vmin=min_val, vmax=max_val)
+    axs[0].xaxis.set_major_locator(plt.MaxNLocator(integer=True))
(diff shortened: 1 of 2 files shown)


examples/circuit_simulations_example.ipynb
```diff
@@ -263,7 +150,11 @@
   },
   {
    "cell_type": "code",
+<<<<<<< HEAD
+   "execution_count": null,
+=======
    "execution_count": 7,
+>>>>>>> main
```
chr-pok, line 157: Unresolved merge conflict. File seems to be corrupted because of that.

obi_one/database/circuit_from_id.py
```diff
@@ -0,0 +1,30 @@
...
+    def circuit_directory(self, dest_dir=Path(), db_client: entitysdk.client.Client = None) -> None:
+        for asset in self.entity(db_client=db_client).assets:
+            if asset.content_type == "application/vnd.directory":
+
+                # Download the content into memory
+                db_client.download_directory(
+                    entity_id=self.entity(db_client=db_client).id,
+                    entity_type=self.entitysdk_type,
+                    asset_id=asset.id,
+                    output_path=dest_dir,
```
chr-pok, line 29: Perhaps before downloading it should be checked that the `dist_dir` is a "clean" directory (i.e., either does not exist or is empty)? So not to mix up files with existing ones from previous runs etc.

obi_one/endpoints/generated_endpoints.py
```diff
@@ -80,44 +81,52 @@ def create_endpoint_for_form(
...
             except Exception:  # noqa: BLE001
                 L.exception("Generic exception")
             else:
                 L.info("Grid scan generated successfully")
-                return result
+                if campaign is not None:
+                    return str(campaign.id)
+                else:
+                    print("No campaign generated")
+                    return "Success but no campaign generated."
```
chr-pok, line 113: Looks a bit strange to return either an ID or a human-readable string. Shouldn't it be something like `None` or `""` in that case?

chr-pok (approved):

PR #240, opened 2025-06-18
Simplex subcircuit extraction

danielaegassan:


danielaegassan:
Adding example to the notebook.

PR #242, opened 2025-06-18
Publish package

james-isbister:
Draft publish: https://test.pypi.org/project/obi-one/
Would need to update README also

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,0 +1,48 @@
+name: PyPI
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+     - main
+  release:
+    types:
+      - published
+
+jobs:
+
+  build_sdist:
+    name: Build sdist
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build SDist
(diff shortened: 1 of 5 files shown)


jdcourcol (dismissed):
I guess you need a github action to publish the package.

james-isbister:
@heerener I guess with this setup we don't need to change `make install` in the `Makefile`? Perhaps I can reverse that change which removed the generation of the ipykernal from make install?

heerener:
> @heerener I guess with this setup we don't need to change `make install` in the `Makefile`? Perhaps I can reverse that change which removed the generation of the ipykernal from make install?

I'm not very familiar with pipx, but it doesn't look like it utilises the Makefile (please ignore the rest of this comment if I'm wrong about this). The publishing workflow doesn't use the Makefile either, so any changes there are entirely without impact to this whole process.

heerener (approved):

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,48 @@
+name: PyPI
+
+on:
+  workflow_dispatch:
+  pull_request:
```
heerener, line 5: Do you really want to trigger this workflow on every pull request?

heerener (approved):

PR #255, opened 2025-06-25
POST endpoint for electrophysiology metrics using form input

ilkilic:
Updates the /electrophysiologyrecording-metrics endpoint to support POST requests using a payload based on `ElectrophysiologyMetricsForm`.

Clients can now compute electrophysiological metrics by submitting the following JSON payload:

```
{
  "initialize": {
    "trace_id": "<uuid>",                  // required: ID of the electrical recording
    "protocols": ["step", "iv"],           // optional: stimulus protocol types
    "requested_metrics": ["AP_height"],    // optional: metrics to extract
    "amplitude": {
      "min_value": 0.1,
      "max_value": 0.4
    }                                      // optional: target amplitude range in nA
  }
}
```

An example usage is available in the test_endpoints notebook


james-isbister (approved):
Looks good. I tested using test_endpoints.ipynb

james-isbister (dismissed):

WonderPG (commented):
Thanks for the changes !

AurelienJaquier:
Hi! I just made a push to this PR. I implemented @WonderPG 's suggestions. Does it look better now?

WonderPG:
@AurelienJaquier It does look much better thank you for that ! Two small comments I would add, by declaring these as such in the endpoint:
```python
requested_metrics: CALCULATED_FEATURES | None = None,
amplitude: AmplitudeInput | None = None,
protocols: STIMULI_TYPES | None = None,
```
they actually are interpreted as a body in FastAPI. I think it would be better if they were query parameters, since `GET` requests should not have a body.
For instance the following should do it:
```python
requested_metrics: CALCULATED_FEATURES | None = Query(default=None),
amplitude: AmplitudeInput = Depends(),
protocols: STIMULI_TYPES | None = Query(default=None),
```
given that the class `AmplitudeInput` is given defaults:
```python
class AmplitudeInput(BaseModel):
    """Amplitude class."""

    min_value: float | None = Field(default=None, description="Minimum amplitude (nA)")
    max_value: float | None = Field(default=None, description="Maximum amplitude (nA)")
```
Please double check that it doesn't break anything.

Also another performance comment, I know this came from our old code but the core code has default handling as such:
```python
# Deal with cases where user did not specify stimulus type or/and feature
    if not stimuli_types:
        # Default to all protocol types if not specified
        logger.warning("No stimulus type specified. Iterating over all POSSIBLE_PROTOCOLS.")
        stimuli_types = list(POSSIBLE_PROTOCOLS.keys())

    if not calculated_feature:
        # Compute ALL of the available features if not specified
        logger.warning("No feature specified. Defaulting to everything.")
        calculated_feature = list(CALCULATED_FEATURES.__args__[0].__args__)  # type: ignore
# Turn amplitude requirement of user into a bluepyefe compatible representation
    if isinstance(amplitude, AmplitudeInput):
        # If the user specified amplitude/a range of amplitudes,
        # the target amplitude is centered on the range and the
        # tolerance is set as half the range
        desired_amplitude = mean(
            [
                amplitude.min_value,
                amplitude.max_value,
            ]
        )

        # If the range is just one number, use 10% of it as tolerance
        if amplitude.min_value == amplitude.max_value:
            desired_tolerance = amplitude.max_value * 0.1
        else:
            desired_tolerance = amplitude.max_value - desired_amplitude
    else:
        # If the amplitudes are not specified, take an arbitrarily high tolerance
        desired_amplitude = 0
        desired_tolerance = 1e12
```
as you guys probably know, computing all features for all protocols and all amplitudes is quite heavy. We have tried to make 2-3 calls async to this endpoint and it ended up crashing the whole `obi-one` pod which had to restart. Maybe having more reasonable default behaviors is the way to go.

darshanmandge:
@WonderPG is the PR good now and do you think we can test in the next.staging?

WonderPG (approved):
LGTM, thank you very much !

Issue #226, opened 2025-06-12
SynapticManipulation Block type with two demo hardcoded connection override blocks


james-isbister:
The Sonata blocks, which are needed:

```
  "connection_overrides": [
     {
          "name": "ach_use",
          "source": "All",
          "target": "All",
          "synapse_configure": "%s.Use *= 0.7050728631217412"
     }
  ]
```

```
"connection_overrides": [
     {
          "name": "Mg",
          "source": "All",
          "target": "All",
          "synapse_configure": "%%s.mg = 2.4"
     }
  ]
```

PR #266, opened 2025-06-27
SynapticManipulation block type with two specific connection override blocks

chr-pok:
Implemented two specific synaptic manipulation classes (blocks) to be used in the `SimulationsForm`:
- `SynapticUseManipulation`: Allows a user to scale the synaptic Use value (globally, for all connections)
- `SynapticMgManipulation`: Allows a user to set the synaptic Magnesium value (globally, for all connections)

Added a simple example notebook as well.

IMPORTANT: In general, these manipulations rely on the order they are put into the dict, as they are execute in order! For the ones we have at the moment, this should not be relevant though. But I have added a note that we should check at some point that the order is always preserved.

Related ticket: https://github.com/openbraininstitute/obi-one/issues/226

### DIFF
--- a/obi_one/scientific/simulation/manipulations.py
+++ b/obi_one/scientific/simulation/manipulations.py
@@ -8,6 +8,10 @@ class SynapticManipulation(Block, ABC):
     def _get_override_name(self) -> str:
         pass
 
+    def config(self) -> dict:
+        self.check_simulation_init()
+        return self._generate_config()
+
     def _generate_config(self) -> dict:
         sonata_config = {
             "name": self._get_override_name(),
--- a/obi_one/scientific/simulation/manipulations.py
+++ b/obi_one/scientific/simulation/manipulations.py
@@ -1,0 +1,41 @@
+from abc import ABC, abstractmethod
+from obi_one.core.block import Block
+
+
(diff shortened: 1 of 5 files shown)


chr-pok:
Just wondering, should we even use default values
```
use_scaling: float | list[float] = 0.7050728631217412
magnesium_value: float | list[float] = 2.4
```
or is this too specific?

chr-pok:
Here is an example simulation config that was generated by the example notebook:
[simulation_config.json](https://github.com/user-attachments/files/20944609/simulation_config.json)

obi_one/scientific/simulation/manipulations.py
```diff
@@ -0,0 +1,45 @@
...
+    def _get_override_name(self) -> str:
+        return "ach_use"
+
+    def _get_synapse_configure(self) -> str:
+        return f"%s.Use *= {self.use_scaling}"
+
+
+class SynapticMgManipulation(SynapticManipulation):
+
+    magnesium_value: float | list[float] = 2.4
```
james-isbister, line 39: default values are fine, but both this and Use should be NonNegativeFloat

james-isbister (approved):

PR #280, opened 2025-07-02
Descriptions

james-isbister:
* Two descriptions updated
* full stops added to several descriptions

The first two commits were already merged in a previous PR

### DIFF
--- a/obi_one/scientific/simulation/stimulus.py
+++ b/obi_one/scientific/simulation/stimulus.py
@@ -161,7 +161,7 @@ class LinearCurrentClampSomaticStimulus(SomaticStimulus):
     amplitude_end: float | list[float] = Field(
         default=0.2,
         title="End Amplitude",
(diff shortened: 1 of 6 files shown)


obi_one/scientific/simulation/stimulus.py
```diff
@@ -161,7 +161,7 @@ class LinearCurrentClampSomaticStimulus(SomaticStimulus):
     amplitude_end: float | list[float] = Field(
         default=0.2,
         title="End Amplitude",
-        description="If given, current is interpolated such that current reaches this value when the stimulus concludes. Otherwise, current stays at amp_start. Given in nanoamps",
+        description="If given, current is interpolated such that current reaches this value when the stimulus concludes. Otherwise, current stays at amp_start. Given in nanoamps.",
```
chr-pok, line 164: Looks like `amp_start` should actually be called `amplitude_start`. However, since these variable names are not shown in the UI, it would be even better not to use variable names but textual descriptions, like "start amplitude".

chr-pok (commented):
Looks good. Just one minor comment.

Issue #293, opened 2025-07-09
Improve return value for unsupported data vs no features in ephys agent


ilkilic:
Right now, the ephys metric agent returns an empty result when no features are extracted and when the data type isn’t supported. It would be helpful to make this distinction clearer.

Suggestion:
Change the return value to clearly show whether:
- No features were extracted
- The data isn’t supported yet

PR #295, opened 2025-07-14
Ephys agent: Handle missing protocols in data with explicit error

ilkilic:
Addresses https://github.com/openbraininstitute/obi-one/issues/293. This improves the API response by raising a 404 error when none of the requested protocols are found in the data, instead of silently returning an empty result.


WonderPG (commented):
I tested it and it works. By any chance, are you able to tell which protocols do have data with minimal computational overhead ? It would be very useful in the error message to know from which protocol the user can extract data. If it is not possible to know without computing the whole thing, it's fine.

ilkilic:
> Also in my terminal, when I triggered the error, I got the following message:
> 
> ```
> TRTNWBReader only supports 'step' protocols, but requested: ['spontaps']. Skipping.
> TRTNWBReader only supports 'step' protocols, but requested: ['sponnohold30']. Skipping.
> ```
> 
> Does it mean that you are using a reader class that supports only the `step` protocol or is it related to the trace ? Indeed `step` was the only protocol that yielded data.

Yes, the current implementation expects that the traces read by TRTNWBReader correspond only to step protocols

WonderPG:
> Yes, that would be ideal, the available protocols could be retrieved with minimal overhead from the `stimuli` field of the ElectricalCellRecording via entitysdk. However, I just noticed that the human data are currently missing this information, we will make sure to update those. Would you prefer that this tool queries entitysdk directly or would you rather handle that on your end?

Since this kind of information is very tightly linked to the computation your endpoint makes, I would add it here. E.g. you could do some validation before computing features. If a requested stimuli is not available (as per entitycore), don't compute it at all. We will most likely let the agent access this information as well on our side, so that the user can either get this information upstream, or get the request denied quickly by your service.

> Yes, the current implementation expects that the traces read by TRTNWBReader correspond only to step protocols

Then shouldn't we allow only step protocols in the input of the endpoint or expand the current implementation ?

ilkilic:
I looked into integrating the validation based on stimuli but since not all traces in entitycore have that field filled in, I kept the existing bpe log based logic. When stimuli are available, I now filter out unsupported types and log a warning if any are missing. An error is raised only if none of the requested types are found. Let me know your thoughts

WonderPG:
It's fine, we can iterate later :)

PR #300, opened 2025-07-23
Named tuple

james-isbister:
* NamedTuple uses NonNegativeInt
* Version bump


WonderPG (approved):
The autogen tool now outputs the following:
```
class Element(RootModel[int]):
    root: int = Field(..., ge=0)


class NamedTuple(BaseModel):
    model_config = ConfigDict(
        extra='ignore',
    )
    type: Literal['NamedTuple'] = Field(..., title='Type')
    name: str = Field(..., title='Name')
    elements: list[Element] = Field(..., title='Elements')
    ```
which is what we wanted. I also tested my tool and it works.

Thank you !

PR #318, opened 2025-08-06
Endpoint definitions moved from package to app

james-isbister:
* Endpoint definitions moved from package to app
* 1 linting encoding fix
* Entitysdk electrical_cell_recording test data updated


pyproject.toml
```diff
@@ -34,10 +34,6 @@ dependencies = [
     "brainbuilder",
     # MORPHOLOGY CONTAINERIZATION
     "bluepysnap",
-    # FAST API SERVICE
-    "fastapi",
-    "uvicorn",
-    "starlette",
```
GianlucaFicarelli, line 40: How is uvicorn going to be installed when running the app?
Maybe these dependencies should be moved to a different dependency-groups instead of being deleted?

pyproject.toml
```diff
@@ -34,10 +34,6 @@ dependencies = [
...
     # JUPYTER
     "jupyter",
     "notebook",
```
GianlucaFicarelli, line 39: Commenting here, but related to `tool.uv.sources` below: do bluepysnap and brainbuilder still need to be installed from git, instead of using a package?

pyproject.toml
GianlucaFicarelli: > So maybe we can keep the dependencies explicit as part of the package also? Or is it cleaner to separate the dependencies for the app/package explicitly?

Separating them sounds good to me, if the final purpose is to make a scientific package that can be installed as a library instead of a service, but it can be done later when the package will be actually published.

james-isbister: Ok, I've readded the dependencies for now, and requested re-review

GianlucaFicarelli (dismissed):

pyproject.toml
```diff
@@ -33,7 +33,7 @@ dependencies = [
     # CIRCUIT EXTRACTION
     "brainbuilder",
```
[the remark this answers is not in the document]
james-isbister: Good point. i've updated this

GianlucaFicarelli (approved):

PR #339, opened 2025-08-21
add query parameters to morphology metrics. Fix openapi.json generation

jdcourcol:
- morphology metrics can be invoked with a query parameter as such:
curl ... /declared/neuron-morphology-metrics/${MORPH}?requested_metrics='soma_radius'\&requested_metrics='total_volume' 

- openapi.json generation was failing

### DIFF
--- a/app/endpoints/declared_endpoints.py
+++ b/app/endpoints/declared_endpoints.py
@@ -34,7 +34,7 @@ def activate_declared_endpoints(router: APIRouter) -> APIRouter:
         reconstruction_morphology_id: str,
         db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
         requested_metrics: Annotated[
-            list[Literal[*MORPHOLOGY_METRICS]] | None,  # type: ignore[misc]  # noqa: E261
+            list[Literal[*MORPHOLOGY_METRICS]] | None,  # type: ignore[misc]
             Query(
                 description="List of requested metrics",
             ),


jdcourcol:
The goal is to prevent metrics extraction from returning all features. Some of them are very large and could fill up the AI agent’s context. Returning them is also unnecessary unless specifically requested by the end user.

WonderPG (approved):
The following class has been successfully created based on the `openapi.json`
```python
class NeuronMorphologyMetricsEndpointDeclaredNeuronMorphologyMetricsReconstructionMorphologyIdGetParametersQuery(
    BaseModel
):
    requested_metrics: (
        list[
            Literal[
                'aspect_ratio',
                'circularity',
                'length_fraction_above_soma',
                'max_radial_distance',
                'number_of_neurites',
                'soma_radius',
                'soma_surface_area',
                'total_length',
                'total_height',
                'total_width',
                'total_depth',
                'total_area',
                'total_volume',
                'section_lengths',
                'segment_radii',
                'number_of_sections',
                'local_bifurcation_angles',
                'remote_bifurcation_angles',
                'section_path_distances',
                'section_radial_distances',
                'section_branch_orders',
                'section_strahler_orders',
            ]
        ]
        | None
    ) = Field(
        default=None, description='List of requested metrics', title='Requested Metrics'
    )
```
Also the defaults of the output class have been set to `None` in the `openapi.json`. I quickly played with the endpoint and it works as expected. So all good on our side thank you !

WonderPG:
Ah maybe one thing I would add, not selecting any metric results in the endpoint computing them all. Is it the expected behavior ?

jdcourcol:
>Ah maybe one thing I would add, not selecting any metric results in the endpoint computing them all. Is it the expected behavior ?

yes, that is the expected behaviour:
if you pass requested_metrics as query parameter, only these ones are computed. Otherwise, it is the current behaviour.

PR #340, opened 2025-08-21
Add setuptools-scm

GianlucaFicarelli:
With this change it should be possible to remove the hardcoded version in pyproject.toml, and read it from the tag created during the release.


chr-pok (approved):
I am not too familiar with these changes, but installation of obi-one seems to work fine.

PR #342, opened 2025-08-22
update stimuli list with what we actually have on the platform

AurelienJaquier:


AurelienJaquier:
@james-isbister is it ok if I merge this PR? I don't know the process for merging PRs on obi-one

Issue #91, opened 2025-04-15
Enable lint in obi-one


james-isbister:

evakennyobi:
Is this something currently being worked on @james-isbister ? thanks

chr-pok:
As discussed, I will start working on that once [PR#292](https://github.com/openbraininstitute/obi-one/pull/292) has been merged into `main`.

chr-pok:
I disabled the following linting checks regarding "TODO" comments (for now), since I think it is important not to loose these commets:
- "FIX002",  # Line contains TODO, consider resolving the issue
- "TD002",   # Missing author in TODO
- "TD003",   # Missing issue link for this TODO

chr-pok:
Won't fix the following either, as this requires discussion with @james-isbister:
~~~
obi_one/core/exception.py:1:7: N801 Class name `OBIONE_Error` should use CapWords convention
~~~

jdcourcol:
I am adding to standup board as this is blocking ML activities (example of code needed to answer chat questions about the circuit)

evakennyobi:
Can you add an end date for this please @james-isbister

evakennyobi:
@MWolfR to check

evakennyobi:
Thanks for the update
Does it unblock the ML team without it being complete?
As that was why it was included in the standup as it (& the other cleanup ticket - https://github.com/openbraininstitute/obi-one/issues/90) are blocking their ticket
https://github.com/openbraininstitute/neuroagent/issues/414

chr-pok:
Just taking a note not to forget about it: @ilkilic found a bug in simulations.py introduced by the linting fixes which prevents user-defined neuron sets to be actually added to the node_sets.json of a simulation. Will be fixed in his new branch.

jdcourcol:
It relates to this file: https://openbraininstitute.sharepoint.com/:x:/s/OBI-Scientificstaff/Eaho4Xbj9KlElbIHuR8dPtgBpuoXzvSOfkR0ICC-xHbiRg?e=JuAtnn

We need the code to answer to the queries in that file to be implemented in obi-one, so that the chat agent can use this code. We said we would do it after linting.
@MWolfR @james-isbister and @chr-pok  will look into that to.
I imagine we need "building blocks" instead of functions answering only to the questions from the spreadsheet.

chr-pok:
[...]

PR #346, opened 2025-08-26
Fix node_sets creation

ilkilic:
Fixes a bug in simulations.py where user-defined neuron sets were not properly written to node_sets.json, a regression introduced by earlier linting https://github.com/openbraininstitute/obi-one/issues/91#issuecomment-3201062830.


chr-pok (approved):
Thanks for the fix, @ilkilic! Node set generation works fine (again)!

PR #359, opened 2025-09-08
Obi one version fix

james-isbister:
* obi_one_version added as optional parameter to Scan and SingleCoordMixin


chr-pok (commented):
De-serialization works nicely now!

However, I am wondering at what point we should actually set the version. Currently, it is set in the model dump during serialization:
`model_dump["obi_one_version"] = version("obi-one")`

But the actual `GridScan` object has the verison set to None (unless you de-serialize a json). I think it would be nicer to set the version right from the beginning in the constructor of the `GridScan` or `SingleCoordinateMixin` class, instead of in the `model_dump`. Or even better, we could add the version directly to the `OBIBaseModel` instead and set it in the constructor there perhaps?

chr-pok (approved):

james-isbister:
I see, that could be nice, indeed. Let's merge the fix, and think about it some more though

PR #375, opened 2025-09-26
Change from Reconstruction to Cell

WonderPG:
Please have a thorough review, I blindly changed all of the `Reconstruction` into `Cell` and `reconstruction` into `cell`. I tested the morphology metric endpoint and it works, I didn't test the rest.


james-isbister (approved):
Looks good to me

PR #379, opened 2025-09-30
fixed biophysical circuit population

dkeller9:
This is a fix for the biophysical circuit endpoint needed by the ML team, which accidently was broken.


WonderPG (approved):
Looks good thank you !

PR #385, opened 2025-10-01
Support step dict

g-bar:


james-isbister (commented):
I've left feedback with my biggest concern here: 
https://github.com/openbraininstitute/prod-circuit-simulation/issues/61#issuecomment-3356931628

I would not be surprised if you need a unique class for each variable which requires constraints. This would be very inelegant imo. And by extension that would mean all other variables if we want the notation to be consistent for the Python library.

PR #396, opened 2025-10-06
Predefined nodeset uses entity_property_type, mapped_circuit_properties endpoint

james-isbister:


app/endpoints/declared_endpoints.py
```diff
@@ -344,6 +345,37 @@ def circuit_nodesets_endpoint(
             ) from err
         return CircuitNodesetsResponse(nodesets=circuit_metrics.names_of_nodesets)
 
+    @router.get(
+        "/mapped-circuit-properties/{circuit_id}",
+        summary="Mapped circuit properties",
+        description="Returns a dictionary of mapped circuit properties.",
+    )
+    def mapped_circuit_properties_endpoint(
```
g-bar, line 353: We already have this endpoint that seems to return the same thing: https://staging.openbraininstitute.org/api/obi-one/docs#/declared/circuit_nodesets_endpoint_declared_circuit__circuit_id__nodesets_get

For example for circuit: `c9e10151-8f07-4158-a3b3-205210ceb075`

`/declared/circuit-metrics/{circuit_id}` returns

```py
{'Circuit.NodeSet': ['with_morphology', 'all_intrinsic']}
```

and `/declared/circuit/{circuit_id}/nodesets`, returns:
```py
{'nodesets': ['with_morphology', 'all_intrinsic']}
```

obi_one/scientific/blocks/neuron_sets.py
```diff
@@ -363,10 +367,9 @@ class PredefinedNeuronSet(AbstractNeuronSet):
         sets file.
     """
 
-    node_set: (
-        Annotated[str, Field(min_length=1)]
-        | Annotated[list[Annotated[str, Field(min_length=1)]], Field(min_length=1)]
-    )
+    node_set: Annotated[
+        NodeSetType, Field(min_length=1, entity_property_type=CircuitPropertyType.NODE_SET)
```
g-bar, line 371: This just becomes `str | list[str]` in the schema, what happens if someone passes an arbitrary string?

obi_one/scientific/unions/unions_neuron_sets.py
```diff
@@ -45,7 +45,8 @@
     | nbS1POmInputs
     | AllNeurons
     | ExcitatoryNeurons
-    | InhibitoryNeurons,
+    | InhibitoryNeurons
+    | PredefinedNeuronSet,
```
WonderPG, line 49: Given that all of these classes in the union come from a pre-defined nodeset, wouldn't it be possible to remove every single class in this union beside `IDNeuronSet` and `PredefinedNeuronSet` to achieve the same granularity as what we have now ?

g-bar: Couldn't these be just strings in the PredefinedNeuronSet dropdown, instead of separate classes?

obi_one/scientific/blocks/neuron_sets.py
james-isbister: It is validated with an error raised but requires opening the circuit so doesn't happen immediately

app/endpoints/declared_endpoints.py
james-isbister: @jankrepl suggested keeping more endpoints for the agent for now. We are still prototyping with the agent, but I agree we should tidy these up once we find what's useful/works

PR #415, opened 2025-10-15
Reorganize neuron sets

james-isbister:
* Reorganization of neuron sets for easier navigation so we can start to think about using different NeuronSets within the UI for simulations and circuit extractions.
* neuron_sets.py split into 9 separate files (base, pair, id, combined, simplex, volumetric, specific, property, predefined)


james-isbister:
Thanks! I'm hopeful this will make things easier as we move neuron sets to the UI and create new neuron sets. We can certainly reduce the number of files in future though if its awkward!

PR #420, opened 2025-10-16
Integrate platform notebooks

james-isbister:
I wanted to see if it was possible to integrate obi_platform_analysis_notebooks into obi-one whilst maintaining all contributions (i.e. commits) from the original repository. I think this would be the case if it was merged as a normal merge (non-squashed).

I think this is a good idea because: 

1) The notebooks will be using the FromID types so this will make syncing a lot easier. 

2) This in turn may speed up the development of the FromID types which == sped up development of Tasks using those FromID types.

3) The community has an easy entry point to contributing code in obi-one. They start by adding a notebook, which we assess, and then we help them convert it to a Task.

4) We would like a one-stop shop repo that the community becomes familiar with, where they can find examples and which would even be something they'd like to contribute to. Having notebooks in a random other repo wouldn't encourage people to contribute, I think

5) Currently in obi_platform_analysis_notebooks there are no installation instructions, only requirements for each notebook. This means that a different virtual environment is needed for each notebook, rather than a single unified environment. I know it wasn't the plan, but I think it also wouldn't make sense to have a different set of dependencies in the notebooks repo anyway, which would need to be kept in sync with obi-one especially if they're using FromID types. Also the notebooks mostly run with the dependencies of obi-one already, and should be brought inline if some are going to be converted to Tasks.

**This is just a proposal to be discussed. There being a draft PR doesn't mean it has to happen**


james-isbister:
To do after platform entitysdk integration

PR #425, opened 2025-10-17
Add synaptome form

ilkilic:


james-isbister:
The notebook also works well for me. Good for @chr-pok to review the code as I was involved in writing it, and he knows this part of the code base

james-isbister:
I merged the new changes in main into dev, and then those changes in dev to this branch, as there were things unrelated to this PR showing up in the file changes

obi_one/scientific/from_id/circuit_from_id.py
```diff
@@ -42,3 +43,38 @@ def stage_circuit(
 
         msg = f"No 'sonata_circuit' asset found for Circuit with ID {self.id_str}."
         raise OBIONEError(msg)
+
+
+class MEModelWithSynapsesCircuitFromID(EntityFromID):
```
chr-pok: I think a check should be added when accessing `self.entity(db_client)` that the fetched Circuit entity is of scale "single". Otherwise, nothing prevents me from creating a `MEModelWithSynapsesCircuitFromID` object using any larger circuit ID, I tried and it did work.

Not exactly sure where such check would go though. It cannot be a post-validator I believe since it would require `db_client` to access the entity. So perhaps by overriding the `entity()` function would be an option:
```
    def entity(self, db_client: entitysdk.client.Client) -> Entity:
        entity = super().entity(db_client=db_client)
        if entity.scale != "single":
            msg = "Entity must be a circuit of scale 'single'."
            raise OBIONEError(msg)
```

chr-pok (commented):
Nice work! The example notebook also works very well!

I see there is a lot of redundancy in the code with the existing `Circuit` and `CircuitSimulation...` classes. But I see it cannot be handled by the generic circuit simulation classes because of the node_set which is not needed here (?).

Other than that, I have just one suggestion to add a check for scale 'single'.

obi_one/scientific/from_id/circuit_from_id.py
james-isbister: Good idea. Added this!

obi_one/scientific/unions/unions_neuron_sets.py
```diff
@@ -57,6 +57,12 @@
 ]
 
 
+MEModelWithSynapsesNeuronSetUnion = Annotated[
+    nbS1VPMInputs | nbS1POmInputs,
```
chr-pok: Just wondering, shouldn't we add `rCA1CA3Inputs` as well here, or was there a problem with that?

It is also missing in `SimulationNeuronSetUnion` btw and could/should be added there as well perhaps.

PR #457, opened 2025-10-24
import ion-channel-builder in obi-one

AurelienJaquier:
import ion-channel-builder in obi-one if present, else fallback to empty functions. -> should allow to use ion-channel-builder in obi-one.
Also fix execute function.


james-isbister (approved):
Not ideal faking the functions like this, but for SfN it works

PR #459, opened 2025-10-27
Mg override fix

james-isbister:
* Fix to solve simulation error
* New subclass GlobalSynapticManipulation which adds `modoverride`
* _get_synapse_configure string corrected for `SynapticMgManipulation`


darshanmandge (approved):
Looks correct as per the specifications in the [documentation](https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#connection-overrides). As you have tested this already, I think it is good.

PR #465, opened 2025-10-29
Default timestamps

james-isbister:


chr-pok (approved):
Seems to be working fine!

Issue #367, opened 2025-09-17
Entitycore integration for circuit extraction


chr-pok:
**Description:**
Circuit extraction currently works with the local file system only. Functionality should be added so that (a) the input circuit can be retrieved from entitycore, (b) the resulting extracted circuit can be registered as a new circuit in entitycore, and (c) the extraction scan config and related activities are registered as well (similar as for simulation campaigns). This will require the definition of new entities --> Sub-issue https://github.com/openbraininstitute/entitycore/issues/408.

Note: Circuit derivations may become redundant by introducing new extraction activities and may be removed at some (later) point.

**Acceptance criteria:**
- [x] Circuit extraction can directly load an input circuit from entitycore
- [x] Circuit extraction can directly register an extracted output circuit in entitycore
- [x] Circuit extraction configs and activities are registered in entitycore (this may not be strictly required and could come later, if needed) --> Sub-issue https://github.com/openbraininstitute/obi-one/issues/473

chr-pok:
The required entity types are being implemented in the `entitycore` branch [circuit-extraction](https://github.com/openbraininstitute/entitycore/tree/circuit-extraction).

Issue #347, opened 2025-08-27
Improve descriptions of CircuitExtraction Form + add block groupings (for generative UI)


james-isbister:

chr-pok:
Being worked on in the branch [circuit_extraction_UI_prep](https://github.com/openbraininstitute/obi-one/tree/circuit_extraction_UI_prep).

PR #469, opened 2025-10-29
Circuit extraction entitycore integration w/o activities

chr-pok:
Circuit extraction currently works with the local file system only. This PR extends the functionality so that (a) the input circuit can be retrieved as a `CircuitFromID` from entitycore, and (b) the resulting extracted circuit will be registered as a new `Circuit` entity in entitycore, including the "sonata_circuit" asset and contribution/publication links.

Note: The example notebook `A_service_and_entitycore/obi_one_entity_tools/entitysdk_example.ipynb` has been extended to demonstrate the new functionality

Related issue: https://github.com/openbraininstitute/obi-one/issues/367

### DIFF
--- a/tests/tasks/test_circuit_extraction.py
+++ b/tests/tasks/test_circuit_extraction.py
@@ -151,65 +151,73 @@ def test_circuit_extraction(tmp_path):
         ),
     ]
 
-    extraction_init = obi.CircuitExtractionScanConfig.Initialize(
-        circuit=circuit_list,
-        do_virtual=[False, True],
-        create_external=[False, True],
-        run_validation=False,
-    )
-    neuron_set = obi.PredefinedNeuronSet(node_set=["L6_IPC", "L6_TPC:A"])
-    info = obi.Info(campaign_name="Test", campaign_description="Test campaign")
+    for do_virtual in [False, True]:
+        for create_external in [False, True]:
+            scan_path = (
+                tmp_path
+                / "grid_scan"
+                / f"do_virtual={do_virtual}"
+                / f"create_external={create_external}"
+            )
+            extraction_init = obi.CircuitExtractionScanConfig.Initialize(
+                circuit=circuit_list,
+                do_virtual=do_virtual,
+                create_external=create_external,
+            )
+            neuron_set = obi.PredefinedNeuronSet(node_set=["L6_IPC", "L6_TPC:A"])
+            info = obi.Info(campaign_name="Test", campaign_description="Test campaign")
 
-    circuit_extractions_scan_config = obi.CircuitExtractionScanConfig(
-        initialize=extraction_init, neuron_set=neuron_set, info=info
-    )
+            circuit_extractions_scan_config = obi.CircuitExtractionScanConfig(
+                initialize=extraction_init, neuron_set=neuron_set, info=info
+            )
 
-    grid_scan = obi.GridScanGenerationTask(
-        form=circuit_extractions_scan_config,
-        output_root=tmp_path / "grid_scan",
-        coordinate_directory_option="ZERO_INDEX",
-    )
-    grid_scan.execute()
-    obi.run_tasks_for_generated_scan(grid_scan)
+            grid_scan = obi.GridScanGenerationTask(
+                form=circuit_extractions_scan_config,
+                output_root=scan_path,
+                coordinate_directory_option="ZERO_INDEX",
+            )
+            grid_scan.execute()
+            obi.run_tasks_for_generated_scan(grid_scan)
 
-    # Rerun --> Error since output file already exists
-    with pytest.raises(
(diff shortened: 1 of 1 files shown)


examples/C_forms/circuit_extraction/circuit_extraction_example.ipynb
```diff
@@ -51,6 +51,7 @@
     "                        run_validation=False,\n",
     "                    ),\n",
     "                    neuron_set=obi.PredefinedNeuronSet(node_set=\"Excitatory\", sample_percentage=[50, 100]),\n",
```
james-isbister: I think the virtual_sources_to_ignore should be made optional.

Currently building the form gives a missing parameter validation issue in this example

examples/A_service_and_entitycore/obi_one_entity_tools/entitysdk_example.ipynb
james-isbister: Building the form here works without missing parameters validation error, so I'm confused why it fails in the other notebook

examples/C_forms/circuit_extraction/circuit_extraction_example.ipynb
james-isbister: We could also remove references to "form" and use "scan_config" instead

obi_one/scientific/tasks/circuit_extraction.py
```diff
@@ -70,6 +83,243 @@ class CircuitExtractionSingleConfig(CircuitExtractionScanConfig, SingleConfigMix
...
+        self._temp_dir = tempfile.TemporaryDirectory()
+        return Path(self._temp_dir.name).resolve()
+
+    def _cleanup_temp_dir(self) -> None:
+        """Clean-up of temporary directory, if any."""
+        if self._temp_dir is not None:
+            self._temp_dir.cleanup()
+            self._temp_dir = None
+
+    def _resolve_circuit(self, *, db_client: Client, entity_cache: bool) -> None:
```
james-isbister: Maybe not for this PR but how much of this is shared with the simulation task resolve circuit? Do you think its reasonable for them to be shared eventually?

chr-pok: Yes, I thought about it but realized that there are subtle differences, for example the simulation task accepts `CircuitFromID, MEModelFromID, MEModelWithSynapsesCircuitFromID` whereas the circuit extraction task only `CircuitFromID`. Also, the internal variables that are set by this function are (slightly) different.

So yes, I think the shared functionality could and should be factored out at some point, but better in a separate PR.

obi_one/scientific/tasks/circuit_extraction.py
```diff
@@ -37,7 +44,9 @@ class CircuitExtractionScanConfig(ScanConfig):
     )
 
     class Initialize(Block):
-        circuit: Circuit | list[Circuit]
+        circuit: CircuitDiscriminator | list[CircuitDiscriminator] = Field(
+            title="Circuit", description="Parent circuit to extract a sub-circuit from."
+        )
         run_validation: bool = False
         do_virtual: bool | list[bool] = Field(
             default=True,
```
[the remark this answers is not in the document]
james-isbister: Ok, these are much better. I would suggest these slight iterations on the two descriptions:

1) Additionally extract virtual neurons which target the cells contained in the specified neuron set (together with their connectivity onto the specified neuron set).

2) Convert neurons which are outside of the specified neuron set, but which target the cells contained therein, into a new external population of virtual neurons (together with their connectivity onto the specified neuron set).

But these could be followed up in the next ticket

examples/C_forms/circuit_extraction/circuit_extraction_example.ipynb
james-isbister: Now I get this error, running this notebook with `run_validation=True`

[circuit_extraction_validation_error.txt](https://github.com/user-attachments/files/23318289/circuit_extraction_validation_error.txt)

obi_one/scientific/from_id/circuit_from_id.py
```diff
@@ -36,7 +36,7 @@ def stage_circuit(
                     )
 
                 circuit = Circuit(
-                    name=dest_dir.name,
+                    name=self.entity(db_client=db_client).name,
```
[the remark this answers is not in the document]
james-isbister: Cool, this works well!

obi_one/scientific/tasks/circuit_extraction.py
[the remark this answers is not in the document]
chr-pok: Thanks! Will take note and follow up in #347.

chr-pok:
Finally, I removed the `run_validation` option and made `do_virtual` and `create_external` non-scannable, as discussed. @james-isbister, would be great if you could have another look please!

Issue #426, opened 2025-10-17
(Connectivity metrics) Only put circuit id in the request body (not path parameter)


jankrepl:
For the connectivity metrics endpoint `circuit_id` is located at two places

1. Path parameter
https://github.com/openbraininstitute/obi-one/blob/e35c2b8109a24c38eab982a43ac4d6a7f30f24fe/app/endpoints/declared_endpoints.py#L401

2. Request body
https://github.com/openbraininstitute/obi-one/blob/e35c2b8109a24c38eab982a43ac4d6a7f30f24fe/obi_one/scientific/library/connectivity_metrics.py#L21

I suggest we only keep it in the request body - that is the correct approach AFAIK. And also, the current code actually does not use the `circuit_id` passed via the path parameter. So the PR will be a oneliner.

chr-pok:
The fix has been merged and deployed in staging (obi-one 2025.10.9).

<img width="1433" height="58" alt="Image" src="https://github.com/user-attachments/assets/6aa26d72-0e1c-4bd6-a3cf-b3b32de139b9" />

I think this issue can be closed.

Issue #374, opened 2025-09-25
Add endpoint for returning connection probabilities


chr-pok:
**Description:**
An endpoint should be added that returns the connection probabilities between groups of neurons in a given circuit, optionally grouped by a certain node property.

**Acceptance criterion:**
Endpoint implemented and tested.

chr-pok:
Worked on in the `obi-one` branch [conn_prob_endpoint](https://github.com/openbraininstitute/obi-one/tree/conn_prob_endpoint).

chr-pok:
Status update: 
- Implemented partial circuit download (required components only)
- Implemented connectivity computation (using connectome-manipulator)
- Added example notebook

Notes:
- Currently returns two connectivity metrics: connection probability and mean number of synapses per connection
- Valid property names/values for node selection and grouping need to be known (can be retrieved using the circuit metrics endpoint)
- Small changes in connectme-manipulator required to properly handle some edge cases

chr-pok:
Agreed on in today's AI meeting:
- Providing pre-/post-synaptic selection dictionaries through a POST endpoint
- AI agent is responsible for querying valid edge population names and group-by properties before calling this endpoint

chr-pok:
[PR#378](https://github.com/openbraininstitute/obi-one/pull/378) in `obi-one` has been created.

Another PR in `connectome-manipulator` is in preparation to properly handle the following special cases:
- Adding support for group-by to be empty, in which case no grouping will be applied (returning just a single value)
- Adding support for ignoring group-by in case the grouping property does not exist in one of the pre- or post-synaptic node population (e.g., virtual populations), in which case the grouping will only be applied to the population where it does exist (but it will still raise an error if the grouping property does not exist in either of the populations)

PR #470, opened 2025-10-30
Small fixes for connectivity metrics endpoint

chr-pok:
- Improved logic to handle empty input values
- Updated connectome-manipulator version requirement (with additional fixes)
- Removed `circuit_id` from path and kept only in the request body

Related issues: https://github.com/openbraininstitute/obi-one/issues/374, https://github.com/openbraininstitute/obi-one/issues/426


jankrepl (approved):
LGTM:)

PR #475, opened 2025-11-06
Rename scan and coordinate dump files

james-isbister:
Change:
* run_coordinate_instance.json -> obi_one_coordinate.json
* run_scan_config.json -> obi_one_scan.json


james-isbister:
I will check with Pavlo whether he relies on these names anywhere. I don't see anything in core-web-app, so I expect not

PR #476, opened 2025-11-08
Em synapse mapping

MWolfR:
From morphology with spines to Synaptome.


chr-pok (commented):
Just some small comments/questions.

But I had troubles actually running the example notebook since I did not know how to select proper input morphologies (the ones I randomly selected from the list did break it).

In general, it would be nice to add a title and some descriptions to the example notebook to help the user understand what's happening in the individual cells.

Issue #482, opened 2025-11-17
Always use the latest entitysdk in docker builds


jankrepl:
Old versions of entitysdk are "useless" because the underlying entitycore API has changed and might not be compatible.

The Dockerfile seems to be using the `uv sync --locked` which basically means the versions of dependencies will be taken from the `uv.lock`

https://github.com/openbraininstitute/obi-one/blob/cf7cf74a41a3af892602b96ca1401e9e632e8936/Dockerfile#L47

Note that at the time of writing this the fixed version of entitysdk is `0.8.7`, however the latest one is `0.10.0`

https://github.com/openbraininstitute/obi-one/blob/cf7cf74a41a3af892602b96ca1401e9e632e8936/uv.lock#L841-L843

Related to this: https://github.com/openbraininstitute/prod-synthesize-morphologies/issues/11

GianlucaFicarelli:
Note also that entitysdk could introduce breaking changes that can require some fixes in obi-one code, so it should be tested.
A possibility is to execute `uv lock --upgrade-package entitysdk`, but it should be done locally, not automatically in the CI.
Another possibility is to execute `make upgrade-deps` (that is an alias for running `uv lock --upgrade`) before making a release, so it will upgrade all the packages, including entitysdk.

GianlucaFicarelli:
I simple check that can be run in the CI is the following:
```
uv lock --upgrade-package entitysdk --locked
```
It will fail if there is a new version of entitysdk that's different from the version in uv.lock

GianlucaFicarelli:
A possible check is in https://github.com/openbraininstitute/obi-one/pull/489
You can see that the CI fails because there was a new release of entitysdk today.

PR #489, opened 2025-11-27
Check that entitysdk is at the latest version

GianlucaFicarelli:
Updated check-deps target to ensure entitysdk is latest.
Fix https://github.com/openbraininstitute/obi-one/issues/482

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ install:  ## Create a virtual environment
 	uv run python -m ipykernel install --user --name=obi-one --display-name "obi-one"
 
 compile-deps:  ## Create or update the lock file, without upgrading the version of the dependencies
-	uv lock
+	uv lock --upgrade-package entitysdk
 
 upgrade-deps:  ## Create or update the lock file, using the latest version of the dependencies
 	uv lock --upgrade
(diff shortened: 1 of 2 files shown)


james-isbister:
Thank you both. I've been in meetings since 10am so just having a chance to look.

Just to check I understand, we can run `make compile-deps` if the CI fails on `make check-deps` due to the entitysdk version? And this will override any version specified in `pyproject.toml` for entitysdk?

GianlucaFicarelli:
> Just to check I understand, we can run `make compile-deps` if the CI fails on `make check-deps` due to the entitysdk version? And this will override any version specified in `pyproject.toml` for entitysdk?

Yes you can run `make compile-deps` to update just entitysdk to the latest version.
You should also periodically run `make upgrade-deps` to update all the packages (once per month for example, but it's up to you). However, this requires more attention because there can be breaking changes in more packages.

The version of entitysdk specified in `pyproject.toml` is `entitysdk>=0.8.7` at the moment (no upper bound), so the command above will update it to the latest version.
If you set a specific version or an upper bound in pyproject (like `entitysdk>=0.8.7,<1.0.0`) then even `make compile-deps` and `make upgrade-deps` cannot upgrade entitysdk to a higher version. You should use that only if you explicitly want to avoid updating entitysdk for any reason.

Issue #437, opened 2025-10-21
Use ConnectomeManipulator parameter config to add synapse parameters to MICrONS MEModel/Morphology


james-isbister:

james-isbister:
I think it would be better if adding synaptic physiology parameters to an anatomical model of one or several neurons was a separate Task to begin with. Especially if we plan to go into details of how the parameters are selected. Once things mature we could think about how Tasks might be combined

Unless this Task is really just adding some default physiology values (with no option to parameterize). If we add something quick to the configuration schema we then have to maintain compatibility so I don't think its something we should rush. It's obviously its own very complex topic in itself, and is something general we might want to do outside this context anyway

Either way we can maybe review and then merge the PR for this task separately first, as its already quite big?

What do you think @MWolfR @chr-pok

james-isbister:
`Does it update the assets of the existing entity or create a new entity.`
Yes, that's the question. I agree if there are no existing synaptic physiology parameters then it would be nice to just update the assets of the existing entity, although it might ideally be an option - as the user might want to try different physiology parameters for the same MEModelWithSynapses. In a scan over synaptic parameters - which would be nice to do I think - you really can't just update the assets so maybe its safer to in general always create a new entity. This would be very similar to how we had separate sscx anatomy and physiology models, with the possibility of multiple physiology models deriving from the anatomy so maybe its a useful pattern.

I'll copy the three bullet points into the related ticket and reply to them!

evakennyobi:
Can you add an end date here please @chr-pok

chr-pok:
[PR#490](https://github.com/openbraininstitute/obi-one/pull/490) with my initial implementation proposal created and ready for review/discussions.

chr-pok:
[PR#490](https://github.com/openbraininstitute/obi-one/pull/490) still under review by @james-isbister and @MWolfR.

evakennyobi:
This needs an updated end date please @chr-pok 
I think the PR is still needing review from @MWolfR

PR #490, opened 2025-11-28
Synapse parameterization task

chr-pok:
Implement of an obi-one task which runs a physiological parameterization of an anatomical synaptome. It uses a synaptome (with or without synaptic physiology) from entitycore as an input and generates a (re-)parameterized synaptome as an output which will be registered as a new entity in entitycore.

Implementation details:
- Added a random seed for reproducibility
- Added an explicit overwrite option for re-parameterization
- Added explicit node property by which pathways are defined
- Added entity cache option: If the staged synaptome already exists in the cache, it will be copied to the output folder. Otherwise, it will be directly staged in the output folder and modified in-place.
- Refactored code into commonly used helper functions, to be shared between EM synapse mapping and parameterization tasks
- Changed compression to `tarfile` for consistency with what we use with all other circuits
- Added a derivation link of type “circuit_rewiring” from the input to the parameterized output synaptome; perhaps we would need a new derivation type for that?

Related issue: https://github.com/openbraininstitute/obi-one/issues/437


chr-pok:
@MWolfR, I noticed  that the input synaptomes (produced by the EM synapse mapping task) are missing a node_sets.json file but which is referenced in the circuit_config.json. It would be good to add that file (even if empty) since the missing file breaks certain bluepysnap operations.

ilkilic:
oups, it wasn't my intention to close that PR, let me fix it

james-isbister (approved):

PR #498, opened 2025-12-05
Add compartment_set stimuli support

ilkilic:
## Summary

This PR adds support for morphology-derived stimulus targeting through `MorphologyLocations` and SONATA `compartment_set` materialization.

Implemented:
- `MorphologyLocations` blocks in simulation configs
- stimulus targeting via generated compartment sets
- morphology-location materialization pipeline
- dedicated `Targeting` UI group
- backward compatibility for legacy `*SomaticStimulus` aliases
- integration with the new typed neuron-set references

### Morphology targeting

Stimuli can now target:

- a neuron set
- an explicit `CompartmentSet`
- generated `MorphologyLocations`

`MorphologyLocations` uses the new typed biophysical neuron-set references to select the cells on which locations are generated.

### Materialization pipeline
- Morphology locations are materialized into SONATA `compartment_set` entries during simulation generation
- Generated compartment sets are exported into `compartment_sets.json`
- Multiple generated compartment sets are consolidated into a single SONATA file

### UI/schema
- Added dedicated `Targeting` section
- Added `MorphologyLocations` references to supported stimuli
- `section_types` support exists in the backend but is currently hidden from the generated UI pending follow-up work

### Compatibility
- Added deserialization support for legacy `*SomaticStimulus` aliases

## Follow-up work

Required follow-up:
- dynamic morphology section-type selection UI (#767)

Additional enhancements tracked separately:
- deterministic morphology-location blocks (#769)
- interactive morphology location selection in the morphology viewer (#770)
- exposing generated `compartment_sets.json` in Simulation input files (#768)

## Tests

Added coverage for:
- morphology-location materialization
- compartment-set export
- morphology path resolution
- legacy stimulus deserialization
- morphology-location validation


obi_one/scientific/tasks/generate_simulations/materialize_locations.py
```diff
@@ -0,0 +1,63 @@
...
+    CompartmentSet,
+    build_compartment_set_for_neuron_set,
+)
+from obi_one.scientific.library.circuit import Circuit
+from obi_one.scientific.unions.unions_compartment_sets import CompartmentSetReference
+
+
+def materialize_locations_to_compartment_sets(
+    *,
+    form: Any,
```
james-isbister: I think we can use 'single_config' rather than form

obi_one/scientific/tasks/generate_simulations/config/base.py
```diff
@@ -50,12 +52,34 @@
 DEFAULT_TIMESTAMPS_NAME = "Default: Simulation Start (0 ms)"
 DEFAULT_DISTRIBUTION_NAME = "Default: Exp, scale 50 ms, 20 Hz"
 
+_STIMULUS_TYPE_ALIASES = {
```
james-isbister: It would be nice to change the names of these Python classes but the biggest issue is the duplication functionality in the frontend. If someone duplicates an old config in the frontend which has the old names, then it will be invalid in the frontend.

Two approaches we might take are:
- Renaming the old classes and creating aliases which are included in the stimulus union but are hidden (i.e. defining class ConstantCurrentClampSomaticStimulus(ConstantCurrentClampStimulus): pass
- Migrating the old configs

I think migrating the configs in entitycore would be the best longer term approach. In this PR I would suggest we revert to the old names with Somatic, and deal with this in a follow up, once everything is working in the frontend.

obi_one/scientific/blocks/stimuli/stimulus.py
```diff
@@ -119,6 +125,55 @@ class ContinuousStimulusWithoutTimestamps(BaseStimulus):
         },
     )
 
+    compartment_set: CompartmentSetReference | None = Field(
```
james-isbister: In the new_neuron_sets branch we have changed the REFERENCE_TYPE key in the schema of the ScanConfig block_dictionary to  represent in the schema that it holds multiple [REFERENCE_TYPES](https://github.com/openbraininstitute/obi-one/blob/517902ea48cf9998407f469201810663935dbc94/obi_one/scientific/tasks/generate_simulations/config/circuit.py#L81) in a list, note the plural.

The REFERENCE_TYPE in the schema of blocks has also changed to a list, plural [REFERENCE_TYPES](https://github.com/openbraininstitute/obi-one/blob/517902ea48cf9998407f469201810663935dbc94/obi_one/scientific/blocks/stimuli/stimulus.py#L118)

This means that a reference_type parameter of a block can now "hunt" for block_references of different types across multiple block dictionaries, and that each block_dictionary can serve different types of block reference.

This means that we can have a single dropdown for selecting the `target` of say a stimulus, which will show references from neuron_set and compartment_set block_dictionaries

I think we can rebase this PR to the new_neuron_sets branch to use this functionality. Although I'm not sure when the best time to merge will be as @chr-pok is also working on that

This removes the need for compartment_set and locations references here, and we can have a single `target`, although for now I would keep the name neuron_set for backwards compatibility, and we can think about migration in a separate PR once this is all working in the frontend.

obi_one/scientific/blocks/morphology_locations/base.py
```diff
@@ -6,23 +6,43 @@
 from pydantic import Field, model_validator
 
 from obi_one.core.block import Block
+from obi_one.core.schema import SchemaKey, UIElement
 
 
 class MorphologyLocationsBlock(Block, abc.ABC):
     """Base class representing parameterized locations on morphology skeletons."""
```
james-isbister, line 26: I think this should take a neuron_set reference, and then locations should be calculated across all neurons in the neuron set

james-isbister (commented):
Looking good @ilkilic Thanks and sorry for the delay

Here are my initial higher-level comments. I can look more at some of the details once we've decided on an approach for the existing comments

obi_one/scientific/tasks/generate_simulations/materialize_locations.py
ilkilic: Done.

obi_one/scientific/tasks/generate_simulations/config/base.py
ilkilic: Sounds good. I have reverted to the old names with Somatic.

obi_one/scientific/blocks/stimuli/stimulus.py
ilkilic: Thanks, agreed. I did not go in that direction yet because I wasn’t sure whether new_neuron_sets was stable enough to depend on already. Do you think it would make sense to rebase onto it now or would it be better to wait until that work is merged?

obi_one/scientific/blocks/morphology_locations/base.py
ilkilic: Done. MorphologyLocationsBlock now has a neuron_set reference and location materialization uses that target.

obi_one/scientific/blocks/stimuli/stimulus.py
ilkilic: Implemented after rebasing. Stimuli now use the existing `neuron_set` field as a single target selector, supporting neuron sets, compartment sets and morphology locations.

obi_one/scientific/unions/unions_morphology_locations_ref.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any, ClassVar
```
james-isbister: Maybe the unions are references for MorphologyLocations could be put in the same file within the new unions_and_references directory to be consistent with the other types

obi_one/scientific/blocks/morphology_locations/base.py
```diff
@@ -6,23 +6,56 @@
...
         default=1,
-        title="Number of locations",
-        description="Number of locations to generate on morphology",
+        title="Number of Locations",
+        description="Number of locations to generate on morphology.",
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.INT_PARAMETER_SWEEP,
+        },
     )
+
```
james-isbister: What if a neuron doesn't have these section types? Maybe they would be fetched based on the neuron set in future, but don't need for now.

james-isbister: Also I think the typing could be something like this:

> from typing import Literal
> SectionType = Literal["axon", "basal", "apical"]
> SectionTypes = tuple[SectionType, ...] | list[tuple[SectionType, ...]]
>
> section_types: SectionTypes = Field(
>     default=("axon", "basal", "apical"),
>     title="Section Types",
>     description=(
>         "Valid neurite section types to generate locations on: "
>         "axon, basal (basal dendrite), apical (apical dendrite). "
>         "Use a tuple for one selection, e.g. ('basal', 'apical'), "
>         "or a list of tuples for parameter scans."
>     ),
>     json_schema_extra={
>         SchemaKey.UI_HIDDEN: True,
>     },
> )

Or: 

> from typing import Literal
> SectionType = Literal[2, 3, 4]
> SectionTypes = tuple[SectionType, ...] | list[tuple[SectionType, ...]]
> 
> section_types: SectionTypes = Field(
>     default=(2, 3, 4),
>     title="Section Types",
>     description=(
>         "Valid neurite section types to generate locations on: "
>         "2: axon, 3: basal dendrite, 4: apical dendrite. "
>         "Use a tuple for one selection, e.g. (3, 4), "
>         "or a list of tuples for parameter scans."
>     ),
>     json_schema_extra={
>         SchemaKey.UI_HIDDEN: True,
>     },
> )

obi_one/scientific/blocks/stimuli/stimulus.py
```diff
@@ -142,13 +177,20 @@ def config(
...
+                (CompartmentSetReference, MorphologyLocationsReference),
+            )
+            and (
+                self.neuron_set.block.get_neuron_set_population_type()
+                not in {
+                    NeuronSetPopulationType.BIOPHYSICAL,
+                    NeuronSetPopulationType.POINT,
+                    NeuronSetPopulationType.NONVIRTUAL,
+                }
+            )
```
james-isbister: This check is maybe redundant now we have the typed population type specific neuron set references

obi_one/scientific/blocks/compartment_sets.py
```diff
@@ -0,0 +1,165 @@
...
+class CompartmentLocation(BaseModel):
+    node_id: int = Field(ge=0)
+    section_id: int = Field(ge=0)
+    offset: float = Field(ge=0.0, le=1.0)
+
+
+class CompartmentSet(Block):
+    """SONATA-compatible compartment_set block."""
+
+    population: str = Field(
```
james-isbister: Here we can use the pattern of selecting a population within the BiophysicalPopulationNeuronSet, so we have a dropdown with only biophysical populations

obi_one/scientific/tasks/generate_simulations/task/task.py
```diff
@@ -322,6 +365,22 @@ def _default_neuron_set_ref(self) -> ALL_NEURON_SETS_REFERENCE_UNION:
     NEURON SETS HAVE TYPES (BIOPHYSICAL, POINT, ETC.)
     """
 
+    @staticmethod
```
james-isbister: Does this need to be passed as an argument?

obi_one/scientific/tasks/generate_simulations/task/task.py
```diff
@@ -428,6 +487,29 @@ def _resolve_neuron_sets_and_write_simulation_node_sets_file(self) -> None:
...
+        # 4. If the config contains compartment_sets, collect them and write a
+        # compartment_sets.json file next to the node sets so stimuli referencing
+        # named compartment sets will resolve correctly.
+        if self._materialized_compartment_sets:
+            compartment_sets_dict: dict = {}
+
+            for cs_key, cs_block in self._materialized_compartment_sets.items():
+                if cs_key != cs_block.block_name:
+                    msg = "Materialized compartment set name mismatch."
+                    raise OBIONEError(msg)
```
james-isbister, line 502: If we create _materialized_compartment_sets, then would this error ever be triggered?

obi_one/scientific/tasks/generate_simulations/task/task.py
```diff
@@ -428,6 +487,29 @@ def _resolve_neuron_sets_and_write_simulation_node_sets_file(self) -> None:
         )
         self._sonata_config["node_sets_file"] = self.NODE_SETS_FILE_NAME
 
+        # 4. If the config contains compartment_sets, collect them and write a
+        # compartment_sets.json file next to the node sets so stimuli referencing
+        # named compartment sets will resolve correctly.
+        if self._materialized_compartment_sets:
```
james-isbister, line 495: This doesn't really match the purpose of the function name "_resolve_neuron_sets_and_write_simulation_node_sets_file"

Maybe this could be combined with the code in materialize_location_targets with a name that includes the writing, if it makes sense?

obi_one/scientific/tasks/generate_simulations/materialize_locations.py
```diff
@@ -0,0 +1,68 @@
...
+        neuron_set_ref = getattr(locations_block, "neuron_set", None)
+        if neuron_set_ref is None:
+            msg = (
+                f"Locations block '{locations_block.block_name}' referenced by stimulus "
+                f"'{stimulus.block_name}' has no neuron_set."
+            )
+            raise ValueError(msg)
+
+        comp_set_name = f"{stimulus.block_name}__locations"
+
```
james-isbister, line 42: I'm not sure we need the block type CompartmentSet, and the CompartmentSetUnion, CompartmentSetReference etc.

The MorphologyLocation blocks could just have a function which resolves to the sonata entry. Wdyt?

james-isbister (commented):
Nice work! Here are my comments

obi_one/scientific/unions/unions_morphology_locations_ref.py
ilkilic: Done, I moved MorphologyLocationsReference into unions_morphology_locations.py so the union and reference live together, consistent with the other union/reference modules.

obi_one/scientific/tasks/generate_simulations/task/task.py
ilkilic: Good point, removed the redundant circuit argument and uses self._circuit directly.

obi_one/scientific/tasks/generate_simulations/task/task.py
ilkilic: Correct, this case should not happen in the normal task flow. I removed it.

obi_one/scientific/tasks/generate_simulations/task/task.py
ilkilic: Good point. I moved the compartment-set writing into a separate `_write_materialized_compartment_sets_file`() method and call it explicitly after node sets are written.

obi_one/scientific/tasks/generate_simulations/materialize_locations.py
ilkilic: Good point, I agree. I refactored this so compartment sets are no longer public blocks/references. MorphologyLocations remains the user-facing API and materialization now produces the SONATA compartment_sets.json entry internally. If we later want users to define compartment sets directly we can add that as a separate advanced block with a clearer workflow.

obi_one/scientific/tasks/generate_simulations/task/task.py
james-isbister: Sorry, this is my mistake. The comment should have been higher up. I was referring to morphology_loader=self._load_morphology,

Is it possible that we would use a different loading function? Or could avoid passing the function here?

I think the passing the circuit in the function below as before was nice actually

obi_one/scientific/blocks/morphology_locations/base.py
ilkilic: Good point. I kept the MorphIO numeric values but restricted them to 2, 3, and 4. For morphologies missing a selected type, I think the follow-up feature/morphology-section-type-selection branch (https://github.com/openbraininstitute/obi-one/pull/822) should handle that by fetching available section types.

obi_one/scientific/tasks/generate_simulations/task/task.py
ilkilic: Ah sorry, yes, that makes sense. I removed the morphology_loader callback and now the materialization helper uses the passed circuit directly to load morphologies.

obi_one/scientific/tasks/generate_simulations/task/task.py
james-isbister: I still see the check on line 499

examples/obi_one/scientific/tasks/generate_simulations/NEURON/circuit/run_circuit_simulation.ipynb
james-isbister, line 1: Could just clear the notebook I think

obi_one/scientific/tasks/generate_simulations/materialize_locations.py
```diff
@@ -0,0 +1,55 @@
...
+        return materialized
+
+    for stimulus in single_config.stimuli.values():
+        target_ref = getattr(stimulus, "neuron_set", None)
+        if not isinstance(target_ref, MorphologyLocationsReference):
+            continue
+
+        locations_block = target_ref.block
+
+        neuron_set_ref = getattr(locations_block, "neuron_set", None)
```
james-isbister, line 33: The default is None (i.e. "Default: All Biophysical Neurons"), so this is very possible and should not raise an error

obi_one/scientific/tasks/generate_simulations/task/task.py
ilkilic: Ah sorry, should be removed now.

examples/obi_one/scientific/tasks/generate_simulations/NEURON/circuit/run_circuit_simulation.ipynb
ilkilic: Done in the last commit

obi_one/scientific/tasks/generate_simulations/materialize_locations.py
james-isbister: This also brought to mind another issue. For large circuits, this is probably too computationally and/or memory expensive to be run on the service. Maybe this is where simulation generation would eventually need to be run on the launch system? Obviously not for this ticket.

Probably for now we should hide the morphology locations blocks if the circuit is larger than small microcircuit to begin? Using the usability feature? and a corresponding bool in the mapped circuit properties? Have you used that pattern before? I can explain otherwise?

james-isbister (commented):
Getting there I think. One thing about running campaign generation on the service just occurred to be though. 

I also replied to a small comment outside the review.

james-isbister (dismissed):
LGTM!

obi_one/scientific/tasks/generate_simulations/config/neuron/neuron_me_model.py
```diff
@@ -88,12 +91,14 @@ class Initialize(NeuronSimulationScanConfig.Initialize):
...
         SchemaKey.GROUP_ORDER: [
             BlockGroup.SETUP_BLOCK_GROUP,
+            BlockGroup.TARGETING_BLOCK_GROUP,
             BlockGroup.STIMULI_RECORDINGS_BLOCK_GROUP,
             BlockGroup.CIRCUIT_MANIPULATIONS_GROUP,
             BlockGroup.EVENTS_GROUP,
         ],
         SchemaKey.DEFAULT_BLOCK_REFERENCE_LABELS: {
             TimestampsReference.__name__: DEFAULT_TIMESTAMPS_NAME,
+            MorphologyLocationsReference.__name__: "Default: None",
```
james-isbister: `Default: No locations` is maybe better?

obi_one/scientific/tasks/generate_simulations/config/base.py
```diff
@@ -52,6 +52,7 @@ class BlockGroup(StrEnum):
 
     SETUP_BLOCK_GROUP = "Setup"
     STIMULI_RECORDINGS_BLOCK_GROUP = "Stimuli & Recordings"
+    TARGETING_BLOCK_GROUP = "Targeting"
```
james-isbister: I think we can group Neuron Sets and Morphology Locations into a single BlockGroup but I'm not sure "Circuit Components" or "Targeting". Maybe we can think of a different term in a follow up PR

james-isbister (dismissed):
LGTM! And works well in the frontend!

I think we can merge the other branch into this one, and then merge this to main after the staging freeze if there is one.

I would make the following textual changes in a separate PR so we can consider the changes clearly:
- Change: MorphologyLocationsReference.__name__: "Default: No Locations"
- Find a better BlockGroup name for grouping Neuron Sets and Morphology Locations
- Improve the block and parameter descriptions of the Morphology Locations to explain how each of them work. I hadn't looked at the Morphology Locations for a while, so I was unclear how the different combinations of clusters and locations would lead to different outcomes (and what each of those parameters meant exactly)
- Add units for some of the distance based parameters of the MorphoogyLocations Blocks
- Add title class variables to the MorphologyLocations blocks with spaces. Currently the class name without spaces is displayed in the frontend

Would be good to also test the performance in staging once its live, and think about expanding to larger circuits.

obi_one/scientific/tasks/generate_simulations/config/base.py
ilkilic: I tried `Targets` in the follow-up PR. It groups Neuron Sets and Morphology Locations together, wdyt?

obi_one/scientific/blocks/morphology_locations/clustered.py
james-isbister: I think the check `if self.n_clusters < 1:` is no longer needed now `n_clusters: PositiveInt`

james-isbister: This still raises an unknown error in the UI. Not sure why the message isn't displayed

Image

james-isbister (commented):
Looking good in the UI. Just one final thing to look at + a minor change

obi_one/scientific/blocks/morphology_locations/clustered.py
ilkilic: I had a quick look in core-web-app and I wonder if the issue might be there. It looks like the coordinate-count handler reads message.detail while the backend returns the validation message in details[0].msg.

@bilalesi would you mind having a quick look? Do you think the coordinate-count handler should use details[0].msg here as well? Thank you!

ilkilic: I updated it in https://github.com/openbraininstitute/obi-one/pull/932

james-isbister (approved):

PR #512, opened 2025-12-16
JSONSchema GUI definition spec

g-bar:
Relates to: https://github.com/openbraininstitute/prod-platform-architecture/issues/173


gui-definition-spec/gui-definition.md
```diff
@@ -0,0 +1,320 @@
+# Specification for JSONSchema GUI definition
+
+## Forms
```
james-isbister: We can use "ScanConfig" or "Config" through-out rather than "Form" which is no longer used

gui-definition-spec/gui-definition.md
```diff
@@ -0,0 +1,320 @@
...
+
+Valid `ui_element` values are:
+
+All elements:
+
+- `null`
+
+Root elements:
+
+- `root_block`
```
james-isbister: We will need a third case (maybe for a separate PR) for the intermediate case when the type of the root_block needs to be selected from a union. Maybe "choise_root_block" or something like that

gui-definition-spec/gui-definition.md
```diff
@@ -0,0 +1,320 @@
...
+        ] = None
+```
+
+### UI design
+
+<img src="designs/reference.png"  width="300" />
+
+## Predefined neuron set
+
+ui_element: `predefined_neuronset`
```
james-isbister: This is an example of a more general case where we want to fetch a list of valid strings (but it could be other types later), which are specific to the primary entity of the config. These would be fetched through an obi-one endpoint, and then presented as a dropdown of options for the parameter. 

It would be good to create a general ui_element for these cases, where we specify the endpoint and key. The endpoint returns a dictionary of multiple key: list pairs (it is more efficient to do it for multiple keys in one call), and the ui_element displays the values of the list for the specified key

We could think about this generalisation in the new year though perhaps

james-isbister (commented):
Thanks Gil, this is great. I think we’re aligned. Here are my initial comments which are mostly about details. I might add some further comments throughout the day

gui-definition-spec/gui-definition.md
james-isbister: This is related to the CircuitExtraction comment from @chr-pok below

gui-definition-spec/gui-definition.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Reference
+
+ui_element: `reference`
+
+- Should have an `anyOf` property.
+  - First element should be an `object` with string fields `block_name` and `block_dict_name`.
+  - First element should have a `type` property with a `const` value. (e.g `NeuronSetReference`).
+  - Second element should be `null`.
+  - **Order matters: null should always come last**
```
james-isbister: I think we should specify the string to be displayed in place of None for Reference type as a property of the ScanConfig, as we currently do with `default_block_reference_labels`.

Also we should describe the behaviour if there isn't a block dictionary containing references of the specified type i.e. don't show the field

Unfortunately, we can't use `ui_element = null` for the block_element in this case, as its the ScanConfig which the block is used in which determines whether the reference field should be shown.

gui-definition-spec/gui-definition.md
g-bar: Sure will make the change.

gui-definition-spec/gui-definition.md
```diff
@@ -0,0 +1,320 @@
...
+
+Reference schema [reference](reference_schemas/reference.json)
+
+### Example Pydantic implementation
+
+_Note: Placing `None` inside `Annotated` ensures the title and description on the parameter and not the inner reference object_
+
+```py
+class Block:
+    node_set: Annotated[
```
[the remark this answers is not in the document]
g-bar: Not sure I follow, can you show an example?

gui-definition-spec/gui-definition.md
g-bar: Regarding this, I want to avoid implicit rules on the frontend to simplify. Also I want to avoid conditional logic for hiding elements. I want consistency.

But actually, you can have that logic on the BE.

Instead of exposing the `default_block_reference_labels` dict, you can keep it internal and then in the parameters' definitions  you set `ui_hidden = 'referenceType' in  'default_block_reference_labels'`

same for the labels.

gui-definition-spec/gui-definition.md
```diff
@@ -0,0 +1,320 @@
...
+## Block elements
+
+Block elements are properties of blocks, they (as _root elements_) must have a `ui_element` property.
+The parents of block elements must be blocks, never forms.
+
+They should contain a `title` and a `description`.
+
+## Input
+
+ui_element: `input`
```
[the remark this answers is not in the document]
chr-pok: So shouldn't we have different input types then, like `string_input`, `int_input`, `float_input`, etc. in the same way as for sweeps but for entering single input values?

gui-definition-spec/gui-definition.md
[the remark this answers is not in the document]
chr-pok: This is related to @james-isbister's comment https://github.com/openbraininstitute/obi-one/pull/512#discussion_r2626199980.

gui-definition-spec/gui-definition.md
```diff
@@ -0,0 +1,320 @@
...
+### Example Pydantic implementation
+
+```py
+
+class CircuitFromId(OBIBaseModel):
+    id_str: str = Field(description="ID of the entity in string format.")
+
+
+class Block:
+    circuit: CircuitFromID = Field(
```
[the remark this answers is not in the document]
g-bar: For now let's keep it as is.

gui-definition-spec/gui-definition.md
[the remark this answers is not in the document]
g-bar: Yes, this works. I'll go with that.

gui-definition-spec/gui-definition.md
james-isbister: So the idea was that `default_block_reference_labels` is different in different `ScanConfigs` and the Block might only be defined once in a separate file.

For example, the Block "SomaticStimulus" which has a parameter `neuron_set: NeuronSetReference` is just defined once in blocks/stimulus.py:
https://github.com/openbraininstitute/obi-one/blob/5bafa9d92d0c73546074108d3eff56df7b243d27/obi_one/scientific/blocks/stimulus.py#L61

But when SomaticStimulus is used in different ScanConfigs (as a parent class, but that's not relevant), we might define 1) a different default string, or 2) that the field `neuron_set` does not need to appear because there are no NeuronSets to reference.

I don't think the suggestion covers this. I get that the rule is two implicit though

One slightly more explicit solution could be a rule that for 2) we can specify:
default_block_reference_labels =  {
                NeuronSetReference.__name__: None
} 

in which case we don't show any NeuronSetReference fields.

This would be my preference. Another option would for now would be that we just enforce that a default string is specified but I think it's nice that the field disappears when there is no alternative to None

gui-definition-spec/gui-definition.md
```diff
@@ -0,0 +1,334 @@
...
+Reference schema [entity_property_dropdown](reference_schemas/entity_property_dropdown.json)
+
+### Example Pydantic implementation
+
+```py
+CircuitNode = Annotated[str, Field(min_length=1)] # Required in the schema
+NodeSetType = CircuitNode | list[CircuitNode] # list[] not required
+
+class Block:
+    node_set: Annotated[
```
james-isbister: Would be good to move the Field to after the equals sign as it is in the other PR:

> node_set: NodeSetType = Field(
>         ui_element="entity_property_dropdown",
>         entity_type=EntityType.CIRCUIT,
>         property=CircuitPropertyType.NODE_SET,
>         title="Node Set",
>         description="Name of the node set to use.",
>         min_length=1,
>     )

james-isbister (approved):

PR #521, opened 2026-01-12
number neurons entity property added for creation and update

james-isbister:
Needed to fix campaign generation in dev before the merge to production

- number_neurons entity property added to Simulation creation and update
- Also move simulation_config upload last for UI which checks its existence as sign generation was successful (short term solution without activities for the generation of each simulation)

### DIFF
--- a/obi_one/scientific/tasks/generate_simulation_task.py
+++ b/obi_one/scientific/tasks/generate_simulation_task.py
@@ -308,12 +308,13 @@ class GenerateSimulationTask(Task):
     def _update_simulation_number_neurons(self, db_client: entitysdk.client.Client | None) -> None:
         if db_client:
             if hasattr(self.config, "neuron_sets"):
-                number_neurons = self._neuron_set_definitions[
+                neuron_set_definition = self._neuron_set_definitions[
                     self.config.initialize.node_set.block_name
                 ]
             else:
-                number_neurons = len(self._neuron_set_definitions[DEFAULT_NODE_SET_NAME]['node_id'])
+                neuron_set_definition = self._neuron_set_definitions[DEFAULT_NODE_SET_NAME]
 
+            number_neurons = len(neuron_set_definition["node_id"])
             db_client.update_entity(
                 entity_id=self.config.single_entity.id,
                 entity_type=entitysdk.models.Simulation,


obi_one/scientific/tasks/generate_simulation_task.py
```diff
@@ -297,6 +297,18 @@ def _resolve_neuron_sets_and_write_simulation_node_sets_file(self) -> None:
         )
         self._sonata_config["node_sets_file"] = self.NODE_SETS_FILE_NAME
 
+    def _update_simulation_number_neurons(self, db_client: entitysdk.client.Client | None) -> None:
+        if db_client:
+            db_client.update_entity(
+                entity_id=self.config.single_entity.id,
+                entity_type=entitysdk.models.Simulation,
+                attrs_or_entity={
+                    "number_neurons": self.config.initialize.node_set.block.get_neuron_set_size(
```
chr-pok: This will resolve the neuron set a second time after it has been added to the circuit which may not be very efficient.

And there is a risk that it may even resolve differently in case we change the behavior in `add_node_set_definition_to_sonata_circuit` w.r.t. `default_population_name` but forget to update `get_neuron_set_size` accordingly.

chr-pok: >Being able to access the ids relies on force_resolve_ids=True though, so if we switch to using the expressions again in future, we wouldn't have the ids. If I've understood correctly

True, but I think we will never switch back anyway since this is the only way to keep clear control over resolving neuron sets in a single population which does not work with symbolic notation.

chr-pok: >Using a cache of the _ids allows us to calculate the size of the neuron set even for expressions, which could still be used in other contexts (not just our SimulationGenerationTask)

Yes, this could be useful indeed.

james-isbister: Shall I go for the cache, then? I think we can always rework this when we add the population dropdown if needed

james-isbister: `Just note that _resolve_ids() takes the population as input. Which means that the cache cannot be a single value but probably will have to be a dict with the population name as key.`

Not sure I understand this. `get_neuron_ids()` which is used by `get_neuron_set_size()` takes only a single population as input

chr-pok: >Could you remind me of the reason for this. Just in the node sets file you can't define a node set as a combination of symbol notation and a population name?

Yes, exactly, this is not supported in SONATAN unfortunately. A population name can only be used standalone or in combination with node IDs, but not in combination with other node properties. At least not in a single expression. You could create separate node sets and combine them in a new node set of course, but this is not what we want for defining a single neuron set.

chr-pok: >Shall I go for the cache, then? I think we can always rework this when we add the population dropdown if needed

Yes, I think we can go for it but it will probably take longer than my proposed solution from the beginning. Since we need to carefully think about when to update and invalidate the cache and test it probably.

chr-pok: Yeah, from the point of view of the simulation campaign this assumption is certainly true. But not if we add this functionality to the NeuronSet base class in general. That's why I was trying to keep most changes limited to the simulation campaign generation task in my suggestion.

james-isbister:
Thanks for the comments and discussion! Good to think it through!

PR #526, opened 2026-01-14
validate obj files

dkeller9:
This validates the obj files to be uploaded. A unit test function and ipynb test are included.

### DIFF
--- a/obi_one/scientific/unions/unions_stimuli.py
+++ b/obi_one/scientific/unions/unions_stimuli.py
@@ -6,7 +6,6 @@ from obi_one.core.block_reference import BlockReference
 from obi_one.scientific.blocks.stimuli.ornstein_uhlenbeck import (
     OrnsteinUhlenbeckConductanceSomaticStimulus,
     OrnsteinUhlenbeckCurrentSomaticStimulus,
-    RelativeOrnsteinUhlenbeckConductanceSomaticStimulus,
     RelativeOrnsteinUhlenbeckCurrentSomaticStimulus,
 )
 from obi_one.scientific.blocks.stimuli.stimulus import (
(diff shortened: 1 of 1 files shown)


app/endpoints/mesh_validation.py
```diff
@@ -0,0 +1,220 @@
...
+    """Helper to raise ValueError for mesh loading failures."""
+    msg = f"Failed to load OBJ file: {error}"
+    raise ValueError(msg) from error
+
+
+def validate_mesh_reader(mesh_file_path: str) -> trimesh.base.Trimesh | trimesh.Scene:
+    """Try all MESH readers. Succeed if at least one works."""
+    try:
+        # trimesh.load can return a Trimesh object or a Scene
+        mesh = trimesh.load(mesh_file_path, file_type="obj")
```
marwan-abdellah: @dkeller9 We might need to load the mesh with pyvista. I think it is much faster than trimesh, but we can give it a try to confirm if it scales with large meshes.

marwan-abdellah: Screenshot from 2026-01-16 14-29-06

app/endpoints/mesh_validation.py
```diff
@@ -0,0 +1,220 @@
...
+            L.warning(f"Failed to delete temp MESH file: {e}")
+
+
+def validate_mesh_file(
+    file: Annotated[UploadFile, File(description="MESH file to upload (.mesh)")],
+    background_tasks: BackgroundTasks,
+) -> MESHValidationResponse:
+    """Validates an uploaded .mesh file using registered readers."""
+    file_extension = pathlib.Path(file.filename).suffix.lower() if file.filename else ""
+    if file_extension != ".obj":
```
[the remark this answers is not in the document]
dkeller9: Yes, should be obj file.

app/endpoints/mesh_validation.py
dkeller9: Okay, swapped to pyvista instead.

PR #530, opened 2026-01-17
Add to numeric value schema check

james-isbister:
* Logic to choose an appropriate value for numeric schema check (as using 1.0 failed for a skeletonization task parameter with constraints 0.1 < x < 0.5
*  Logical checks on default, minimum, maximum values
* Check list[numeric] is second value in union with matching constraints as first numeric in union

* Note: Need to return to solution for handling exclusive min/max values

### DIFF
--- a/scripts/validate_block.py
+++ b/scripts/validate_block.py
@@ -64,13 +64,13 @@ def determine_minimum_valid_numeric_value(schema: dict) -> float | int:
         minimum = single_type.get("minimum", None)
         if minimum is None:
             minimum = single_type.get("exclusiveMinimum", -sys.maxsize)
-            minimum = minimum + math.ulp(minimum)
+            minimum = minimum + 1
 
 
         maximum = single_type.get("maximum", None)
         if maximum is None:
             maximum = single_type.get("exclusiveMaximum", sys.maxsize)
-            maximum = maximum - math.ulp(maximum)
+            maximum = maximum - 1
 
     elif single_type.get("type") == "number":
         minimum = single_type.get("minimum", None)


scripts/validate_block.py
```diff
@@ -54,17 +54,124 @@ def validate_string_param(schema: dict, param: str, ref: str) -> None:
...
+    minimum = single_type.get("minimum")
+    maximum = single_type.get("maximum")
+    exclusive_minimum = single_type.get("exclusiveMinimum")
+    exclusive_maximum = single_type.get("exclusiveMaximum")
+
+    
+    # Logical checks for min/max consistency
+
+    # If exclusiveMin and exclusiveMax, exclusiveMax should be greater than exclusiveMin
+    if exclusive_minimum is not None and exclusive_maximum is not None:
```
g-bar: We discussed this on teams when we first implemented the simulation configuration screen for CNS, the frontend doesn't support `exclusive_minimum` and `exclusive_maximum`, it will just get ignored if in the schema, it only uses `minimum` and `maximum`.

But the `exclusive` values can always be modelled by a minimum plus / minus a small boundary  value, if you recall all float values have an absolute boundary of `2.225×10−308` approximately (by the [IEEE 754 standard](https://en.wikipedia.org/wiki/Double-precision_floating-point_format) which most modern programming languagees use ) So any `exclusive` value can be modeled by a maximum / minimum minus / plus `2.225×10−308`.

james-isbister: I think it would be nice to still use the notation for exclusive constraints such as PositiveFloat and gt in Pydantic, rather than using float constraints in Pydantic of say:
`Field(ge=2.0 + sys.float_info.min`

I also checked in the schema and this resolves to 
`"minimum": 2` 

so this doesn't seem to solve the issue.

My suggestion would be that I could add checks that only one of exclusiveMinimum or minimum appears in the schema to the schema test script?

Then for the UI, if exclusiveMinimum appears, we would define some expected behaviour (in the schema spec). i.e. if the user enters a value less than the exclusiveMinimum, the value resolves to exclusiveMinimum + 0.00001 for floats and exclusiveMinimum + 1 for integers, say

What do you think? 

I'll also make a ticket one day for also displaying the constraints in the description hover over

PR #538, opened 2026-01-21
Estimate

jdcourcol:
This is the implementation of https://github.com/openbraininstitute/launch-system/blob/main/docs/overview.md#task-cost-estimation

however
- we should use the actual circuit & simulation target to compute the number of neurons, it uses neuron_count instead
- implements only the circuit simulation and extraction use case
- extraction still requires its own accounting enum

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,9 +82,6 @@ connectivity = [
 ]
 
 [dependency-groups]
-audit = [
-    "pip-audit>=2.9.0",
-]
 dev = [
     "coverage[toml]",
     "jsonschema>=4.25.1",
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -35,11 +35,6 @@ jobs:
           PYTHONPATH: .
       - name: Run pip-audit
         run: make pip-audit
-      - name: Upload pip-audit output
-        uses: actions/upload-artifact@v4
-        with:
-          name: pip-audit-output.json
-          path: pip-audit-output.json
       - name: Cleanup
         if: ${{ always() }}
         run: |
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,8 @@ help:  ## Show this help
 
 install:  ## Create a virtual environment
 	CMAKE_POLICY_VERSION_MINIMUM=3.5 uv sync --extra connectivity
+
+install-ipython: install ## Create a virtual environment and install the ipython kernel
 	uv run python -m ipykernel install --user --name=obi-one --display-name "obi-one"
 
 compile-deps:  ## Create or update the lock file, without upgrading the version of the dependencies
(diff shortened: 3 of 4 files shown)


app/endpoints/task_launch.py
```diff
@@ -30,11 +33,20 @@ class TaskConfigType(StrEnum):
     """List of entitycore config types supported for job submission."""
 
     CIRCUIT_EXTRACTION = entitysdk.models.CircuitExtractionConfig.__name__
+    SIMULATION = entitysdk.models.Simulation.__name__
```
eleftherioszisis: CIRCUIT_SIMULATION would be more symmetric to CIRCUIT_EXTRACTION  I believe.

pyproject.toml
```diff
@@ -72,7 +72,8 @@ dependencies = [
     "obi_notebook",
     "morph_spines",
     # CAVEclient
-    "caveclient>=7.11.0"
+    "caveclient>=7.11.0",
```
eleftherioszisis: Is this used?

app/endpoints/task_launch.py
```diff
@@ -148,6 +160,62 @@ def _generate_failure_callback(
...
+            raise HTTPException(status_code=HTTPStatus.BAD_REQUEST, detail=msg)
+        # Get the scale and map it to service subtype
+        # Scale is stored as a string in the entity (e.g., "small", "microcircuit")
+        circuit_scale = str(circuit_entity.scale)
+        scale_to_subtype = {
+            "small": ServiceSubtype.SMALL_SIM,
+            "microcircuit": ServiceSubtype.MICROCIRCUIT_SIM,
+            "region": ServiceSubtype.REGION_SIM,
+            "system": ServiceSubtype.SYSTEM_SIM,
+            "whole_brain": ServiceSubtype.WHOLE_BRAIN_SIM,
```
eleftherioszisis: Don't we have access to the scale StrEnum from entitysdk to use here instead of hardcoding the values?

app/endpoints/task_launch.py
```diff
@@ -148,6 +160,62 @@ def _generate_failure_callback(
...
+        )
+        if circuit_entity is None:
+            msg = (
+                f"Circuit entity with ID '{circuit_id}' referenced by simulation "
+                f"could not be retrieved."
+            )
+            raise HTTPException(status_code=HTTPStatus.BAD_REQUEST, detail=msg)
+        # Get the scale and map it to service subtype
+        # Scale is stored as a string in the entity (e.g., "small", "microcircuit")
+        circuit_scale = str(circuit_entity.scale)
```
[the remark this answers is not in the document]
jdcourcol: ok

app/endpoints/task_launch.py
```diff
@@ -148,6 +162,55 @@ def _generate_failure_callback(
...
+    entity_type: TaskConfigType,
+    entity_id: str,
+) -> dict:
+    """Evaluates accounting parameters from the task configuration.
+
+    Returns the service subtype and count needed for cost estimation.
+    For Simulation configs, determines the service subtype based on the circuit scale
+    and uses the neuron_count from the simulation entity for the count.
+    """
+    if entity_type == TaskConfigType.SIMULATION:
```
chr-pok: I would imaging that this function will grow a lot once more and more tasks will be added, since each of them will have different evaluation rules.

Just wondering if it would make sense to implement this part in a more generic way, by just calling a generic obi-one function (like we do with `run_task_for_single_config_asset` when launching tasks). Such function could determine the type of the config and would call a `.evaluate()` function of the corresponding class (like we do with `.execute()`). In this way, the `.execute()` and `.evaluate()` functions would always be together in the same task class.

jdcourcol: I meant "the task to launch a circuit simulation"

eleftherioszisis (approved):

PR #550, opened 2026-01-28
Add config entry to set CORS origin with a regex

pgetta:
This is to eventually enable CORS for https://*.preview.openbraininstitute.org - origins needed by Core Web app preview deployments (where each PR will have it's own).


james-isbister (commented):
Thanks @pgetta I look forward to seeing the core-web-app branches in preview.openbrain

All seems good, perhaps @GianlucaFicarelli can confirm though

PR #554, opened 2026-01-28
Root element names

james-isbister:
* root_block -> block_single
* block_single -> block_union
* Changed in package, docs and tests
* Also fixed test to use `validate_block_single` (previously `validate_root_block` which wasn't getting used)


james-isbister:
Went for:
* root_block -> block_single
* block_single -> block_union

cc @g-bar @bilalesi

PR #562, opened 2026-01-30
Update docs inline with Pydantic upgrades

james-isbister:


docs/gui-definition-spec/components/ui_hidden/ui_hidden.md
```diff
@@ -10,5 +10,8 @@ class Block:
                         ui_hidden=True,
                         ui_element="string_input",
                         title="title",
-                        description="description")
+                        description="description",
+                        json_schema_extra={"ui_element": "reference",
```
AurelienJaquier: Same comment here about string input and reference

PR #563, opened 2026-01-30
Fix Pydantic deprecation warnings + enforce no Pydantic warnings, update gui-specs appropriately

james-isbister:
Already reviewed by @AurelienJaquier in multiple PRs to this branch. cc @chr-pok @g-bar @GianlucaFicarelli @ilkilic 

This leaves no deprecation warnings for the obi-one package. Although there are still non Pydantic warnings to fix in the tests

**Fix pydantic deprecations**

* Fix `PydanticDeprecatedSince20: Using extra keyword arguments on Field is deprecated and will be removed. Use json_schema_extra instead.` in multiple files by moving to json_schema_extra dictionary and changing name -> title
* Fix `PydanticDeprecatedSince20: Support for class-based `config` is deprecated, use ConfigDict instead.` in multiple files
* Fix: `PydanticDeprecatedSince20: `json_encoders`` is deprecated

**Minor changes noticed when fixing deprecations**
* Remove unused function 
* Update comments on serialization
* Rename serializtion.py -> deserialize.py

**Enforce no Pydantic warnings in CI, so that the deprecation warnings don't return**

**Update gui-spec docs appropriately**
* Notably non-standard keys in the Field definitions should be encoded as, for example:

```
json_schema_extra={"ui_element": "string_input",
                                            "ui_hidden": True}
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -191,9 +191,6 @@ max-statements = 50
 max-public-methods = 60
 
 [tool.pytest.ini_options]
-filterwarnings = [
-    "ignore::pydantic.warnings.PydanticDeprecatedSince20",
-]
 minversion = "6.0"
 addopts = [
     "--import-mode=importlib",


AurelienJaquier (approved):
looks good to me!

PR #568, opened 2026-02-04
Fix empty morphometric values and complete neurite metric template

ilkilic:
This PR follows up on #567.

While the previous PR fixed neurite-specific metric computation, two issues were still left:
- Empty neurom results ([]) could still end up being written as measurements, which led to 500 errors
- Axon and apical dendrite were still missing several per-section metrics that basal dendrite already had in the template

Changes in this PR:
- Treat empty neurom results as missing values
- Skip writing measurement items when there is no data
- Drop measurement kinds where all values are empty
- Align the template so axon and apical dendrite have the same per-section metrics as basal dendrite

### DIFF
--- a/app/endpoints/useful_functions/useful_functions.py
+++ b/app/endpoints/useful_functions/useful_functions.py
@@ -239,6 +239,23 @@ def update_measurement_items(
         _update_scalar_items(measurement_items, entry_value)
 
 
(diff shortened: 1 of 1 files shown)


dkeller9 (approved):
It should be okay to run this script to fix the entries.

PR #572, opened 2026-02-06
Remove non-existent keyword from oneshot accounting method

eleftherioszisis:


chr-pok (approved):
Thanks, should be fine!

PR #574, opened 2026-02-08
Converting + Uploading BlueConfig Simulation Campaigns

james-isbister:
* Add BlueConfig simulation campaign -> sonata conversion and upload to projects/sim_campaign_upload
* Use for cell assemblies and spike sorting campaigns
* Readme for steps
* Notebook for downloading all spike sorting blue configs


chr-pok:
> Maybe you could take a quick look @chr-pok and we can discuss?

Could you put the converted files on OneDrive please so I can have a look?

> Added: Ah, I have the suspicion that the ids in id_mapping.json have the offset of 5,000,000 and 6,000,000 removed for VPM and POm respectively?

Yes, I think all IDs in the id_mapping.json file are related to the SONATA version of the circuit where all IDs of a given node population start with 0.

chr-pok:
If I remember correctly, all the campaign configs on AWS are missing. Last time I got it from your old hard drive I think. But I don't know about the assemblies one, sorry. We may have it on Zenodo perhaps?

james-isbister:
Ah ok, I'll just create one from an example on my hard drive, then. Thanks!

PR #581, opened 2026-02-10
Fix accounting callback payload

eleftherioszisis:
Error:
```
      {
        "status": "failure",
        "details": {
          "url": "https://staging.cell-a.openbraininstitute.org/api/accounting/usage/oneshot",
          "json": {
            "type": "oneshot",
            "count": 1,
            "job_id": "003e61b4-706d-4d5b-a860-73e5f2c823cf",
            "proj_id": "2720f785-a3a2-4472-969d-19a53891c817",
            "subtype": "small-circuit-sim",
            "timestamp": "2026-02-10T11:00:54.370659+00:00"
          },
          "method": "POST",
          "params": null
        },
        "event_type": "job_on_success",
        "action_type": "http_request_with_token",
        "error_reason": {
          "details": null,
          "message": "HTTP status error 422: {\"error_code\":\"INVALID_REQUEST\",\"message\":\"Validation error\",\"details\":[{\"type\":\"int_parsing\",\"loc\":[\"body\",\"timestamp\",0],\"msg\":\"Input should be a valid integer, unable to parse string as an integer\"}]}",
          "error_code": "GENERIC_ERROR",
          "http_status_code": 500
        }
      }
```
count needs to be passed as a string. Following a discussion with @pgetta I also changed the timestamp to be calculated the same way as in accounting sdk.

### DIFF
--- a/app/services/accounting.py
+++ b/app/services/accounting.py
@@ -1,13 +1,14 @@
-import time
 from http import HTTPStatus
 from uuid import UUID
 
+import httpx
 from entitysdk import Client, ProjectContext, models
 from entitysdk.types import CircuitScale
 from fastapi import HTTPException
 from obp_accounting_sdk import AccountingSessionFactory, OneshotSession
 from obp_accounting_sdk.constants import ServiceSubtype
 from obp_accounting_sdk.errors import BaseAccountingError, InsufficientFundsError
+from obp_accounting_sdk.utils import get_current_timestamp
 
 from app.config import settings
 from app.errors import ApiError, ApiErrorCode
@@ -17,6 +18,7 @@ from app.schemas.auth import UserContext
 from app.schemas.callback import CallBack, HttpRequestCallBackConfig
 from app.schemas.task import TaskAccountingInfo, TaskDefinition
 from app.types import CallBackAction, CallBackEvent, TaskType
+from app.utils.http import make_http_request
 
 CIRCUIT_SCALE_TO_SERVICE_SUBTYPE = {
     CircuitScale.small: ServiceSubtype.SMALL_SIM,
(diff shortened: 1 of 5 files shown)


app/services/accounting.py
```diff
@@ -201,8 +201,8 @@ def _generate_accounting_success_callback(
             "subtype": service_subtype,
             "proj_id": project_id,
             "job_id": accounting_job_id,
-            "count": count,
-            "timestamp": datetime.now(UTC).isoformat(),
+            "count": str(count),
+            "timestamp": str(int(time.time())),
```
jdcourcol: wait , there is an issue here. The callback is generated at simulation launch, however this timestamp should be at simulation ending no ?

jdcourcol: @pgetta ?

pgetta: For the oneshot jobs the timestamp [is used](https://github.com/openbraininstitute/accounting-service/blob/main/app/task/queue_consumer/oneshot.py#L48) to set the value for all of these:
* `started_at`
* `last_alive_at`
* `finished_at`

For the long run jobs these values are updated separately during multiple calls.

So indeed, for the oneshot jobs it doesn't matter that much, but if we want accuracy - the timestamp should reflect the job execution timing as much as possible.

eleftherioszisis: To fix this we would need to extend launch-system callback functionality with labeled function support (e.g. time_now) which would be executed during the callback execution.

Until then there isn't a better way to calculate this timestamp.

jdcourcol:
>To fix this we would need to extend launch-system callback functionality with labeled function support (e.g. time_now) which would be executed during the callback execution.

it is equivalent to not passing the timestamp and use "now" as the default no ?

app/services/accounting.py
eleftherioszisis: Good point. I'll do that.

PR #585, opened 2026-02-12
Refactor morphology validation endpoint

GianlucaFicarelli:
Main changes:
- convert the async endpoint to sync so it's executed in the fastapi thread pool, since most of the logic is blocking and was executed in threads anyway
- move the morphology and file logic to a separate file so it can be reused
- simplify the cleanup of the temp directory

Note:
the endpoint is called `test-neuron-file`, but I wonder if it should be changed to something different, since this endpoint is not just validation but also conversion to other formats.

### DIFF
--- a/app/endpoints/morphology_validation.py
+++ b/app/endpoints/morphology_validation.py
@@ -1,222 +1,70 @@
-import asyncio
-import pathlib
-import shutil
-import tempfile
-import zipfile
 from http import HTTPStatus
+from pathlib import Path
 from typing import Annotated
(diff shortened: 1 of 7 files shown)


app/endpoints/morphology_validation.py
dkeller9: The changes are fine. I tested the endpoint locally and it runs fine. The name of the endpoint has already been much discussed so we should keep it the same.

dkeller9 (approved):

PR #592, opened 2026-02-13
Hide block_dictionary and block_union options based on usability endpoint

james-isbister:
Working with https://github.com/openbraininstitute/core-web-app/pull/1407

* Specify how to represent that a block should be greyed out in the block selector (central column) for both block_dictionary and block_union ui_elements, based on the entity which is being used.**

* `block_usability_entity_dependent` (bool) can be set to true in Block non-validating schema, to signify that the usability of a block depends on the value of a key in the dictionary returned by an endpoint in the dictionary `property_endpoint` specified in the parent ScanConfig.
* `block_usability_entity_dependent` is set to False by default in all Blocks
* Additional non-validating properties should also be specified in the Block if `block_usability_entity_dependent=True`. Particularly:
  - `block_usability_group` (Specifies the entity type that this relates to)
  - `block_usability_property_group` (Specifies the key in the dictionary returned by the `usability_endpoint`)
  - `block_usability_false_message` (Specifies the message to be displayed as a hover over in the case that the Block is greyed out due to `block_usability_entity_dependent=False`)

**Additional required changes**
* `properties_endpoint` in `CircuitSimulationScanConfig` to explicitly signify the endpoints that should be used for `entity_property_dropdown` and `block_usability_entity_dependent` 
* New `json_schema_extra_additions: ClassVar[dict`] can be added to any class deriving from OBIBaseModel (i.e. Block, Task). Parameters in `json_schema_extra_additions` are added to json_schema_extra (non-validating properties) iteratively through all subclasses by .update(). Previously `model_config = ConfigDict(json_schema_extra={...})`in sublass would overwrite parent class model_config
* New endpioint: `/circuit-simulation-options-usability/{circuit_id}`

### DIFF
--- a/app/endpoints/circuit_properties.py
+++ b/app/endpoints/circuit_properties.py
@@ -144,7 +144,7 @@ def mapped_circuit_properties_endpoint(
                 "detail": f"Internal error retrieving the circuit {circuit_id}.",
             },
         ) from err
-    
+
     # Add usability
     try:
         circuit = db_client.get_entity(entity_id=circuit_id, entity_type=Circuit)
@@ -162,6 +162,6 @@ def mapped_circuit_properties_endpoint(
         CircuitUsability.SHOW_INPUT_RESISTANCE_BASED_STIMULI: False,
     }
 
-    mapped_circuit_properties['usability'] = simulation_options_usability
+    mapped_circuit_properties["usability"] = simulation_options_usability
 
     return mapped_circuit_properties
--- a/app/endpoints/scan_config.py
+++ b/app/endpoints/scan_config.py
@@ -27,7 +27,6 @@ from obi_one.scientific.tasks.morphology_metrics import (
 )
 from obi_one.scientific.tasks.schema_example import SchemaExampleScanConfig
 from obi_one.scientific.tasks.skeletonization import SkeletonizationScanConfig
-from obi_one.scientific.unions.aliases import SimulationsForm
 
 router = APIRouter(prefix="/generated", tags=["generated"], dependencies=[Depends(user_verified)])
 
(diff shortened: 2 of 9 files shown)


james-isbister:
Bug fixed

docs/gui-definition-spec/components/block_usability_entity_dependent/block_usability_entity_dependent.md
```diff
@@ -5,13 +5,13 @@ A `block` can be optionally greyed out from selection for `block_unions` and `bl
...
 - `block_usability_entity_dependent` (This is set as False for all Blocks automatically)
 
-If the value of `block_usability_entity_dependent` is set to `True`, then the following properties must also be specified:
-- `block_usability_property_group`: The name of a key for which there is corresponding key specified in the `property_endpoints` dictionary of the parent scan config (see below). The value for this key in the `property_endpoints` dictionary references an endpoint which will return a `property ditionary`. Within this property dictionary there is a dictionary titled `usability`, which contains various boolean values which can be used to optionally display different elements.
-- `block_usability_property`: The name of the boolean to use from the `usability` dictionary. If true the block should be useable. If false it should be greyed out and unusable.
-- `block_usability_false_message`: The message displayed if the value for the key specified by boolean is false.
+If the value of `block_usability_entity_dependent` is set to `True`, then a dictionary `block_usability_dictionary` should be present with the following properties:
```
g-bar: If we have the `block_usability_dictionary`, then we don't need `block_usability_entity_dependent`, we just make the dict optional and if the block is usability dependent is implicit in the presence or absence of the dict.

james-isbister:
Removed the boolean @g-bar

PR #596, opened 2026-02-17
Skip metrics for absent neurite types

ilkilic:
Skip computing axon/apical/basal metrics when the neurite is not present in the morphology.
Previously empty neurom results produced zero-valued measurements (e.g. axon length = 0).
Absent neurites now yield no measurements instead of zeros.


dkeller9 (approved):
LGTM

PR #600, opened 2026-02-17
new expandable_list ui element

AurelienJaquier:
This is for this issue: https://github.com/openbraininstitute/prod-build-ion-channel-model/issues/77

For my use case, I need a list with an arbitrary number of (duration, voltage) pairs. I think it would be nice if this new ui_element is generalised to other use cases, but I am not sure how easy it would be to do.

I don't have any figma yet, so I did a quick drawing for the time being


james-isbister (commented):
Nice! Some initial thought

james-isbister (commented):
Not sure whether we want to support parameter sweeps in this PR or leave it to a later PR. Is the ComplexVariableHolder logic ready to be considered?

For one, the design doesn't show us how parameter sweeps will look. 

In theory specifying voltage and duration as "float_parameter_sweeps" as you have done might allow us to re-use some logic for display and validation of voltages and durations within each step. For @g-bar to comment on this.

There is also something possibly quite general about specifying ui_elements within each  DurationVoltageCombination. i.e. specifying multiple sets of parameters (per "step" in this case) where each parameter has it's own ui_element type and bounds.  

Maybe we can keep an eye to generalisation in the future, but just implement what is needed for "voltage_duration_ui_element" for now

docs/gui-definition-spec/components/voltage_duration_ui_element/voltage_duration_ui_element.md
```diff
@@ -0,0 +1,57 @@
...
+        json_schema_extra={
+            "ui_element": "float_parameter_sweep",
+            "unit": "ms",
+        },
+    )
+
+
+class Block:
+
+    duration_voltage_combinations: list[DurationVoltageCombination] = Field(
```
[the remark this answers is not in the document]
AurelienJaquier: Yes, exactly.
So `[{voltage: 1, duration: 1}, {voltage:2, duration:1}]` is a stimulus with two levels
and `[{voltage: [1,2], duration: 1}]` is a one level stimulus with parameter scan over the voltage, creating two configs (and thus two simulations): one config with a single level `[{voltage:1, duration:1}]` and another config with another single level `[{voltage:2, duration:1}]`

AurelienJaquier:
Is this good enough that I can ask for an UI implementation in the next ticket assignment meeting @g-bar ?

james-isbister:
Do we have a design for scanning over duration and/or voltage within a single DurationVoltageCombination?

AurelienJaquier:
> Do we have a design for scanning over duration and/or voltage within a single DurationVoltageCombination?

I have this design: https://www.figma.com/design/akGPTH0WwNFDfSWSs3qAnh/OBI---UI-UX-2025?node-id=9277-12607&t=w7zJ6XMnNnrijloP-0
with the `+` button on the right to add values to the parameter scan.
I then expect to new fields to appear in the same way it does for any other parameter scan. Or do we need a design for that also?

james-isbister:
Otherwise, the schema looks good to me. On my side it's good to propose in the ticket meeting though, although ideally we would have the design for the parameter scans before then

I'll still need to look more at the details of scanning over variables within a ComplexVariableHolder though, and how this would extend to other use cases, but this is separate from the frontend implementation

james-isbister:
`I have this design`
Cool, I'd just put the "+" symbols within circles to be consistent with the current UI.

Then you could update the png in this PR with that

AurelienJaquier:
> Cool, I'd just put the "+" symbols within circles to be consistent with the current UI.

@loris-olivier-obi says he wants to have a simple `+` design without circles around for all the parameter scans. That is why we don't have the circles here

james-isbister:
I think we should stay consistent for now. We shouldn't change the design across workflows in one small unrelated PR that other people aren't aware of

g-bar:
The schema should be good, we can use the same component / design for the parameter sweep fields.

docs/gui-definition-spec/components/voltage_duration_ui_element/voltage_duration_ui_element.md
```diff
@@ -0,0 +1,57 @@
...
+```py
+
+class DurationVoltageCombination(ComplexVariableHolder):
+    """Class for storing pairs of duration and voltage combinations for stimulation protocols."""
+
+    voltage: float | list[float] = Field(
+        title="Voltage for each level",
+        description="The voltage for each level, given in millivolts (mV).",
+        json_schema_extra={
+            "ui_element": "float_parameter_sweep",
```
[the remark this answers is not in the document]
g-bar: There is no use for this nested ui_element, since the 'voltage' and 'duration' fields are hardcoded in the schema there is no point to "dynamically" render the inner elements since they never change for the voltage duration ui_element, this would only useful with a "generic" ui_element where the inner fields can vary.

I'm not sure how even making this generic would work, since the pydantic models are static so there is no way to define the field names dynamically.

So it should be removed from the schema, I will just ignore it for now, and hardcode ParameterSweep component for both of those fields on the FE.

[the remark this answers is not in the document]
AurelienJaquier: Alright, noted.
I have updated this file, the json schema and the `DurationVoltageCombination` class not to have the ui element for `voltage` and `duration` in my latest commit.

obi_one/core/block_subunit/complex_variable_holder.py
```diff
@@ -0,0 +1,55 @@
...
+class ComplexVariableHolder(OBIBaseModel, extra="forbid"):
+    """Has structure List[Tuple[AnyOf(int, float, list[float], list[int])]]."""
+
+    _multiple_value_parameters: list[MultiValueScanParam] = PrivateAttr(default=[])
+
+    # checks this for parameter scan
+    def multiple_value_parameters(self, base_location_list: list[str]) -> list[MultiValueScanParam]:
+        self._multiple_value_parameters = []
+
+        for key, value in self.__dict__.items():
```
james-isbister: The logic inside this for loop seems like it could be simplified. We create a multi_values array of length 1, and then iterate through it.

obi_one/core/scan_generation.py
```diff
@@ -108,6 +109,27 @@ def coordinate_parameters(self) -> list[SingleCoordinateScanParams]:
         msg = "coordinate_parameters() must be implemented by a subclass of Scan."
         raise NotImplementedError(msg)
 
+    def set_nested_single_coordinate_scan_param_value(
```
james-isbister, line 112: I like the use of recursion but it is is a bit confusing that the first variable of the function is type annotated as 'ScanConfigsUnion' of name 'single_coord_config' when it could also be a Block or a variable if I understand correctly?

obi_one/core/scan_generation.py
```diff
@@ -108,6 +109,27 @@ def coordinate_parameters(self) -> list[SingleCoordinateScanParams]:
...
+        self, single_coord_config: ScanConfigsUnion, location_list: list, value: Any
+    ) -> Any:
+        if location_list == []:
+            return value
+
+        if isinstance(single_coord_config, (list, dict, tuple)):
```
james-isbister: I was thinking that ComplexVariableHolder would have two functions: 1) for filling the MultiValueScanParam and 2) for unpacking the SingleValueScanParam (into the single coordinates parameter)

Subclasses of ComplexVariableHolder could then overwrite this logic

I'd recommend that we try to generalise so that the neuronal parameter modification supports parameter scans (just in the package not in the ui for now) so that we're sure we have a schema which generalises and we don't have to go and make edits to config in entitycore unnecessarily 

Particularly, this class would be another case of a ComplexVariableHolder, with scan parameters added in the section_list_modifications dictionaries with keys that we don't know in advance (because they depend on the data)

The pattern would look very similar for the properties neuron set which we hope to implement soon

james-isbister (commented):
Have added additional comments outside of this review (as I didn't want to keep you waiting any longer), and I wasn't sure when I'd finish the review.

obi_one/core/block_subunit/complex_variable_holder.py
AurelienJaquier: Done in latest commit

obi_one/core/scan_generation.py
AurelienJaquier: I renamed it single_coord_config_subelement in the latest commit. Is it better now?

obi_one/core/scan_generation.py
AurelienJaquier: I made `ByNeuronModification` a `ComplexVariableHolder`. I had to change `Block.multiple_value_parameters()` to make it work, because before I only expected `ComplexVariableHolder`s to be in `lists`, but I guess having them outside of `lists` is a use-case we want to support.
Edit: I also made it so we can scan over `BySectionListModification`. However, that means that any `dict` in a `ComplexVariableHolder` that contains `lists` will be scanned over. It is the behavior that we want right now, but we should keep that in mind if we implement other `ComplexVariableHolders` containing `dicts` in the future.

james-isbister (commented):
I think your new proposal for the ComplexVariableHolder is nice and would extend to the PropertyNeuronSet which will use a dictionary. although I'll need to look at the details a little more. Is there an example notebook where I could test the the scanning behaviour of the new stimulus and the neuronal modification?

AurelienJaquier:
> I think your new proposal for the ComplexVariableHolder is nice and would extend to the PropertyNeuronSet which will use a dictionary. although I'll need to look at the details a little more. Is there an example notebook where I could test the the scanning behaviour of the new stimulus and the neuronal modification?

I have this script that I used to test my implementation
[create_obione_form.py](https://github.com/user-attachments/files/26962547/create_obione_form.py)

james-isbister:
Please could you add it as an example notebook, also with the new stimulus?

AurelienJaquier:
> Please could you add it as an example notebook, also with the new stimulus?

Done! Added two examples: neuronal modification scan example and icm with mulit-level seclamp example

james-isbister:
Sorry, I didn't see your last comment. I'll try to take a look today

james-isbister (approved):
The two examples work well. Thanks for providing them, and for finding a good solution for the ComplexVariableHolders!

PR #604, opened 2026-02-20
fix in icm modeling

AurelienJaquier:
the figure_filepaths dict is slightly different now. We have to make sure we properly isolate the figure paths in the register_plots_and_json function

### DIFF
--- a/app/endpoints/morphology_metrics.py
+++ b/app/endpoints/morphology_metrics.py
@@ -1,12 +1,18 @@
+import tempfile
 from http import HTTPStatus
 from typing import Annotated, Literal
 
 import entitysdk.client
 import entitysdk.exception
+from entitysdk.models.cell_morphology import CellMorphology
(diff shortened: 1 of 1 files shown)


obi_one/scientific/tasks/ion_channel_modeling.py
```diff
@@ -392,11 +392,11 @@ def register_plots_and_json(
...
         # get the paths of the pdf figures
         paths_to_register = [
-            value
+            figpath
             for key1, d in figure_filepaths.items()
             if key1 != "thumbnail"
-            for key, value in d.items()
-            if key != "order"
+            for figpath in d.values()
+            if figpath[-4:] == ".pdf"
```
ilkilic: Maybe add a small guard in case of uppercase (e.g. Path(figpath).suffix.lower() == ".pdf")

PR #609, opened 2026-02-25
Small circuit visualization endpoints

g-bar:


app/endpoints/circuit_visualization.py
```diff
@@ -0,0 +1,463 @@
...
+
+    return res
+
+
+def get_morphology(
+    parent_dir: Path, client: Client, circuit_id: UUID, asset_id: UUID, morph_path: Path
+) -> MorphologyDict:
+    output_path = parent_dir / morph_path
+
+    if not output_path.exists():
```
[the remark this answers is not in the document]
g-bar: I addressed this in latest commit.

james-isbister (commented):
Looks good. 

It would be great if you could make a quick notebook for calling each of these endpoints, like: https://github.com/openbraininstitute/obi-one/blob/main/examples/A_service_and_entitycore/test_endpoints/test_endpoints.ipynb

obi_one/scientific/library/circuit_visualization.py
```diff
@@ -0,0 +1,476 @@
...
+Morphology = dict[str, NeuronSectionInfo]
+
+
+def circuit_asset_id(client: Client, circuit_id: UUID) -> UUID:
+    try:
+        circuit = client.get_entity(entity_id=circuit_id, entity_type=Circuit)
+    except Exception as e:
+        L.exception(e)
+        raise HTTPException(
+            status_code=HTTPStatus.NOT_FOUND,
```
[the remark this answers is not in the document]
g-bar: Yes, good practice, the repo doesn't have any domain specific errors so far, will keep as is for now, to be addressed in future PRs.

GianlucaFicarelli:
> Also it would be useful to add some tests at some stage, but as this is to support a feature you're responsible for in the frontend it's up to you if and when you want to add these
> 
> We have examples of tests for endpoints here though: https://github.com/openbraininstitute/obi-one/tree/main/tests/routers/declared

I agree and strongly suggest to add tests in this PR, or it would add technical debt.

app/endpoints/circuit_visualization.py
```diff
@@ -0,0 +1,71 @@
...
+                db_client,
+                circuit_id,
+                asset_id,
+                Path(urllib.parse.unquote(morphology_path + ".swc")),
+            )
+        except HTTPException:
+            raise
+        except Exception as e:
+            L.exception(e)
+            raise HTTPException(status_code=404, detail="Morphology not found") from e
```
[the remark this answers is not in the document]
g-bar: OK, removed the try, catch block for now.

app/endpoints/circuit_visualization.py
```diff
@@ -0,0 +1,65 @@
...
+    description="Returns a list of nodes for visualization",
+)
+def circuit_nodes(
+    circuit_id: UUID,
+    db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
+    temp_dir: TempDirDep,
+) -> Nodes:
+    asset_id = circuit_asset_id(db_client, circuit_id)
+
+    parent_path = Path(temp_dir).resolve()
```
GianlucaFicarelli: I don't think this line is needed, temp_dir is already a Path, and I don't see the reason to call resolve since it's an absolute path already.

obi_one/scientific/library/circuit_visualization.py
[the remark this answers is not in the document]
james-isbister: This is a good spot. I agree that HTTPExceptions should be removed from the package

app/endpoints/circuit_visualization.py
g-bar: Removed the line

app/services/circuit_visualization.py
```diff
@@ -0,0 +1,478 @@
...
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST,
+            detail={
+                "code": ApiErrorCode.INVALID_REQUEST,
+                "detail": f"Missing file {asset_path}",
+            },
+        ) from e
+
+    try:
+        with h5py.File(nodes_file_path, "r") as f:
```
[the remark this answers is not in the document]
g-bar: Yeah, kinda.
First I was using bluepysnap but it requred the entire sonata_circuit directory whereas now I only download the files I need.

Then I managed to get around that but the code was twice as slow and didn't really save many lines
compared to the current, so I scrapped it.

app/schemas/visualization.py
```diff
@@ -0,0 +1,51 @@
+from typing import Annotated
```
GianlucaFicarelli: I suppose that this file can be named `circuit_visualization.py` instead of `visualization.py`, to match the same names used in endpoints and services?

app/services/circuit_visualization.py
```diff
@@ -0,0 +1,474 @@
...
+                entity_id=circuit_id,
+                asset_id=asset_id,
+                asset_path=morph_path,
+            )
+        except Exception as e:
+            raise HTTPException(status_code=404, detail="Morphology not found") from e
+
+    try:
+        if output_path.suffix.lower() == ".h5":
+            morph = morphio.Collection(output_path).load(morph_name)
```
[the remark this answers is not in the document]
g-bar: So for the morphology endpoint, the code is only called with 1 node.

For the soma_radius case there might be up to 10 nodes but only if the morphology is in `h5` if it's `swc` which is the majority of cases you still need to load each file individually.

So it's a microoptimization for a rare case, I don't think it's worth it, imho.

app/services/circuit_visualization.py
```diff
@@ -0,0 +1,433 @@
...
+                output_path=output_path,
+                entity_id=circuit_id,
+                asset_id=asset_id,
+                asset_path=morph_path,
+            )
+        except Exception as e:
+            raise HTTPException(status_code=404, detail="Morphology not found") from e
+
+    try:
+        if output_path.suffix.lower() == ".h5":
```
GianlucaFicarelli: If I'm not wrong, the `.h5` suffix can be used for both collections and single morphologies in h5 format, and I'm not sure that calling Collection with a single morphology is correct.

app/services/circuit_visualization.py
```diff
@@ -0,0 +1,474 @@
...
+                "code": ApiErrorCode.INVALID_REQUEST,
+                "detail": f"Missing file {asset_path}",
+            },
+        ) from e
+
+    try:
+        storage = libsonata.NodeStorage(str(nodes_file_path))
+        population = storage.open_population(population_name)
+
+        selection = libsonata.Selection(np.arange(population.size))
```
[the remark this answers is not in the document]
g-bar: Yes, this is important.

The endpoint already checks that scale must be 'small' or 'pair' and there shouldn't be any circuits larger than 10 with those scales.

Just to be safe I added a check that returns an error if nodes exceeds 10 for now, but in theory should never be hit.

app/endpoints/circuit_visualization.py
```diff
@@ -0,0 +1,60 @@
...
+) -> Nodes:
+    asset_id = circuit_asset_id(db_client, circuit_id)
+
+    config = download_circuit_config(db_client, circuit_id, asset_id, temp_dir)
+
+    return get_nodes(config, temp_dir, db_client, circuit_id, asset_id)
+
+
+@router.get(
+    "/{circuit_id}/morphologies/{morphology_file:path}/{morphology_name:path}",
```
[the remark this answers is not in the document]
g-bar: Changed the implementation.
Now the morphology_name is an optional query param, only needed if it's not already in the path.

> In the url above I also find difficult to tell what is the path of the endpoint and the path of the morphology.

`/{circuit_id}/morphologies/morphologies/merged-morphologies.h5/rp110125_L5-2_idF_-_Scale_x1.000_y1.025_z1.000` is not a valid endpoint pydantic will return 401. The path needs to be url encoded, query param would still need to be url_encoded so it doesn't make any difference.
 I documented usage in the swagger.

app/services/circuit_visualization.py
```diff
@@ -0,0 +1,445 @@
...
+        morph_files = population.get_attribute("morphology", selection)
+
+        nodes_list = []
+        for i in range(population.size):
+            m_name = morph_files[i]
+            m_path = morphologies_path.path.relative_to(parent_dir)
+
+            if m_path.suffix:
+                m_file = m_path
+                m_name = None if m_name in str(m_file) else m_name
```
GianlucaFicarelli: Instead of looking for m_name in m_file, would it work if you match exactly `m_name == m_path.stem`?
However, it seems a bit fragile to rely on the comparison to set m_name.
Wouldn't it be simpler if MorphPath contains the information that is pointing to a collection (and path is a file) or to a directory of single morphologies?

app/services/circuit_visualization.py
```diff
@@ -0,0 +1,445 @@
...
+        qy = population.get_attribute("orientation_y", selection)
+        qz = population.get_attribute("orientation_z", selection)
+        qw = population.get_attribute("orientation_w", selection)
+
+        morph_files = population.get_attribute("morphology", selection)
+
+        nodes_list = []
+        for i in range(population.size):
+            m_name = morph_files[i]
+            m_path = morphologies_path.path.relative_to(parent_dir)
```
GianlucaFicarelli: Minor, m_path can be moved outside of the loop

app/services/circuit_visualization.py
g-bar: There is no reliable way to tell if a morph file is a collection.

We could technically pass the dir or file and the name, extension but that introduces several complications, it would require to pass all those as query params, and introduce similar logic relying on the path to determine the file (since that downloads single files not the full sonata circuit) to download in the morphology endpoint. 

For those reason I think the original implementation is the most concise / least complex (pass morph_name as query param and drop it if not needed). Especially since otherwise the only advantage is to save literally 2 lines of code in the load_morphology function. So I'll go with that.

GianlucaFicarelli: My suggestion was to add an attribute `is_collection` when MorphPath is loaded in `resolve_morph_path` so it can be used with something like:
```python
if morphologies_path.is_collection:
    m_file = m_path
else:
    m_file = m_path / f"{m_name}.{morphologies_path.format}"
```
However, the implementation using
```python
m_file = m_path if m_path.suffix else m_path / f"{m_name}.{morphologies_path.format}"
```
isn't much different, so we can go with that.

GianlucaFicarelli (approved):

pyproject.toml
```diff
@@ -75,6 +75,8 @@ dependencies = [
     "morph_spines",
     # CAVEclient
     "caveclient>=7.11.0",
+    "obp-accounting-sdk>=0.5.0",
```
jplanasc, line 78: `obp-accounting-sdk` was already there, just a few lines below with a higher version requirement. @g-bar if you're ok, I'm in the middle of merging my PR to main (resolving conflicts) and I will fix this in my PR.

Just letting you know.

g-bar: Not sure how it got there, I usually add deps using uv add, so it should've detected if it was already there. 
Thank you!

PR #621, opened 2026-03-09
Add ICM Simulation Support via Task Manager

eleftherioszisis:
## Summary
This PR adds Ion Channel Model (ICM) simulation support through the task manager flow, including task routing, execution wiring, and supporting tests.

### What’s included
- Add ICM simulation execution through task-manager-compatible task types.
- Extend task-type/config mappings so ICM execution tasks resolve correctly from task metadata.
- Wire task execution logic for ICM simulation runs, including staging/processing/registration paths.
- Update app mappings/dependency wiring needed for task-manager-driven execution.
- Make asset config label optional for tasks that only use entities as configs

### DIFF
--- a/app/services/task.py
+++ b/app/services/task.py
@@ -139,8 +139,8 @@ def _generic_job_data(
         "resources": resources,
         "inputs": [
             f"--task-type {task_definition.task_type}",
-            f"--config_type {task_definition.config_type_name}",
-            f"--config_id {config_id}",
+            f"--config_entity_type {task_definition.config_type_name}",
+            f"--config_entity_id {config_id}",
             f"--entity_cache {entity_cache}",
             f"--scan_output_root {output_root}",
             f"--virtual_lab_id {virtual_lab_id}",
--- a/launch_scripts/launch_task_for_single_config_asset/code.py
+++ b/launch_scripts/launch_task_for_single_config_asset/code.py
@@ -62,8 +62,8 @@ def main() -> int:
             description="Script to launch a task for a single configuration asset."
         )
         parser.add_argument("--task-type", required=True, help="Task type")
-        parser.add_argument("--config_type", required=True, help="EntitySDK entity type as string")
-        parser.add_argument("--config_id", required=True, help="Entity ID as string")
+        parser.add_argument("--config_entity_type", required=True, help="EntitySDK entity type as string")
+        parser.add_argument("--config_entity_id", required=True, help="Entity ID as string")
         parser.add_argument(
             "--execution_activity_type",
             required=False,
@@ -99,7 +99,7 @@ def main() -> int:
 
     try:
         # Get entity type
-        config_type = getattr(models, args.config_type)
+        config_entity_type = getattr(models, args.config_entity_type)
 
         # Get DB client (incl. file mounting)
         token_manager = TokenFromFunction(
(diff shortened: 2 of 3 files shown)


obi_one/scientific/tasks/ion_channel_model_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+    @property
+    def single_entity(self) -> models.Simulation:
+        return self._single_entity
+
+    @single_entity.setter
+    def single_entity(self, entity: models.Simulation) -> None:
+        self._single_entity = entity
+
+
+def get_task_config(
```
jdcourcol: I would not call that "get_task_config" because that is the generation config. The TaskConfig is "Simulation" in that case.

obi_one/core/run_tasks.py
```diff
@@ -76,3 +83,45 @@ def run_task_for_single_config_asset(
...
+            entity_id=entity_id, entity_type=entity_type, asset_id=config_asset_id
+        ).decode(encoding="utf-8")
+
+        json_dict = json.loads(json_str)
+        json_dict["scan_output_root"] = scan_output_root
+        json_dict["coordinate_output_root"] = Path(scan_output_root) / str(json_dict["idx"])
+        single_config = deserialize_obi_object_from_json_data(json_dict)
+
+    else:
+        single_config = get_task_type_single_config(task_type)(scan_output_root=scan_output_root)
```
james-isbister: My mistake, I think coordinate_output_root=Path(scan_output_root) / str(json_dict["idx"]) in the constructor here two. Maybe that path can be created before the if statement as its shared between if and else

obi_one/scientific/library/simulation/entrypoint.py
```diff
@@ -0,0 +1,374 @@
...
+        simulation_config: Path to the simulation configuration file
+        simulator: Which simulator to use. Must be one of: 'bluecellulab' or 'neurodamus'.
+        libnrnmech_path: Path to mechanisms shared object
+
+    Raises:
+        ValueError: If the requested backend is not implemented.
+    """
+    logger.info("Starting simulation with %s backend", simulator)
+    simulator = simulator.lower()
+    if simulator == "bluecellulab":
```
james-isbister: Can bluecellulab and neurodamus be made constants in library.constants?

obi_one/scientific/library/circuit.py
```diff
@@ -124,3 +133,19 @@ def _default_edge_population_name(c: snap.Circuit) -> str:
     def default_edge_population_name(self) -> str:
         """Returns the default edge population name."""
         return self._default_edge_population_name(self.sonata_circuit)
+
+    @property
+    def mechanisms_dir(self) -> Path:
+        # TODO: There should be only ONE convention for output dir of mechs
```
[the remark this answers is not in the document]
AurelienJaquier: Is the convention in circuit to use `MOD` as the mechanisms directory name?
We usually use `mechanisms` in single cell simulation, but we can switch to `MOD` to be consistent with circuit

obi_one/core/run_tasks.py
eleftherioszisis: Hmm, I am not familiar with the idx pattern here. Do you have an example of what is expected? Is the scan_output_root wrong for single config in this case?

app/services/task.py
```diff
@@ -145,9 +138,9 @@ def _generic_job_data(
         "code": task_definition.code.model_dump(mode="json"),
         "resources": resources,
         "inputs": [
-            f"--entity_type {task_definition.config_type_name}",
-            f"--entity_id {config_id}",
-            f"--config_asset_id {config_asset_id}",
+            f"--task-type {task_definition.task_type}",
+            f"--config_type {task_definition.config_type_name}",
+            f"--config_id {config_id}",
```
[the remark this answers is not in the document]
eleftherioszisis: Not sure it is confusing as `config_id` given that it doesn't mention anything about the asset as was the case with `config_asset_id`.

PR #630, opened 2026-03-13
have correct range metadata when registering model in ic build

AurelienJaquier:
for https://github.com/openbraininstitute/prod-build-ion-channel-model/issues/91
depends on https://github.com/openbraininstitute/ion-channel-builder/pull/22 (to add to dependencies when merged)

also:
- make sure we have at least on ic model for ic simulation
- seclamp improvements: have 3 steps, rename the fields, remove neuron sets from UI, remove total duration and use sum of three steps as total duration under the hood, have default values for step durations that makes the clamp last as long as default value for simulation duration


AurelienJaquier:
tested locally, it works

AurelienJaquier:
@james-isbister is it ok to merge this?

PR #636, opened 2026-03-18
Support for task type without obi-one-coordinate asset + remove need to pass config_asset_id

james-isbister:
1) **Make TaskDefinition.config_asset_label optional (None can be passed)**. 
2) **Remove need to pass config_asset_id**. The asset id is determined and fetched within run_task_type if TaskDefinition.config_asset_label not None. If None, create a default SingleConfig for the task (empty if no parameters required by the task)
4) **IonChannelModelSimulationExecution Task has its own SingleConfig** which has the single_entity set (Simulation) in run_task (as previously for all tasks). All configs and assets are fetched through the Simulation single_entity (including generation config, which is our non-ideal way atm for fetching the ion channel models (until we migrate the simulation beta products to a new entity relationship structure))

**Details**
* get_task_type_config_asset_label used in app/mappings.py. May return None where a config asset does not need to be staged in [launch_scripts/launch_task_for_single_config_asset/code.py](https://github.com/openbraininstitute/obi-one/blob/9fdef184402f97ef63028fef78eb4caf67d25f71/launch_scripts/launch_task_for_single_config_asset/code.py#L132) because all of the information which is needed to run the task is deterministically extractable from a single entity, given the id, inside the task.execute(). Particularly for executing a simulation, the only thing which is needed is the Simulation entity_id
* get_task_type_config_asset_label. also used within core.run_tasks.run_task_type() so config_asset_id no longer needs to be passed - and therefore is no longer needed by: launch_scripts/launch_task_for_single_config_asset/code.py
* ion_channel_model_simulation, IonChannelModelSimulation renamed to ion_channel_model_simulation_execution, IonChannelModelSimulationExecution
* Moved db_sdk.get_config_asset to obi_one, as used in package now, not in app

### DIFF
--- a/app/mappings.py
+++ b/app/mappings.py
@@ -13,9 +13,6 @@ from app.schemas.task import (
     TaskDefinition,
 )
 from app.types import BuiltinScript, TaskType
-from obi_one.scientific.unions.config_task_map import (
-    get_task_type_config_asset_label,
-)
 
 APP_TAG = f"tag:{settings.APP_VERSION.split('-')[0]}"
 OBI_ONE_CODE_PATH = str(Path(settings.OBI_ONE_LAUNCH_PATH) / "code.py")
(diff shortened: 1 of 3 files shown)


app/mappings.py
```diff
@@ -45,7 +47,7 @@
         config_type=models.SkeletonizationConfig,
         activity_type=models.SkeletonizationExecution,
         accounting_service_subtype=ServiceSubtype.NEURON_MESH_SKELETONIZATION,
-        config_asset_label=AssetLabel.skeletonization_config,
+        config_asset_label=get_task_type_config_asset_label(TaskType.morphology_skeletonization),
```
eleftherioszisis: If there is 1-1 mapping between task_type and asset label, then it is not required to be declared in the definitions, isn't that correct?

PR #641, opened 2026-03-20
improve seclamp

AurelienJaquier:
Have a proper SEClamp with 3 steps

The seclamp with an arbitrary number of steps will be another stimulus, and will be done in another PR.


obi_one/scientific/blocks/stimuli/stimulus.py
```diff
@@ -661,7 +661,7 @@ class SEClampSomaticStimulus(ContinuousStimulusWithoutTimestamps):
         default=_DEFAULT_SIMULATION_LENGTH_MILLISECONDS / 4,
         title="Level 1 Duration",
         description="Time duration in milliseconds for the 1st Voltage level "
```
darshanmandge: `Duration 1 (dur1) of NEURON SEClamp stimulus (in ms)`. I would prefer not to use `aka` in text. Similarly for amplitudes: `Amplitude 1 (amp1) of NEURON SEClamp stimulus (in mV)`.

eleftherioszisis: > I like having a proper name that can be understood by anyone, even people that have never used NEURON, in the title. But we can still mention the NEURON name in the definition.

If that flexibility is acceptable, I agree 100%. We should use self-descriptive names which will improve UX.

PR #644, opened 2026-03-23
Introduce generic configs

eleftherioszisis:
* Add generic configs for the task endpoints
* Do not adapt simulation tasks because they require a migration first

See https://github.com/openbraininstitute/prod-build-circuit/issues/23

### DIFF
--- a/app/mappings.py
+++ b/app/mappings.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 
 from entitysdk import models
+from entitysdk.types import TaskActivityType, TaskConfigType
 from obp_accounting_sdk.constants import ServiceSubtype
 
 from app.config import settings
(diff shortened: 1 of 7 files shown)


james-isbister (dismissed):
LGTM!

james-isbister (commented):
LGTM - just one thing

james-isbister (approved):

james-isbister:
Ok I'd prefer to merge https://github.com/openbraininstitute/obi-one/pull/619 to main first, as:
- It is independent of the task manager (I can remove the task manager part I wrote yesterday, which you are implementing here) so can be reviewed as a separate PR
- The changes to EMSynapseMapping were mostly just reorganizing into separate files, and we will come back to this when we deploy it in the UI, so we can ignore this for now.
- It will allow @chr-pok to start using it for circuit-extraction (which will be the first proper test)
- We can make adjustments to the code in this and follow up PRs

PR #649, opened 2026-03-26
Mesh morphology endpoint

dkeller9:
The user can specify a morphology id, and the endpoint meshes it and registers the glb mesh with the morphology.


marwan-abdellah (commented):
I have reviewed the part of the mesh generation from the morphology, and it is perfectly fine. For the other changes, I would recommend someone to have a look to confirm. 

@dkeller9

pyproject.toml
```diff
@@ -48,6 +47,8 @@ dependencies = [
     "morph_spines",
     "morph-tool",
     "neurom",
+    "neuromorphomesh",
+    "numpy==2.0.0",
```
eleftherioszisis: Why is numpy pinned here? Having so strict constraints will quickly lead to dependency conflicts. Wouldn't `numpy>=2.0.0` work?

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+
+            glb_path_str = _mesh_swc(str(swc_path), output_directory=tmp_dir)
+            glb_path = Path(glb_path_str)
+
+            if not glb_path.exists() or glb_path.stat().st_size == 0:
+                msg = f"Meshing produced no output at {glb_path_str}"
+                raise RuntimeError(msg)  # noqa: TRY301
+
+            existing_glb_assets = [a for a in morph.assets if a.content_type == "model/gltf-binary"]
+            _delete_existing_glb_assets(db_client, cell_morphology_id, existing_glb_assets)
```
eleftherioszisis: It is not the responsibility of this endpoint to delete assets. It is quite unexpected to call a POST endpoint and under the hood to perform DELETEs. If it is not allowed in the case of an existing glb, it should raise instead.

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+            file_path=glb_path,
+            file_content_type="model/gltf-binary",
+            asset_label="cell_surface_mesh",
+        )
+    except entitysdk.exception.EntitySDKError as err:
+        L.error(f"Failed to upload GLB asset for {cell_morphology_id}: {err}")
+        raise HTTPException(
+            status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            detail={
+                "code": ApiErrorCode.INTERNAL_ERROR,
```
eleftherioszisis: Minor: Wouldn't it be more suitable if this was a ApiErrorCode.DATABASE_CLIENT_ERROR instead ?

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+                entity_id=cell_morphology_id,
+                entity_type=CellMorphology,
+                asset_id=asset.id,
+            )
+        except entitysdk.exception.EntitySDKError as err:
+            L.error(f"Failed to delete existing GLB asset {asset.id}: {err}")
+            raise HTTPException(
+                status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+                detail={
+                    "code": ApiErrorCode.INTERNAL_ERROR,
```
eleftherioszisis: Minor: Wouldn't it be more suitable if this was a ApiErrorCode.DATABASE_CLIENT_ERROR instead ?

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            swc_path = Path(tmp_dir) / f"{uuid.uuid4()}.swc"
+            swc_path.write_bytes(swc_bytes)
+
+            glb_path_str = _mesh_swc(str(swc_path), output_directory=tmp_dir)
+            glb_path = Path(glb_path_str)
+
+            if not glb_path.exists() or glb_path.stat().st_size == 0:
+                msg = f"Meshing produced no output at {glb_path_str}"
+                raise RuntimeError(msg)  # noqa: TRY301
```
eleftherioszisis: Minor: I would separate the two and return ApiErrors instead, one for existence and one for size. That way it would be easier to debug when meshing fails.

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+def register_morphology_mesh(
+    cell_morphology_id: str,
+    db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
+    virtual_lab_id: Annotated[str, Query(description="Virtual lab ID")],
+    project_id: Annotated[str, Query(description="Project ID")],
+) -> dict:
+    db_client.project_context = ProjectContext(
+        virtual_lab_id=virtual_lab_id,
+        project_id=project_id,
+    )
```
eleftherioszisis: Why not provide the correct vlab/proj in the request header instead of adding a second pair to the payload and mutating client?

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+        L.error(f"Failed to fetch entity {cell_morphology_id}: {err}")
+        raise HTTPException(
+            status_code=HTTPStatus.NOT_FOUND,
+            detail={
+                "code": ApiErrorCode.NOT_FOUND,
+                "detail": f"Cell morphology {cell_morphology_id} not found.",
+            },
+        ) from err
+
+    swc_asset = next((a for a in morph.assets if a.content_type == "application/swc"), None)
```
eleftherioszisis:
```suggestion
    swc_asset = db_client.select_assets(entity=morph, selection={"content_type": ContentType.application_swc}).one()
```
Use ContentType from entitysdk.types. I assume you want to enforce existence because you are checking below if it is None or not. With `one()` it will raise if it's not exactly 1 match.

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+            status_code=HTTPStatus.UNPROCESSABLE_ENTITY,
+            detail={
+                "code": ApiErrorCode.INVALID_REQUEST,
+                "detail": f"Cell morphology {cell_morphology_id} has no SWC asset.",
+            },
+        )
+
+    L.info(f"register_morphology_mesh: downloading SWC asset {swc_asset.id}")
+    try:
+        swc_bytes: bytes = db_client.download_content(
```
eleftherioszisis, line 180: You can also use fetch_content to leverage the local asset store if available.
```python

from entitysdk.types import FetchContentStrategy

swc_bytes = db_client.fetch_content(..., startegy=FetchContentStrategy.copy_or_download)
```

pyproject.toml
dkeller9: Any version before 2.0.0 is too old, but any numpy version after 2.0.0 incurs this error in one of the tests someone wrote:

FAILED tests/obi_one/scientific/blocks/neuron_sets/test_simplex.py::test_simplex_neuron_sets - AttributeError: module 'numpy' has no attribute 'in1d'
=========== 1 failed, 605 passed, 386 warnings in 115.89s (0:01:55) ============

make: *** [Makefile:103: test-docker] Error 1

eleftherioszisis (changes requested):

james-isbister:
Can the endpoint be created depending on whether the private package is installed? 

`make run-local` should work without the need for private repos to be installed, but I think that would solve the issue

dkeller9:
@danifr is there a way to do what @james-isbister is suggesting in the above comment?

james-isbister:
@dkeller9 You should be able to put the definition of the endpoint:
i.e.
```
`@router.post(
    "/convert-morphology-to-registered-mesh/{cell_morphology_id}",
    summary="Compute & register a GLB surface mesh for an existing morphology",
    description=(
        "Downloads the SWC asset for the given cell morphology, converts it to an annotated "
        "GLB mesh using neuromorphomesh, uploads the result as a new asset on the same entity, "
        "and returns the new asset id."
    ),
)
def register_morphology_mesh(
    cell_morphology_id: str,
    db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
    virtual_lab_id: Annotated[str, Query(description="Virtual lab ID")],
    project_id: Annotated[str, Query(description="Project ID")],
) -> dict:
    db_client.project_context = ProjectContext(
        virtual_lab_id=virtual_lab_id,
        project_id=project_id,
    )`
```

within an if or try/except statement which checks if the package is installed

@eleftherioszisis can probably advise on the best way to do something to that effect

pyproject.toml
danifr: in1d was deprecated since version 2.0: According to the docs you can use [isin](https://numpy.org/doc/2.3/reference/generated/numpy.isin.html#numpy.isin) instead of [in1d](https://numpy.org/doc/2.3/reference/generated/numpy.in1d.html#numpy.in1d) for new code.

dkeller9:
Thanks, I will try to implement something based on this suggestion.

app/endpoints/convert_morphology_to_registered_mesh.py
james-isbister: Yes, db_client should already have the project context from Depends(get_client) so passing the lab/project id and setting the project context shouldn't be needed

eleftherioszisis:
> @dkeller9 You should be able to put the definition of the endpoint: i.e.
> 
> ```
> `@router.post(
>     "/convert-morphology-to-registered-mesh/{cell_morphology_id}",
>     summary="Compute & register a GLB surface mesh for an existing morphology",
>     description=(
>         "Downloads the SWC asset for the given cell morphology, converts it to an annotated "
>         "GLB mesh using neuromorphomesh, uploads the result as a new asset on the same entity, "
>         "and returns the new asset id."
>     ),
> )
> def register_morphology_mesh(
>     cell_morphology_id: str,
>     db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
>     virtual_lab_id: Annotated[str, Query(description="Virtual lab ID")],
>     project_id: Annotated[str, Query(description="Project ID")],
> ) -> dict:
>     db_client.project_context = ProjectContext(
>         virtual_lab_id=virtual_lab_id,
>         project_id=project_id,
>     )`
> ```
> 
> within an if or try/except statement which checks if the package is installed
> 
> @eleftherioszisis can probably advise on the best way to do something to that effect

Leave that for the very end. I want to check if there is a better way by allowing installing the deps even locally instead of making everything optional. If not possible we can consider this way.

.github/workflows/publish.yml
```diff
@@ -20,6 +20,22 @@ jobs:
...
+      - name: Get CodeArtifact token
+        run: |
+          CODEARTIFACT_AUTH_TOKEN=$(aws codeartifact get-authorization-token \
+            --domain openbraininstitute \
+            --query authorizationToken \
+            --output text)
+          echo "::add-mask::$CODEARTIFACT_AUTH_TOKEN"
+          echo "UV_INDEX_OBI_CODEARTIFACT_PASSWORD=$CODEARTIFACT_AUTH_TOKEN" >> $GITHUB_ENV
+          echo "UV_INDEX_OBI_CODEARTIFACT_USERNAME=aws" >> $GITHUB_ENV
+
```
GianlucaFicarelli, line 41: As a reminder, this PR shouldn't be merged until the Docker image is made private.
It can be done in this PR, or even better in a separate PR to check that everything works as expected before adding a private package.

james-isbister:
`I want to check if there is a better way by allowing installing the deps even locally instead of making everything optional.`

I was hoping that you wouldn't need to be part of openbraininstitute or use "make run-local" so others can in theory develop with us

GianlucaFicarelli (dismissed):

eleftherioszisis:
> `I want to check if there is a better way by allowing installing the deps even locally instead of making everything optional.`
> 
> I was hoping that you wouldn't need to be part of openbraininstitute or use "make run-local" so others can in theory develop with us

Yeah, if we need external collaborators to work with the code my approach won't work unfortunately. Therefore, we are back at lazy imports/skipping tests..

pyproject.toml
danifr: the numpy test related error is due to ConnectomeUtilities

I thought it was gonna be a nice and easy change but I ended up needing to touch quite a lot of lines of code in order to make the CI tests pass: https://github.com/openbraininstitute/ConnectomeUtilities/pull/5

I assigned the PR to Michael but maybe James or Christoph could also take a look (since they are contributors to that repo)

.github/workflows/publish.yml
danifr: This is a valid concern. Thanks for pointing it out.  I can prepare a fix similar to what I have done for Blueness here https://github.com/openbraininstitute/Bluenaas/pull/150/changes#diff-bcecc648a0fede4c0ce18d730c5d88b4007fa0144bbb1b93d7ab7ac16a7da1eeR28

Use `--mount=type=secret` and also in this case remove the ARG variables containing sensitive info.

@dkeller9 do you mind if I add another commit to this PR?

dkeller9: @danifr please go ahead with your additional commit, thank you.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: The endpoint has been changed to now return an error if there are existing glb assets.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: ApiErrorCode.DATABASE_CLIENT_ERROR is now used.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: ApiErrorCode.DATABASE_CLIENT_ERROR is now used.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: The separation has now been performed.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: Good point, now corrected.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: This suggestion has now been included.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: I would like to leave the retrieve as it is since performance is already fine.

dkeller9:
I've added the conditional use of the private repository as @james-isbister suggested. @eleftherioszisis most of your comments have now been addressed, please verify if okay.

app/endpoints/convert_morphology_to_registered_mesh.py
eleftherioszisis: This:
```python
    existing_glb_asset = next(
        iter(
            db_client.select_assets(
                entity=morph,
                selection={"content_type": "model/gltf-binary"},
            )
        ),
        None,
    )
```
should be either:
```python
existing_glb_asset = db_client.select_assets(
    entity=morph,
    selection={"content_type": ContentType.model_gltf_binary},
).first()
```
if you want to have it returning None if no match or the first result if it matches 1 or more.

Or:
```python
existing_glb_asset = db_client.select_assets(
    entity=morph,
    selection={"content_type": ContentType.model_gltf_binary},
).one()
```
if you want it to match exactly one and raise for 0 or more than 1.

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,213 @@
...
+    L.info(
+        f"register_morphology_mesh: uploading GLB asset for {cell_morphology_id} "
+        f"({glb_path.stat().st_size} bytes)"
+    )
+    try:
+        return db_client.upload_file(
+            entity_id=cell_morphology_id,
+            entity_type=CellMorphology,
+            file_path=glb_path,
+            file_content_type="model/gltf-binary",
```
eleftherioszisis: Use ContentType StrEnum from entitysdk.types

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,213 @@
...
+            asset_label="cell_surface_mesh",
```
eleftherioszisis: Use AssetLabel StrEnum from entitysdk.types

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,213 @@
...
+    if not HAS_MESHING:
+        raise HTTPException(
+            status_code=HTTPStatus.NOT_IMPLEMENTED,
+            detail="Meshing dependencies are not installed on this instance.",
+        )
+    try:
+        morph = db_client.get_entity(entity_id=cell_morphology_id, entity_type=CellMorphology)
+    except entitysdk.exception.EntitySDKError as err:
+        L.error(f"Failed to fetch entity {cell_morphology_id}: {err}")
+        raise HTTPException(
```
eleftherioszisis: Use ApiError from app/errors.py

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,213 @@
...
+        "and returns the new asset id."
+    ),
+)
+def register_morphology_mesh(
+    cell_morphology_id: str,
+    db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
+) -> dict:
+    if not HAS_MESHING:
+        raise HTTPException(
+            status_code=HTTPStatus.NOT_IMPLEMENTED,
```
eleftherioszisis: Use ApiError from app/errors.py

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,213 @@
...
+        ) from err
+    L.info(f"register_morphology_mesh: checking for existing GLB assets on {cell_morphology_id}")
+    _check_no_existing_glb_assets(db_client, cell_morphology_id, morph)
+
+    swc_asset = db_client.select_assets(
+        entity=morph, selection={"content_type": ContentType.application_swc}
+    ).one()
+    if swc_asset is None:
+        L.error(f"No SWC asset found on morphology {cell_morphology_id}")
+        raise HTTPException(
```
eleftherioszisis: Use ApiError from app/errors.py

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,213 @@
...
+    L.info(f"register_morphology_mesh: downloading SWC asset {swc_asset.id}")
+    try:
+        swc_bytes: bytes = db_client.download_content(
+            entity_id=cell_morphology_id,
+            entity_type=CellMorphology,
+            asset_id=swc_asset.id,
+        )
+    except entitysdk.exception.EntitySDKError as err:
+        L.error(f"Failed to download SWC asset {swc_asset.id}: {err}")
+        raise HTTPException(
```
eleftherioszisis: Use ApiError from app/errors.py

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,213 @@
...
+            status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            detail={
+                "code": ApiErrorCode.INTERNAL_ERROR,
+                "detail": "Failed to download the SWC asset.",
+            },
+        ) from err
+
+    if not swc_bytes:
+        raise HTTPException(
```
eleftherioszisis: same here

uv.lock
GianlucaFicarelli, line 1: The only update to uv.lock in this PR should be related to the new installed package `neuromorphomesh`.
For this, it should be possible to revert the uv.lock file, and execute `make compile-deps` instead of `make upgrade-deps`.

pyproject.toml
GianlucaFicarelli: If this PR needs to be merged before the fix in ConnectomeUtilities, this dependency should be changed from `numpy==2.0.0` to `numpy<2.4.0` (with a comment for the reason), since in1d has been removed in https://numpy.org/doc/2.4/release/2.4.0-notes.html#removed-numpy-in1d but it was still available until then.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: Now the AssetLabel strenum is used.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: Now using ContentType StrEnum

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: Now using ApiError instead.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: Now using ApiError instead.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: Now using ApiError instead.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: Now using ApiError instead.

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: Now using ApiError instead.

uv.lock
dkeller9: The `uv.lock` file should now be the result of `make compile-deps` run on the `uv.lock` from main.

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,206 @@
...
+) -> None:
+    existing_glb_asset = next(
+        iter(
+            db_client.select_assets(
+                entity=morph,
+                selection={"content_type": ContentType.model_gltf_binary},
+            )
+        ),
+        None,
+    )
```
eleftherioszisis: Please see my comment: https://github.com/openbraininstitute/obi-one/pull/649#discussion_r3079131508

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: This is now changed to the .first() method

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,201 @@
...
+            http_status_code=HTTPStatus.NOT_FOUND,
+        ) from err
+
+    L.info(f"register_morphology_mesh: checking for existing GLB assets on {cell_morphology_id}")
+    _check_no_existing_glb_assets(db_client, cell_morphology_id, morph)
+
+    swc_asset = db_client.select_assets(
+        entity=morph, selection={"content_type": ContentType.application_swc}
+    ).one()
+    if swc_asset is None:
```
eleftherioszisis: `one()` will raise if there is no asset found, thererfore the `if swc_asset is None` will never be reached. You can catch the error and raise the helpful ApiError.

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+        ) from err
+
+    L.info(f"register_morphology_mesh: checking for existing GLB assets on {cell_morphology_id}")
+    _check_no_existing_glb_assets(db_client, cell_morphology_id, morph)
+
+    try:
+        swc_asset = db_client.select_assets(
+            entity=morph, selection={"content_type": ContentType.application_swc}
+        ).one()
+    except (entitysdk.exception.NotFoundError, entitysdk.exception.NoResultFoundError) as err:
```
eleftherioszisis: These errors do not exist. Use IteratorResultError instead.

dkeller9: Now using the IteratorResultError.

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,201 @@
...
+            error_code=ApiErrorCode.NOT_FOUND,
+            http_status_code=HTTPStatus.NOT_FOUND,
+        ) from err
+
+    L.info(f"register_morphology_mesh: checking for existing GLB assets on {cell_morphology_id}")
+    _check_no_existing_glb_assets(db_client, cell_morphology_id, morph)
+
+    try:
+        swc_asset = db_client.select_assets(
+            entity=morph, selection={"content_type": ContentType.application_swc}
```
GianlucaFicarelli: I think it would be safer and more explicit to filter also by `"label": AssetLabel.morphology` just in case we'll add other asset labels. At the moment there are only:
- AssetLabel.morphology
- AssetLabel.morphology_with_spines
- AssetLabel.cell_surface_mesh

and content_type swc is allowed only for AssetLabel.morphology.

dkeller9: Good catch, the additional label filter is now added.

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+    summary="Compute & register a GLB surface mesh for an existing morphology",
+    description=(
+        "Downloads the SWC asset for the given cell morphology, converts it to an annotated "
+        "GLB mesh using neuromorphomesh, uploads the result as a new asset on the same entity, "
+        "and returns the new asset id."
+    ),
+)
+def register_morphology_mesh(
+    cell_morphology_id: str,
+    db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
```
eleftherioszisis:
```suggestion
    db_client: DatabaseClientDep
```
importing it from app/dependecies/entitysdk.py

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+    "/convert-morphology-to-registered-mesh/{cell_morphology_id}",
+    summary="Compute & register a GLB surface mesh for an existing morphology",
+    description=(
+        "Downloads the SWC asset for the given cell morphology, converts it to an annotated "
+        "GLB mesh using neuromorphomesh, uploads the result as a new asset on the same entity, "
+        "and returns the new asset id."
+    ),
+)
+def register_morphology_mesh(
+    cell_morphology_id: str,
```
eleftherioszisis: I would use uuid.UUID so that the string is validated to be a valid UUID instead of any string.

dkeller9: Now changed to uuid.UUID

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: This change has been made.

app/endpoints/convert_morphology_to_registered_mesh.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def _check_no_existing_glb_assets(
+    db_client: entitysdk.client.Client,
+    cell_morphology_id: str,
+    morph: CellMorphology,
+) -> None:
+    existing_glb_asset = db_client.select_assets(
+        entity=morph,
+        selection={"content_type": ContentType.model_gltf_binary},
```
eleftherioszisis: Minor: Similar to what @GianlucaFicarelli mentioned below, we can make this more specific too by adding the label.

```suggestion
        selection={"content_type": ContentType.model_gltf_binary, "label": AssetLabel.cell_surface_mesh},
```

app/endpoints/convert_morphology_to_registered_mesh.py
dkeller9: The error is now caught and ApiError used.

pyproject.toml
dkeller9: It is set to `numpy<2.4.0` until the fix in ConnectomeUtilities is merged.

dkeller9:
@eleftherioszisis @GianlucaFicarelli thank you for the great suggestions. Are they now resolved to your satisfaction?

uv.lock
GianlucaFicarelli: uv.lock looks good now, there is just a difference for bluepyefe updated to 2.3.53, but this difference should disappear after merging with main, that now specifies `bluepyefe>=2.3.53`

james-isbister:
Thank you for making the service runnable without the private package. All good on my side, but I'll leave @GianlucaFicarelli or @eleftherioszisis to approve

pyproject.toml
```diff
@@ -47,6 +47,7 @@ dependencies = [
     "morph_spines",
     "morph-tool",
     "neurom",
+    "numpy<2.4.0",
```
GianlucaFicarelli, line 50: This line can be removed because it's now included in the main branch already.

.github/workflows/publish.yml
```diff
@@ -40,13 +42,15 @@ jobs:
       - name: Authenticate with AWS ECR account for pushing the image
         uses: aws-actions/configure-aws-credentials@v4
```
danifr: Not a showstopper but there is a regression introduced after the conflict resolution, this should be `configure-aws-credentials@v6`

danifr: All aws-actions/configure-aws-credentials should use v6

dkeller9: now updated.

eleftherioszisis (approved):

tests/obi_one/core/test_entity_from_id.py
```diff
@@ -13,9 +13,8 @@ def test_file_value(self):
 
 class TestEntityFromIDStr:
     def test_str_representation(self):
-        """Test __str__ without needing a real entity class."""
-
         class ConcreteEntityFromID(EntityFromID):
+            type: str = "EntityFromID"
```
GianlucaFicarelli, line 17: It's weird that it's needed to add an explicit `type` to this class and all the others below, while the code related to these tests haven't been touched by this PR 🤔 
It could be related to the use of patch together with [this code](https://github.com/openbraininstitute/obi-one/blob/7184f91abf91567977c5917a0c2832e4984d8fe2/obi_one/core/base.py#L20-L24), as seen already in https://github.com/openbraininstitute/obi-one/pull/734/changes
The fix in that case was to use monkeypatch instead of patch from unittest.mock

GianlucaFicarelli (commented):
Although already merged I had a couple of pending comments, but overall looks ok to me.

pyproject.toml
dkeller9: fixed in a separate PR https://github.com/openbraininstitute/obi-one/pull/746

PR #655, opened 2026-03-29
New Neuron Sets

james-isbister:
Todo: Update new reference functional logic in ScanConfig, BlockReferences etc


tests/ui_schema/validate_block.py
g-bar: Perhaps we can update this test to check that all elements of the union, except the last `None` are a valid refs?

james-isbister: Yes, good idea, will add it at some point

docs/gui-definition-spec/components/neuron_property_filter/neuron_property_filter.md
```diff
@@ -0,0 +1,131 @@
...
+`obi_one/scientific/blocks/neuron_sets_2/property.py`). It lets the user build
+one or more filters over node properties of the population selected on the
+parent block.
+
+`PropertyNeuronSet` is a subclass of `PopulationNeuronSet`, which exposes a
+user-selectable `population` field rendered as an
+[entity_property_dropdown](../entity_property_dropdown/entity_property_dropdown.md).
+The set of properties and the set of accepted values for each property are
+both scoped to that selected population — the UI must re-read its options
+whenever `population` changes.
```
g-bar, line 15: Given this dependency
We should add a check to the validation scripts for 'neuron_property_filter' ui_element , that the parent block must always also contain a 'population' key with a `entity_property_dropdown` ui_element

alternatively since they seem to always come together and both are required, we could combine them into a single ui_element?

docs/gui-definition-spec/components/neuron_property_filter/neuron_property_filter.md
```diff
@@ -0,0 +1,131 @@
...
+        title="Filter",
+        description="Filter dictionary. The list per key is NOT a multi-dimensional "
+                    "parameter sweep but the set of accepted values for the property, "
+                    "e.g. {'layer': ['2', '3']}.",
+    )
+
+
+class PropertyNeuronSet(PopulationNeuronSet):
+
+    property_filter: NeuronPropertyFilter | list[NeuronPropertyFilter] = Field(
```
g-bar: I don't understand the distinction between

`NeuronPropertyFilter` and `list[NeuronPropertyFilter] with a single element`

the second one is "sweep" but sweeping over 1 element is functionally identical to the single value

so this could be simplified as

`list[NeuronPropertyfilter]` and enforce a min_length of 1

docs/gui-definition-spec/components/neuron_property_filter/neuron_property_filter.md
james-isbister: I liked the idea of a single ui element and just started implementing it, but I then remembered that population is inhertited by PropertyNeuronSet from PopulationNeuronSet where we'll handle validation of the "population" and any other shared related logic, so better to keep it as is.

I can update the tests at some point!

docs/gui-definition-spec/components/neuron_property_filter/neuron_property_filter.md
james-isbister: Two reasons for this:
1) On the backend side, a list of length one can still be used to represent a special case where we want sub-directories to be created as if it were a scan dimension
2) More importantly, when the ScanConfig is cast to a SingleConfig, there is only one NeuronPropertyFilter accesible through property_filter rather than property_filter[0]. So the same story as float | list[float]

g-bar: Not sure I follow, 

It seems to me that since they're functionally identical, and the interface is the exact same (at least for the float_parameter_sweep they're different), this comes down to an internal implementation detail of obi-one, it's weird that that's driving the external data interface.

So when you iterate over the scan config containiing a `list[float]`:
you iterate through it to generate the individual configs, but iterating over a list with one element yields the same result, so the `float | list[float]` actually requires extra logic to differentiate the cases. 

Or alternatively

You could abstract the logic so that if it's a list[float] with one element you extract it into a float internally, but externally it's always list[float]?

g-bar: In other words: How do you handle the case where the client sends a list with one element?

james-isbister: > It seems to me that since they're functionally identical, and the interface is the exact same (at least for the float_parameter_sweep they're different), this comes down to an internal implementation detail of obi-one, it's weird that that's driving the external data interface.

It's true on the frontend side there are only two cases during creation (single value, list of length > 1), but the python library intentionally offers the third case (list of length 1) to have a particular desired effect on the outputted directory structure. The UI should still be able to read in a config with a list of length 1 actually

> So when you iterate over the scan config containiing a list[float]:
>you iterate through it to generate the individual configs, but iterating over a list with one element yields the same result, so the float | list[float] actually requires extra logic to differentiate the cases.

As above, we intentionally have extra logic to differentiate the two cases

> Or alternatively
> 
> You could abstract the logic so that if it's a list[float] with one element you extract it into a float internally, but externally it's always list[float]?

Yes, this would have been nice, but I didn't want to get into overriding python types or using getters for every float. Using [0] would have been easier

g-bar: I see, couple more questions if you don't mind, just so that I fully understand.

> but the python library intentionally offers the third case (list of length 1) to have a particular desired effect on the outputted directory structure. The UI should still be able to read in a config with a list of length 1 actually

I could potentially send a list of length 1 instead of a single value and the backend handles it just fine (except for some effect in the directory structure?) but the configs are the exact same?

In other words there seems to be a choice between either of this paradigms:

- float and list[float] with a single value are equivalent (functionally, despite internal directory structure), then float is redundant and we could reduce it to list[float].

- float and list[float] with length one are not equivalent and the latter has "undesired"? effects so we should explicitly disallow lists of size 1. 

- float and list[float] with a single value are not equivalent, then we should make it explicit in the schema and the UI? If the user doesn't care about that distinction then we're back to case one and just make it list[float]?

james-isbister: Of course. Actually on the frontend side there would be an effect because in the second tab (i.e. "Simulations"), the cards for each coordinate on the left hand side show the values of the parameters which are being scanned over.

Putting a single value in a list triggers the parameter to be treated as a "scan parameter" in obi-one, which represented as such in the persisted obi_one_scan.json and obi_one_coordinate.json assets, as well as the scan_parameters properties of the campaign and single config entities in entitycore. Not sure which of these the frontend uses for the display

g-bar: Sure, so is that effect in the simulations tab desirable ? or not?

If it's not, then we must explicitly disallow lists of lengths one in the schema. (minLength=2), 
because even if the frontend doesn't send them, the ML can produce it since it's  schema valid or a user interacting with obi-one directly through the rest api can also send a list of length 1.

If it's desirable then we can just make it list[float]? 

So we have the choice between list[float] or float | list[float] with  minLength = 2, leaving as is opens the door for edge cases where scan configs even with a single sim config, sometimes show up as parameter sweep in the simulations tab.

In fact we probably *must* add a min length to the array anyway,  as it currently is an empty array is valid by the schema too, that probably breaks sth.

james-isbister: I've annotated list[NeuronPropertyFilter] with min_length=1. The case where the length is 1 could in theory be created through the python package (for the directory structure "benefit") with the results also registered in entitycore. 

It's a fair point that the agent could also create a list of length 1 but in general it gets the logic that lists should only be used when there is more than one value. Even if it did create a list of length one, it wouldn't break anything in the coordinates tab, as its just treated as if it was a scan parameter (even though there's only one value).

docs/gui-definition-spec/components/neuron_property_filter/neuron_property_filter.md
```diff
@@ -0,0 +1,131 @@
...
+  a multi-value parameter sweep — it is the set of values that match this
+  property (e.g. `{"layer": ["2", "3"]}` means "layer is 2 or 3"). Sweeps
+  over filters are expressed at the outer level by providing
+  `list[NeuronPropertyFilter]`.
+
+### Example Pydantic implementation
+
+```py
+class NeuronPropertyFilter(OBIBaseModel):
+    filter_dict: dict[str, list] = Field(
```
g-bar: This should be

```py
filter_dict: dict[str, list[str]] = Field(
``` 

so that only a list of strings is valid in the filter_dict
And ensure that's the case in the validation scripts.

docs/gui-definition-spec/components/neuron_property_filter/neuron_property_filter.md
g-bar: The min_length=1 should be checked in the validation_script before merging

docs/gui-definition-spec/components/neuron_property_filter/neuron_property_filter.md
james-isbister: @chr-pok should it be list[str | int | float]? I forget whether some of the properties like layer are represented as strings or integers?

chr-pok: Indeed, layer could be either int (older circuits) or str (newer circuits). In general, I think it should be `list[str | int]` which should cover all categorical types. Not float, since the property filter does not filter based on ranges.

obi_one/scientific/unions/unions_neuron_sets_2.py
```diff
@@ -0,0 +1,120 @@
...
+    VirtualIDNeuronSet,
+)
+from obi_one.scientific.blocks.neuron_sets_2.property import (
+    BiophysicalPropertyNeuronSet,
+    VirtualPropertyNeuronSet,
+    PointPropertyNeuronSet,
+)
+from obi_one.scientific.blocks.neuron_sets.predefined import PredefinedNeuronSet
+
+_BIOPHYSICAL_NEURON_SETS = BiophysicalIDNeuronSet | BiophysicalPropertyNeuronSet | PredefinedNeuronSet
```
chr-pok: Not exactly sure how to deal with `PredefinedNeuronSet`. It could be anything, not necessarily biophysical. I think we should create predefined subtypes, like `PredefinedBiophysicalNeuronSet`, etc.

docs/gui-definition-spec/components/neuron_property_filter/neuron_property_filter.md
g-bar: Can there be `ints` and `str` in the same property list? or it can be 

```py
filter_dict: dict[str, list[str] | list[int]] = Field
```

james-isbister: Yes, that's better. Have pushed the change

obi_one/scientific/blocks/neuron_sets_2/property.py
```diff
@@ -106,7 +106,7 @@ class PropertyNeuronSet(NeuronSet):
             SchemaKey.UI_ELEMENT: UIElement.NEURON_PROPERTY_FILTER,
             SchemaKey.PROPERTY_GROUP: MappedPropertiesGroup.CIRCUIT,
             SchemaKey.PROPERTY: CircuitMappedProperties.NODE_PROPERTY_UNIQUE_VALUES_BY_POPULATION,
-            "population_source_dropdown": "population",
+            "population_source_dropdown_key": "population",
```
g-bar: This should be formalized in the spec and reference schema, as well as checked in the validation scripts.

obi_one/scientific/blocks/neuron_sets_2/property.py
```diff
@@ -95,7 +95,7 @@ class PropertyNeuronSet(NeuronSet):
             SchemaKey.UI_ELEMENT: UIElement.ENTITY_PROPERTY_DROPDOWN,
             SchemaKey.PROPERTY_GROUP: MappedPropertiesGroup.CIRCUIT,
             SchemaKey.PROPERTY: CircuitMappedProperties.BIOPHYSICAL_NEURONAL_POPULATION,
-            "filter_key": "property_filter",
+            "property_filter_key": "property_filter",
```
g-bar: Same here

chr-pok (approved):
Looks good and tested locally. More tests to be done once merged into main.

PR #672, opened 2026-04-07
Support multi-neuron input in EM synapse mapping tasks

### LINKED ISSUE #701: Support multi-neuron input in EM synapse mapping tasks
## Description
We should support building a circuit with an arbitrary number of skeletonized morphologies.

The current EM synapse mapping task is currently delivering the synaptome of a single neuron coming from the MICrONS dataset. We should extend this functionality to support multiple MICrONS neurons as input. The output is a multi-neuron circuit with physical / virtual node populations. Connections are chemical if connecting two physical neurons; otherwise they're virtual. Virtual connections represent the incoming (pre) inputs to the physical nodes.

## Acceptance criteria
- [ ] We can build a circuit out of one or multiple skeletonized morphologies with the structure explained above.

jplanasc:
The current EM synapse mapping task is currently delivering the synaptome of a single neuron coming from the MICrONS dataset. This PR extends this task to support multiple MICrONS as input. The output is a multi-neuron circuit with physical / virtual node populations. Connections are chemical if connecting two physical neurons; otherwise they're virtual. Virtual connections represent the incoming (pre) inputs to the physical nodes.

Fixes #701


obi_one/scientific/tasks/em_synapse_mapping/config.py
```diff
@@ -103,3 +106,61 @@ class EMSynapseMappingSingleConfig(EMSynapseMappingScanConfig, SingleConfigMixin
...
+
+class EMSynapseMappingMultipleConfig(OBIBaseModel):
+    """Config for multi-neuron synapse mapping."""
+
+    name: ClassVar[str] = "Map synapse locations (multi-neuron)"
+    description: ClassVar[str] = (
+        "Map afferent synapses from EM onto multiple spiny morphologies, "
+        "producing a circuit with physical internal connections and virtual external inputs."
+    )
+    cave_token: str | None = Field(default=None, title="CAVEclient access token")
```
eleftherioszisis: secrets should be fetched from the environment to which they are injected. They should never be visible in configs, logs, etc. If they need to be present in a pydantic model, it is usually better to use `SecretStr` to store them. See https://docs.pydantic.dev/2.0/usage/types/secrets/

When running the task, the env var `CAVECLIENT_MICRONS_API_KEY` will be available.
https://github.com/openbraininstitute/prod-build-circuit/issues/22#issuecomment-4177538773

james-isbister: Yes, this can be removed from the config as it is in main. 

Also the pt_root_id no longer needs to be in the config and can be fetched through the entity as it is in main

obi_one/scientific/tasks/em_synapse_mapping/config.py
```diff
@@ -103,3 +106,61 @@ class EMSynapseMappingSingleConfig(EMSynapseMappingScanConfig, SingleConfigMixin
...
+        description=("A spiny neuron morphology from an EM dataset."),
+    )
+    pt_root_id: int | None = Field(
+        title="Neuron identifier within the EM dense reconstruction dataset.",
+        description="If not provided, it will be inferred from the provenance of the morphology.",
+        default=None,
+    )
+
+
+class EMSynapseMappingMultipleConfig(OBIBaseModel):
```
james-isbister: I don't think its immediate as we're focussed on the generation of the circuit atm but the recent changes in main seem to have been lost: 
https://github.com/openbraininstitute/obi-one/blob/main/obi_one/scientific/tasks/em_synapse_mapping/config.py

i.e. InfoScanConfig, SingleConfig, json_schema_extra_additions, _campaign_task_config_type, _campaign_generation_task_activity_type, input_entities, json_schema_extra, _single_task_config_type, _single_task_activity_type

obi_one/scientific/tasks/em_synapse_mapping/config.py
```diff
@@ -103,3 +106,61 @@ class EMSynapseMappingSingleConfig(EMSynapseMappingScanConfig, SingleConfigMixin
...
+
+
+class NeuronEntry(OBIBaseModel):
+    """A single neuron entry for the multi-neuron synapse mapping task."""
+
+    neuron: CellMorphologyFromID | MEModelFromID = Field(
+        title="EM skeletonized morphology",
+        description=("A spiny neuron morphology from an EM dataset."),
+    )
+    pt_root_id: int | None = Field(
```
james-isbister: I think we don't need this NeuronEntry wrapper - we discussed with @MWolfR that we would only support morphologies / me models with derived from a cell mesh with a pt_root_id

obi_one/scientific/tasks/em_synapse_mapping/config.py
```diff
@@ -103,3 +106,61 @@ class EMSynapseMappingSingleConfig(EMSynapseMappingScanConfig, SingleConfigMixin
...
+    name: ClassVar[str] = "Map synapse locations (multi-neuron)"
+    description: ClassVar[str] = (
+        "Map afferent synapses from EM onto multiple spiny morphologies, "
+        "producing a circuit with physical internal connections and virtual external inputs."
+    )
+    cave_token: str | None = Field(default=None, title="CAVEclient access token")
+    coordinate_output_root: Path = Field(title="Output directory")
+
+    class Initialize(Block):
+        neurons: list[NeuronEntry] = Field(
```
james-isbister: Again, just a configuration / UI detail but using list notation in a scan config represents that we want to scan over multiple neurons

We'll need a new UI Element for selecting multiple entities simultaneously. I'll think about how we might represent this

james-isbister (commented):
Just some comments about the config

We can leave this until later though once the task code is finalised though

obi_one/scientific/tasks/em_synapse_mapping/config.py
jplanasc: Thanks for the feedback @eleftherioszisis  and @james-isbister , it should be fixed now.

obi_one/scientific/tasks/em_synapse_mapping/config.py
jplanasc: Oh sorry, didn't think about the UI, I was only thinking about the notebooks.

I can also think on my side how this could be approached alternatively.

jplanasc: @james-isbister , would a Tuple be better, as per:

https://github.com/openbraininstitute/obi-one/blob/735ce6a8af1a9bcde31b5fc0f5431fe6132b50e8/obi_one/core/block.py#L13-L19

?

obi_one/scientific/tasks/em_synapse_mapping/config.py
jplanasc: Hmmm... I've both checked with `diff` and "human, manual diff" and I don't see lost changes in `config.py`?

obi_one/scientific/tasks/em_synapse_mapping/config.py
jplanasc: Ok, good point. I removed it.

Thanks!

obi_one/scientific/tasks/em_synapse_mapping/config.py
```diff
@@ -103,3 +106,48 @@ class EMSynapseMappingSingleConfig(EMSynapseMappingScanConfig, SingleConfigMixin
...
+    description: ClassVar[str] = (
+        "Map afferent synapses from EM onto multiple spiny morphologies, "
+        "producing a circuit with physical internal connections and virtual external inputs."
+    )
+    coordinate_output_root: Path = Field(title="Output directory")
+
+    class Initialize(Block):
+        # We use a tuple instead of a list to avoid getting it taken as scan dimensions in the
+        # scan config.
+        neurons: tuple[CellMorphologyFromID | MEModelFromID, ...] = Field(
```
james-isbister: Looks good. we could also eventually extend this to:

> tuple[CellMorphologyFromID | MEModelFromID, ...] | list[tuple[CellMorphologyFromID | MEModelFromID, ...]]

To allow scans over sets of neurons, to build multiple circuits in one workflow. Although this isn't needed now.

james-isbister (commented):
Looks great. I would merge the two separate tasks into one config and task which supports either any number of neurons >= 1, for several reasons:
1) User experience:
- There is one less workflow to overload the user's attention with and unnecessarily "learn" in the UI
- The user can select 
2) Reduced code maintenance
3) Guaranteed consistency between the circuits produced for 1 neuron and multiple neurons.
4) Guaranteed consistency between the Python interfaces for 1 neuron and multiple neurons. Someone can quickly run a quick test for a single neuron before extending to multiple.
5) Task manager: we'd unnecessarily need to specify and test another task in the task manager [mappings](https://github.com/openbraininstitute/obi-one/blob/jplanas/multineuron_microns_circuit/app/mappings.py)
6) UI duplication: someone can duplicate a previously run workflow for multiple neurons (based on the serialization of the scan configuration in entitycore) and run it for a single neuron


Also, to get the multiple EM synapse mapping workflow ready for the UI, we'd need to copy the pattern of UIElements, and other boilerplate stuff from the single neuron case. Another reason to merge the two imo

obi_one/scientific/tasks/em_synapse_mapping/config.py
james-isbister: I see there are two configs and two tasks now. I was looking for these in the new config:
`InfoScanConfig, SingleConfig, json_schema_extra_additions, _campaign_task_config_type, _campaign_generation_task_activity_type, input_entities, json_schema_extra, _single_task_config_type, _single_task_activity_type`

jplanasc:
Thanks @james-isbister for the comments. Initially, I didn't want to change the original code, but since I think now the code has been validated and @MWolfR gave the ok on the resulting output, I will merge the 2 pathways.

With respect to:
> Looks good. we could also eventually extend this to:
> 
>    `tuple[CellMorphologyFromID | MEModelFromID, ...] | list[tuple[CellMorphologyFromID | MEModelFromID, ...]]`
> 
> To allow scans over sets of neurons, to build multiple circuits in one workflow. Although this isn't needed now.

If you're ok with that, I'd tackle it into a separate issue?

james-isbister:
`If you're ok with that, I'd tackle it into a separate issue?`
Actually @AurelienJaquier is going to handle that [here](https://github.com/openbraininstitute/prod-build-ion-channel-model/issues/106)

obi_one/scientific/tasks/em_synapse_mapping/config.py
```diff
@@ -40,13 +41,15 @@ class EMSynapseMappingScanConfig(InfoScanConfig):
...
-            description="""A neuron morphology with spines obtained from an electron-microscopy
-            datasets through the skeletonization task.""",
+        # We use a tuple instead of a list to avoid getting it taken as scan dimensions
+        # in the scan config.
+        neurons: tuple[CellMorphologyFromID | MEModelFromID, ...] = Field(
+            title="Neurons",
+            description="Neurons to include in the circuit (>= 1).",
+            min_length=1,
             json_schema_extra={
                 SchemaKey.UI_ELEMENT: UIElement.MODEL_IDENTIFIER,
```
james-isbister: We'll need a new UIElement here to signal to the frontend that the user can select multiple entities. It could be something like "UIElement.MODEL_IDENTIFIER_MULTI", but we can always change this later

We would ideally need something like, this to go with it:

> SchemaKey.ENTITY_QUERY_MULTI: {
>        [{"type": EntityType.cell_morphology,
>         "filters": {"cell_morphology_protocol": "ultraliser"}},
>         {"type": EntityType.me_model,
>         "filters": {...}},

To specify how to filter the suitable entities. Not sure if we have the final entitycore filters @MWolfR ?

Or we can just hard code this in the frontend for next week i.e. the user selects the entities before they get to configuration and then they are set in stone

examples/H_map_em_synapses/map_em_synapses.ipynb
james-isbister, line 1: I think keeping the two notebooks is nice, as you asked @jplanasc

obi_one/scientific/tasks/em_synapse_mapping/config.py
jplanasc: @james-isbister should this be part of this PR or in a separate one?

jplanasc:
@james-isbister 

> I would merge the two separate tasks into one config and task which supports either any number of neurons >= 1.

Done and tested in CI + the 2 notebooks of em synapse mapping.

obi_one/scientific/tasks/em_synapse_mapping/config.py
james-isbister: As the initial goal for the UI is now only to allow selection of one morphology, I think for now we can just change the value of UIElement to "em_synapse_mapping_single_morphology_temporary" in this PR, and the filter to:
`"filters": {"has_segmented_spines": True}},`

jplanasc: I discussed with @bilalesi , I think these are the changes he wanted:

https://github.com/openbraininstitute/obi-one/pull/672/changes/4c5caf39b92e5ab9480888b3de1bbf924ce05f5b

jplanasc:
Additional checks: successfully launched a task through the launch system.

eleftherioszisis (approved):

Issue #671, opened 2026-04-07
Revisit circuit metrics endpoint


chr-pok:
## Description
The circuit metrics endpoint works for small circuits but returns a lot of information for large circuits (with level_of_detail_nodes >= 1) which causes the endpoint to freeze. This seems to be not related to the speed of the mounted file system for accessing the circuit or computing the metrics from the circuit (which takes < 1s locally), but for returning the huge json body.

**Example:**
Circuit metrics dict of `nbS1-O1` with nodes/edges level of detail 1 has an "me_combo" list with 88'345 strings and a "morphologies" list with 70'881 strings in `property_unique_values` and `property_value_counts` respectively in the json body. This happens because "me_combo" and "morphologies" are defined as enums in the nodes.h5 file (i.e., part of @library) for some reason.

**Proposed solution:**
Reduced metrics returned for large circuits, i.e., introducing a max. length of lists being returned or returning long lists only for levels of detail >= 2.

**Related issue:**
Resource allocation for circuit extraction (https://github.com/openbraininstitute/prod-build-circuit/issues/16) which relies on circuit metrics (only number of nodes/edges per population, but which requires levels of detail 1).

## Acceptance criteria
- [ ] Circuit metrics with nodes/edges level of detail 1 for large circuit returned within reasonable time (order of seconds)

jdcourcol:
both options (threshold and pagination) can be envisioned.

evakennyobi:
Alternative options to be discussed with @MWolfR

chr-pok:
As discussed in today's standup, [PR#679](https://github.com/openbraininstitute/obi-one/pull/679) (which is not directly related to this issue) to be merged first. After that, [PR#673](https://github.com/openbraininstitute/obi-one/pull/673) to be reviewed/discussed with @MWolfR.

MWolfR:
> Reduced metrics returned for large circuits, i.e., introducing a max. length of lists being returned or returning long lists only for levels of detail >= 2.

I propose a mixture of both: There is a max. length for the lists and that value is somewhat increased (but still limited) for level of detail >= 2. E.g.: Maximum length of 25, increased to 100 at higher levels of detail. 

Specific behavior: If a property has more than the indicated number of unique values it is treated as if it was not categorical, i.e., did not use `@library`.

Note: Make sure that this is consistent between `unique_node_property_values_from_population` and `number_of_nodes_per_unique_value_from_population`.

chr-pok:
[...]

PR #673, opened 2026-04-07
Upper limit on the number of unique values in the circuit metrics endpoint

chr-pok:
Added an upper limit on the number of unique values returned by the circuit metrics endpoint. This is important since huge json response bodies will cause the endpoint to freeze.

The upper limit is dependent on the level of details:
- Basic: 25
- Advances: 50
- Full: 100

If a property has more than the indicated number of unique values, it is treated as non-categorical, i.e., not even shown in the dict.

Related issue: https://github.com/openbraininstitute/obi-one/issues/671


MWolfR (approved):
Looks good to me.

PR #675, opened 2026-04-07
initial skeletonization accounting implementation

jdcourcol:
it relates to https://github.com/openbraininstitute/Ultraliser/issues/170#issuecomment-4045432299

bottom line is:
"find the asset, link to it, compute the surface with pylmesh, send that number to accounting service."


obi_one/scientific/tasks/skeletonization/estimate.py
```diff
@@ -0,0 +1,80 @@
...
+        db_client.fetch_file(
+            entity_id=em_cell_mesh.id,
+            entity_type=models.EMCellMesh,
+            asset_id=asset.id,
+            output_path=output_path,
+            strategy=FetchFileStrategy.link_or_download,
+        )
+
+        mesh = pylmesh.load_mesh(str(output_path))
+        return mesh.surface_area()
```
eleftherioszisis: You are making an assumption here that the selected asset is a glb, which may not be the case. There are resources with obj/glb duplicate asset meshes.

If glb is the only format we are going to support I would replace the finding of the asset and the fetching with:

```python
db_client.fetch_assets(
    entity_or_id=entity,
    selection={
        "label": AssetLabel.cell_surface_mesh,
        "content_type": ContentType.model_gltf_binary,
    },
    output_path=output_path,
).one()
```

`one()` will enforce that the result of the filter has exactly one result and will raise for less or more.

obi_one/scientific/tasks/skeletonization/estimate.py
```diff
@@ -0,0 +1,80 @@
...
+    in the configuration.
+
+    Args:
+        db_client: Database client for fetching the config.
+        config_id: UUID of the skeletonization config.
+
+    Returns:
+        The count to use for cost estimation (based on surface area).
+    """
+    config = db_client.get_task_config(
```
eleftherioszisis: entitysdk has no such method.

obi_one/scientific/tasks/skeletonization/estimate.py
```diff
@@ -0,0 +1,80 @@
...
+        config_id=config_id,
+        config_type=SkeletonizationSingleConfig,
```
eleftherioszisis: skeletonization has switched to generic configs. This should be models.TaskConfig

The SkeletonizationSingleConfig can be fetched and deserialized from the config asset. See:
https://github.com/openbraininstitute/obi-one/blob/735ce6a8af1a9bcde31b5fc0f5431fe6132b50e8/obi_one/core/run_tasks.py#L109-L116

obi_one/scientific/tasks/skeletonization/estimate.py
```diff
@@ -0,0 +1,80 @@
...
+        output_path = Path(tmp_dir) / f"{asset.id}.glb"
+        db_client.fetch_file(
+            entity_id=em_cell_mesh.id,
+            entity_type=models.EMCellMesh,
+            asset_id=asset.id,
+            output_path=output_path,
+            strategy=FetchFileStrategy.link_or_download,
+        )
+
+        mesh = pylmesh.load_mesh(str(output_path))
```
eleftherioszisis, line 41: Do we have any idea of the size of the meshes that will have to be loaded in the service's memory to calculate the surface area?

obi_one/scientific/tasks/skeletonization/estimate.py
jdcourcol: done

obi_one/scientific/tasks/skeletonization/estimate.py
jdcourcol: done

obi_one/scientific/tasks/skeletonization/estimate.py
jdcourcol: done

obi_one/scientific/tasks/skeletonization/estimate.py
jdcourcol: ~ 400MB

jdcourcol:
This is what the service returns for a KAUST mesh:

`{"task_type":"morphology_skeletonization","config_id":"d67c1b70-e05c-4e51-bc6b-7bec956c95d4","cost":0.0,"parameters":{"service_subtype":"neuron-mesh-skeletonization","count":5314}}`

obi_one/scientific/tasks/skeletonization/estimate.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+    # Find the task_config asset
+    asset = next(
+        (a for a in task_config.assets if a.label == AssetLabel.task_config),
+        None,
+    )
+    if asset is None:
+        msg = f"No task_config asset found for TaskConfig {config_id}"
+        raise ValueError(msg)
```
eleftherioszisis: there is an existing util that can be used:
```python
from obi_one.utils.db_sdk import get_entity_asset_by_label
```


```suggestion
    # Find the task_config asset
    asset = get_entity_asset_by_label(client=db_client, config=task_config, asset_label=AssetLabel.task_config)
```

obi_one/scientific/tasks/skeletonization/estimate.py
```diff
@@ -0,0 +1,104 @@
...
+    task_config = db_client.get_entity(
+        entity_id=config_id,
+        entity_type=models.TaskConfig,
+    )
+
+    # Find the task_config asset
+    asset = get_entity_asset_by_label(
+        client=db_client, config=task_config, asset_label=AssetLabel.task_config
+    )
+    if asset is None:
```
eleftherioszisis: get_entity_asset_by_label does not allow asset to be None

eleftherioszisis:
Looks good. I'll approve once it's rebased.

jdcourcol:
done :)

eleftherioszisis (approved):

PR #678, opened 2026-04-09
Allow selecting cluster resources based on config resource in circuit simulation

eleftherioszisis:


app/services/resource_estimation/circuit_simulation.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+
+    # get instance types that support the neuron number in ascending order
+    instances = sorted(
+        [
+            info
+            for info in CLUSTER_INSTANCES_INFO[compute_cell]
+            if info.max_neurons <= number_of_neurons
+        ],
+        key=lambda o: o.number_of_neurons,
```
mgeplf: is that right?

eleftherioszisis: Fixed

app/mappings.py
```diff
@@ -105,3 +106,30 @@
...
+            name="small",
+            max_neurons=100,
+            memory_per_instance_gb=1,
+        ),
+        ClusterInstanceInfo(
+            name="large",
+            max_neurons=10000,
+            memory_per_instance_gb=10,
+        ),
+    ],
```
eleftherioszisis, line 134: I need the actual correct numbers here

mgeplf: Something like:
```
CLUSTER_INSTANCES_INFO = {
    "cell_a":
    [
     {"name": "small", "max_neurons": 100, "memory_per_instance_gb": 16},
     {"name": "large", "max_neurons": 0, "memory_per_instance_gb": 384},
    ]

    "cell_b":
    [
     {"name": "small", "max_neurons": 100, "memory_per_instance_gb": 16},
     {"name": "large", "max_neurons": 0, "memory_per_instance_gb": 788},
    ]
}
```
Would be a start; 0 means unlimited, btw.

eleftherioszisis: Could we maybe set a hard upper threshold, e.g. 100_000_000? I am not keen on having no control over how big of a simulation one can launch.

mgeplf: Sure; maybe even something smaller if you prefer.

eleftherioszisis: We agreed to set it to 1_000_000 for now, which would result to allocating around 40 - 80 nodes depending on the vendor. IMO before increasing it we should be careful with a) cost and b) feasibility to allocate that large number of nodes.

jdcourcol: this is fine.

app/services/resource_estimation/circuit_simulation.py
```diff
@@ -0,0 +1,65 @@
+import math
+
+from entitysdk import Client, models
+
+from app.errors import ApiError, ApiErrorCode
+from app.mappings import CLUSTER_INSTANCES_INFO
+from app.schemas.task import ClusterResources, TaskDefinition, TaskLaunchSubmit
+
+CONSTANT = 0.03
```
mgeplf: we should probably say how we got this, something like:
```
This was chosen based off the simulations run and recorded here:
https://openbraininstitute.sharepoint.com/:x:/s/OpenBrainInstitute/IQBgZ53Oe1GhQZQOjc6b_NqlAfobEWCRSznthvhc3X4CHZA?e=ayfe1a

The lowest `Mem(GB) / cell` was ~ 0.025, so a bit of margin was chosen to allow for the variations in connectivity.
```

app/services/resource_estimation/circuit_simulation.py
```diff
@@ -0,0 +1,65 @@
...
+    if compute_cell not in CLUSTER_INSTANCES_INFO:
+        raise ApiError(
+            message=(
+                f"There is no cluster isntance info declared for compute_cell '{compute_cell}'"
+            ),
+            error_code=ApiErrorCode.RESOURCE_ESTIMATION_ERROR,
+        )
+
+    # get instance types that support the neuron number in ascending order
+    instances = sorted(
```
mgeplf, line 41: would it make sense to have them ordered in priority, or have a priority flag?
It would be annoying to get need a `large` instance to run a 50 neuron circuit; since `large` will probably take a long time to get a reservation, but `small` will likely be nearly instantaneous.

app/services/resource_estimation/circuit_simulation.py
```diff
@@ -0,0 +1,65 @@
...
+        entity_id=json_model.config_id,
+        entity_type=models.Simulation,
+    )
+    circuit = db_client.get_entity(
+        entity_id=config.entity_id,
+        entity_type=models.Circuit,
+    )
+    number_of_neurons = circuit.number_neurons
+
+    if compute_cell not in CLUSTER_INSTANCES_INFO:
```
mgeplf: prob. can check this before doing the `config` and `circuit` fetching

app/services/resource_estimation/circuit_simulation.py
GianlucaFicarelli: CONSTANT is quite generic, can we use a name more self-explanatory, or add a comment?

app/services/resource_estimation/circuit_simulation.py
eleftherioszisis: done

app/services/resource_estimation/circuit_simulation.py
eleftherioszisis: Indeed. Changed it into MEM_GB_PER_CELL

app/services/resource_estimation/circuit_simulation.py
eleftherioszisis: It's already ordered from smaller to bigger instance in terms of max_neurons. If we get more instances and the max_neurons is not reliable to sort with we could introduce a priority I suppose.

app/services/resource_estimation/circuit_simulation.py
eleftherioszisis: added comment

mgeplf (approved):

Issue #662, opened 2026-03-31
Migrate circuit metrics endpoint to libsonata/bluepysnap, since it does not work for Allen-V1 (and may fail for other circuits as well)


chr-pok:
## Environment
staging

## Operating system & resolution
Mac 14"

## Description
The circuit metrics endpoint returns `Internal Server Error` when querying metrics from the `Allen-V1` circuit. It could also fail for other circuits with a specific config or .h5 file structure, since these files are currently read and parsed directly using hard-coded keys and assumptions about the file structure.

In order to solve the current error and avoid similar errors in the future, the circuit metrics endpoint should be migrated to use `libsonata` (or `bluepysnap`, whatever is more useful) instead for more robust metric retrieval.

More details:
```
File [/opt/conda/lib/python3.12/site-packages/obi_one/scientific/library/circuit_metrics.py:46](https://jupyterhub.staging.openbrainplatform.com/opt/conda/lib/python3.12/site-packages/obi_one/scientific/library/circuit_metrics.py#line=45), in _get_asset_path(config_path, manifest, temp_dir)
     44 for k, v in manifest.items():
     45     config_path = config_path.replace(k, v)
---> 46 path_to_fetch = Path(config_path).relative_to(temp_dir)
     47 return path_to_fetch

File [/opt/conda/lib/python3.12/pathlib.py:682](https://jupyterhub.staging.openbrainplatform.com/opt/conda/lib/python3.12/pathlib.py#line=681), in PurePath.relative_to(self, other, walk_up, *_deprecated)
    680     break
    681 elif not walk_up:
--> 682     raise ValueError(f"{str(self)!r} is not in the subpath of {str(other)!r}")
    683 elif path.name == '..':
    684     raise ValueError(f"'..' segment in {str(other)!r} cannot be walked")

ValueError: 'node_sets.json' is not in the subpath of '[/tmp/tmpjqor6_54](https://jupyterhub.staging.openbrainplatform.com/tmp/tmpjqor6_54)'
```

## To Reproduce
- Run [circuit_metrics_example.ipynb](https://github.com/openbraininstitute/obi-one/blob/main/examples/G_circuit_metrics/circuit_metrics_example.ipynb) with `Allen-V1` circuit (ID 422a1ddd-8661-42fb-8432-7c879141dddf)

OR
[...]

PR #679, opened 2026-04-09
Fix circuit stats

MWolfR:
Fixes according to https://github.com/openbraininstitute/obi-one/issues/662

### DIFF
--- a/obi_one/scientific/library/circuit_metrics.py
+++ b/obi_one/scientific/library/circuit_metrics.py
@@ -4,6 +4,7 @@ import tempfile
 from collections.abc import Iterator, Mapping
 from enum import IntEnum, StrEnum, auto
 from pathlib import Path
+from os.path import realpath
 
 import h5py
 import libsonata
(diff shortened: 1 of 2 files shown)


obi_one/scientific/library/circuit_metrics.py
```diff
@@ -185,60 +206,29 @@ def names_from_node_sets_file(
...
-    grp = h5["nodes"][population_name]["0"]
-    for k in grp.get("@library", {}):
-        prop_vals = grp["@library"][k][:]
-        prop_idx_counts = pd.Series(grp[k][:]).value_counts()
-        prop_idx_counts = prop_idx_counts.reindex(range(len(prop_vals)), fill_value=0)
-        prop_idx_counts.index = pd.Index(prop_vals)
-        vals_dict[k] = prop_idx_counts.to_dict()
+    for name in pop.enumeration_names:
+        prop_vals = pop.enumeration_values(name)
+        prop_counts = pd.Series(pop.get_attribute(name, pop.select_all())).value_counts()
```
mgeplf: this can probably be done more efficiently with `pop.get_enumeration`; that returns the index into the return `pop.enumeration_values(name)` of the selection

MWolfR: Thanks. I updated the PR. 
As a side note, I like the new implementation not only because of efficiency, but it will also return categories where the count is actually 0. This can happen in extracted circuits, for example.

MWolfR:
Tests are now succeeding. Approval possible @mgeplf ?

mgeplf (approved):
>  Approval possible @mgeplf ?

I'm not very familiar with this repository, so I don't feel like my approval says much, but from the point of switching to libsonata, it all seems to make sense.

PR #682, opened 2026-04-13
Skeletonization add_raw_segmented_spines parameter

james-isbister:
* Add add_raw_segmented_spines parameter


eleftherioszisis:
How is this indicated in the final entitycore resource? Don't we need to do something with it yet?

james-isbister:
Good question. That's one for @marwan-abdellah @MWolfR @jdcourcol

MWolfR:
I do not think this needs to be in the final entitycore resource. It will still be tractable through the `SkeletonizationExecution` associated with the entity, right?

james-isbister:
Yes, it could be read from the obi_one_coordinate.json or the h5 file itself

Ok, so @eleftherioszisis @marwan-abdellah please review this PR as is

eleftherioszisis (approved):

PR #694, opened 2026-04-16
1) Add dynamics_param_names to circuit_metrics and use to optionally expose relative conductanec ou stimuli. 2) change default _spike_location to AIS, 3) Update EntitySDK

james-isbister:
* Add dynamics_param_names for nodes detail >= CircuitStatsLevelOfDetail.basic for biophysical and virtual populations. 
* Use nodes detail CircuitStatsLevelOfDetail.basic in mapped-circuit-metrics endpoint, and check if input_resistance in dynamics_param_names to set SHOW_INPUT_RESISTANCE_BASED_STIMULI
* Bump EntitySDK
* Change default _spike_location to AIS


app/endpoints/circuit_properties.py
```diff
@@ -169,7 +171,8 @@ def mapped_circuit_properties_endpoint(
             circuit = db_client.get_entity(entity_id=circuit_id, entity_type=Circuit)
             simulation_options_usability = {
                 CircuitUsability.SHOW_ELECTRIC_FIELD_STIMULI: circuit.scale == "microcircuit",
-                CircuitUsability.SHOW_INPUT_RESISTANCE_BASED_STIMULI: False,
+                CircuitUsability.SHOW_INPUT_RESISTANCE_BASED_STIMULI: "input_resistance"
+                in circuit_metrics.biophysical_node_populations[0].dynamics_param_names,
```
MWolfR: Simply the first biophysical one? Well I guess this is not the only place where we assume there's only a single biophysical population. But eventually we need to find something better.

app/endpoints/circuit_properties.py
```diff
@@ -169,7 +171,8 @@ def mapped_circuit_properties_endpoint(
             circuit = db_client.get_entity(entity_id=circuit_id, entity_type=Circuit)
             simulation_options_usability = {
                 CircuitUsability.SHOW_ELECTRIC_FIELD_STIMULI: circuit.scale == "microcircuit",
```
MWolfR: Compare to entitysdk.types.CircuitScale.microcircuit instead.

app/endpoints/circuit_properties.py
```diff
@@ -169,7 +171,8 @@ def mapped_circuit_properties_endpoint(
...
-                CircuitUsability.SHOW_INPUT_RESISTANCE_BASED_STIMULI: False,
+                CircuitUsability.SHOW_INPUT_RESISTANCE_BASED_STIMULI: "input_resistance"
```
MWolfR: Maybe assign "input_resistance" string to a variable at the top of the file and compare to that. Makes the presence of magic strings more transparent.

obi_one/scientific/library/circuit_metrics.py
```diff
@@ -540,6 +549,7 @@ def get_circuit_metrics(  # noqa: PLR0914
                 property_names=node_props[nodepop]["property_list"],
                 property_unique_values=node_props[nodepop]["property_unique_values"],
                 property_value_counts=node_props[nodepop]["property_value_counts"],
+                dynamics_param_names=node_props[nodepop]["dynamics_param_names"],
```
MWolfR, line 552: Looks to me like "dynamics_param_names" is not guaranteed to exist in the dict. For non-biophysical node populations. 
Correct me if I am wrong, but I think the following is required:

Either add that strong to `lst_req_props` (line 324) or access it here with .get().

In the first case, for a non-biophysical population we will get an empty list. In the second case, for a non-biophysical population we will get `None`. Whichever makes more sense to you.

And make sure that it is tested with non-biophysical populations.

MWolfR (changes requested):
See individual comments.

obi_one/scientific/library/circuit_metrics.py
james-isbister: The for loop is only for biophysical populations. I didn't add it to lst_req_props because that code is also for virtual populations

MWolfR: That makes sense.
Just one last question: What would happen if there is - for some reason - no dynamics parameter in a biophysical population? in line 217:
`return list(pop.dynamics_attribute_names)`

this assumes that in that case `libsonata` returns an empty list. Can you just quickly check and confirm that this is the case? That it is not returning `None` instead?

james-isbister: Both when dynamics_params is either deleted or empty it returns an empty set which, list(pop.dynamics_attribute_names) would cast as an empty list:
Screenshot 2026-04-17 at 11 29 02

MWolfR (approved):

PR #704, opened 2026-04-20
Fixed circuit metrics call for resource estimation

chr-pok:
Small type fix required after changes in [PR#679](https://github.com/openbraininstitute/obi-one/pull/679) in order to make resource estimation for circuit extraction work again.


james-isbister (approved):
LGTM!

PR #707, opened 2026-04-20
Resolve circuit_simulation by target simulator and circuit scale

eleftherioszisis:
## Summary

This PR introduces deterministic routing from generic `circuit_simulation` requests to concrete simulation task types, based on simulator and circuit scale metadata from simulation config and circuit.
* `circuit_simulation` is now treated as a group task and resolved before launch/estimate.
* Resolution is done in `app/services/task.py::select_simulation_task`.
* Endpoints `POST /declared/task/launch` and `POST /declared/task/estimate` now map the task type and proceed with the mapped concrete task.

## How task selection works
`select_simulation_task` determines target_simulator in this order:

* Fetch simulation entity by config_id
* Load simulation config asset with:
  * label = sonata_simulation_config
  * content_type = application_json
* Parse with `libsonata.SimulationConfig`
* If `simulation_config.target_simulator exists`, use it
* Else fallback to `circuit.target_simulator` from the linked circuit (simulation.entity_id)

Resolved `target_simulator` is then used as follows:
* `LearningEngine` -> `TaskType.circuit_simulation_inait_machine`
* `NEURON` or `CORENEURON`:
  * `circuit.scale` in {single, pair, small} -> `TaskType.circuit_simulation_neuron`
  * otherwise -> `TaskType.circuit_simulation_neurodamus_cluster`
* any other simulator -> raises RuntimeError("Unsupported target simulator ...")

## Endpoint Behavior Changes
In `app/endpoints/task.py`:
* For both `/launch` and `/estimate`, when input `task_type == circuit_simulation`:
    * call select_simulation_task(...)
    * replace incoming model task type with concrete task type
After mapping, existing flow continues using `TASK_DEFINITIONS[resolved_task_type]`.

## Launch Payload / Executor Impact
In `submit_task_job` (`app/services/task.py`), concrete type controls payload and executor:

* circuit_simulation_inait_machine
  * payload: _inait_job_data
  * executor: single_node_job
* circuit_simulation_neurodamus_cluster
  * payload: _circuit_simulation_job_data
  * executor: distributed_job
* other tasks
  * payload: _generic_job_data
  * executor: single_node_job

Issue: https://github.com/openbraininstitute/prod-circuit-simulation/issues/97

### DIFF
--- a/app/services/task.py
+++ b/app/services/task.py
@@ -19,7 +19,8 @@ from app.schemas.task import (
     TaskLaunchInfo,
     TaskLaunchSubmit,
 )
-from app.types import CallBackAction, CallBackEvent, TargetSimulator, TaskType
+from app.types import CallBackAction, CallBackEvent, TaskType
+from obi_one.types import TargetSimulator
 from obi_one.utils import db_sdk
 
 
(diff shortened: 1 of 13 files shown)


app/services/task.py
```diff
@@ -251,7 +253,21 @@ def estimate_task_resources(
                 task_definition=task_definition,
                 compute_cell=compute_cell,
             )
-        case TaskType.circuit_simulation:
+        case TaskType.circuit_simulation_neuron:
```
mgeplf: for all the ones that are the same, can they use `|` so it's clear they all make the same call?

app/services/task.py
```diff
@@ -64,7 +72,17 @@ def submit_task_job(
     all_callbacks = [failure_callback, *callbacks]
 
     match task_definition.task_type:
-        case TaskType.circuit_simulation:
+        case TaskType.circuit_simulation_inait_machine:
+            executor_type = ExecutorType.single_node_job
```
jdcourcol, line 76: why do we need this executor_type  ? I see we put that in the activity but it doesn't seem something we pass to the launch system API. So it looks like "implicit".

eleftherioszisis: It is indeed only used to declare the executor type to the activity registered for this job.

jdcourcol (approved):
I am good with that.

app/services/task.py
jdcourcol: I think we should change that or remove it because that is strange as it is not coming from launch-system.
but that is not the topic of this PR

PR #711, opened 2026-04-21
Replace pylmesh with pyvista

### LINKED ISSUE #710: pylmesh vs pyvista
## Description
A dependency to `pylmesh` was recently added in #675 . While I didn't follow the exact requirements for that, I realized the following:
1.  `pylmesh` seems kind of a heavy package to install
2. Plus, on mac, default installation fails because it requires to `brew install libomp`
3. It seems like pylmesh is calculating the mesh surface area, which is something that the already existing dependency `pyvista` can do as well.

@jdcourcol : See the linked PR ( #711 ) and feel free to approve it or discard it.

## Acceptance criteria
This is more a question than a request, so no acceptance criteria.

Time estimate: 0.5 days

CC: @james-isbister , @marwan-abdellah

jplanasc:
Fixes #710


marwan-abdellah:
@jplanasc This is not correct. pyvista has its own limitations. It cannot read any compressed GLB meshes (with Draco compression) which will be mainly supported in EC!

jplanasc:
The mesh computation is memory-critical, pyvista or trimesh would not be a good fit.

PR #712, opened 2026-04-22
Enable type checker

GianlucaFicarelli:
To establish a baseline , ignore existing errors with specific directives generated with:
```
uv run --with ty ty check app obi_one --add-ignore
```

See https://pydevtools.com/handbook/how-to/how-to-migrate-from-mypy-to-ty/#automated-error-suppression-with-add-ignore

To ignore this commit in git blame, the file `.git-blame-ignore-revs` should be updated with the final commit hash after merging this PR.
This is supported also in GitHub, see https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# .git-blame-ignore-revs
+# xxxxxxx # ignore type errors for baseline with ty
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ format:  ## Run formatters
 lint:  ## Run linters
 	uv run -m ruff format --check
 	uv run -m ruff check
-	#uv run -m pyright obi_one
+	uv run --with ty ty check app obi_one
 
 format_count: ## Count the number of errors by file
 	uv run -m ruff check --output-format=json | jq '.[].filename' | sort | uniq -c
--- a/app/application.py
+++ b/app/application.py
@@ -120,7 +120,7 @@ app = FastAPI(
         ApiError: api_error_handler,
         RequestValidationError: validation_exception_handler,
         EntitySDKError: entity_sdk_error_handler,
-    },
+    },  # ty:ignore[invalid-argument-type]
     root_path=settings.ROOT_PATH,
     strict_content_type=False,
 )
--- a/app/config.py
+++ b/app/config.py
@@ -68,7 +68,7 @@ class Settings(BaseSettings):
 
     def build_launch_system_url(self, subdomain: str) -> str:
         """Return the launch-system URL with the subdomain placeholder resolved."""
-        return self.LAUNCH_SYSTEM_URL_TEMPLATE.replace(self.SUBDOMAIN_PLACEHOLDER, subdomain)
+        return self.LAUNCH_SYSTEM_URL_TEMPLATE.replace(self.SUBDOMAIN_PLACEHOLDER, subdomain)  # ty:ignore[unresolved-attribute]
 
     # Path to the obi-one repository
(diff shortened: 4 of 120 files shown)


GianlucaFicarelli:
The 2 [errors in the CI](https://github.com/openbraininstitute/obi-one/actions/runs/24777919119/job/72500927753?pr=712) are unrelated to this PR, and caused by the new entitysdk version that fails if the cell morphology protocol is missing.
The code and tests could probably be simplified because of that, but in a separate PR.

```
FAILED tests/obi_one/scientific/tasks/skeletonization/test_skeletonization.py::test_register_output_resource_creates_protocol_when_missing - pydantic_core._pydantic_core.ValidationError: 1 validation error for CellMorphology
cell_morphology_protocol
  Field required [type=missing, input_value={'created_by': {'id': Non...4f51-8f37-17bdc575ad37'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
FAILED tests/obi_one/scientific/tasks/skeletonization/test_skeletonization.py::test_register_output_resource_reuses_existing_protocol - pydantic_core._pydantic_core.ValidationError: 1 validation error for CellMorphology
cell_morphology_protocol
  Field required [type=missing, input_value={'created_by': {'id': Non...46ab-8592-bcf7ff0cf4ed'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
====== 2 failed, 623 passed, 410 warnings, 2 errors in 119.05s (0:01:59) =======
```

GianlucaFicarelli:
> The 2 [errors in the CI](https://github.com/openbraininstitute/obi-one/actions/runs/24777919119/job/72500927753?pr=712) are unrelated to this PR, and caused by the new entitysdk version that fails if the cell morphology protocol is missing. The code and tests could probably be simplified because of that, but in a separate PR.
> 

PR: https://github.com/openbraininstitute/obi-one/pull/713

GianlucaFicarelli:
This is ready for discussion and review. The insufficient coverage message can be ignored since it's not caused by this PR.

eleftherioszisis (dismissed):
Awesome!

james-isbister (commented):
Thanks for this. Just wondering can we check that the command has been run in the CI? Otherwise people might add new stuff without the ignore markers added

james-isbister:
Ah wait... I guess the CI will fail if they don't add ignore markers or fix new code? Through one of the updated make commands? That's the point 😆

GianlucaFicarelli:
> Ah wait... I guess the CI will fail if they don't add ignore markers or fix new code? Through one of the updated make commands? That's the point 😆

Yes the checker is now run in the CI, but in the new code the ignore markers should not be added unless there isn't another way to fix the error (for example: it depends on an external library, or it's a detection error of the linter).
In this way the quality of the code and type annotations should always get better from that baseline.
We should also make a ticket for addressing gradually the pre-existing type errors.

eleftherioszisis (dismissed):

james-isbister:
Now the webinar is complete, I think we can re-accept and merge straightaway if you could re-run for the current contents of main.

GianlucaFicarelli:
@james-isbister @eleftherioszisis are we good with these changes?
It would be nice to merge before other commits cause conflicts, but be aware that other pending PRs may need to fix type errors before merging.

I'll open separate PRs for:
- ignore the final commit in git blame
- remove the upper bound in ruff version (there shouldn't be an upper bound, unless temporary)

james-isbister (approved):
LGTM!

PR #714, opened 2026-04-22
Sets the number of connections properly.

MWolfR:
The number of connections is meant to be like the number of synapses, but duplicates from the same source neuron onto the same target neuron are counted only once.

I generated that number by considering the length of the DataFrame of pre-post pairs with `.drop_duplicates()`.


jplanasc (approved):
LGTM.

PR #717, opened 2026-04-23
Replace pvista with pylmesh

### LINKED ISSUE #718: Replace pyvista with trimesh / pylmesh
## Description
In the context of reducing the installation time of `obi-one`, we should try to remove `pyvista` completely and use alternatives, like `trimesh` or `pylmesh` (which are already dependencies). `pyvista` brings `vtk` with it, which is a considerably heavy dependency.

## Acceptance criteria
- [ ] pyvista is not a dependency of obi-one

Time estimate: 3 days

dkeller9:
This replaces pyvista with pylmesh for the mesh validation.

Fixes #718


jplanasc (approved):
LGTM

PR #719, opened 2026-04-24
Calculate metrics from h5

dkeller9:
Change the "register-morphology-with-calculated-metrics" endpoint to run the morphometrics on h5 output.


ilkilic (approved):
I just needed this, thanks a lot!

PR #721, opened 2026-04-24
new docs component for multiple entities

AurelienJaquier:


AurelienJaquier:
Hi @james-isbister. I am continuing the implementation of the multiple entities with this PR defining the docs/components. Can I ask you to check this PR?
I believe there were some discussion to not use a NamedTuple here so I just put a tuple. Would that be ok?

docs/gui-definition-spec/components/multiple_entities/multiple_entities.md
```diff
@@ -0,0 +1,28 @@
+## Model identifier multiple UI element
+
+ui_element: `UIElement.MODEL_IDENTIFIER_MULTIPLE`
+
+Reference schema [MODEL_IDENTIFIER_MULTIPLE](reference_schemas/model_identifier_multiple.json)
+
+
+### UI design
+
+#### In initialization block
```
james-isbister: The entity selection could be in any Block

docs/gui-definition-spec/components/multiple_entities/multiple_entities.md
```diff
@@ -0,0 +1,28 @@
...
+
+<img src="designs/multiple_entities.png"  width="300" />
+
+In initialization block, user can select multiple entities in an entity group by clicking on the `Add <entity type>` button. Those entities are stored in a tuple.
+
+User can also scan over groups of entities. They can add a new entity group by clicking on `Add group`. The groups are stored within a list.
+
+By default, when the user clicks on the workflow, they have to first select an entity. When going on initilization, they have this entity being in a group. They can then selects more entities or add groups.
```
james-isbister: If the case requires a set of entities then I think they can select multiple entities for a single group in the first step (i.e. before they get to the configuration)?

james-isbister (commented):
Nice. 

1. For the NamedTuple let's also get @chr-pok 's opinion. We discussed it briefly last week. The advantage would be that the user (or a private collaborator / user viewing the public workflow) can remember what each group of entities refers to. It could also be used for example in the EMSynapseMapping case for naming the output entity created for each group (i.e. circuit). For these reasons, I would lean towards keeping it as a NamedTuple.

2. Also perhaps we can document and unify the designs for the 4 similar cases we now have, each which have could be represented by a different UIElement:
A. Single entity (i.e. circuit simulation)
B. Scan over single entities (useful already for scanning over multiple EMCellMesh entities in skeletonization).
C. Multi entities no scan (not sure we have a use case atm, but this would just hide the "Add group" option).
D. Multi entities scan (EMSynapseMapping)

3. Also, did we agree to view the mini-detail view of each entity on the right hand side, when it's clicked in the central column? There was discussion of a way of toggling between the mini details view and a visualisation of same a morphology

chr-pok:
> 1. For the NamedTuple let's also get @chr-pok 's opinion. We discussed it briefly last week. The advantage would be that the user (or a private collaborator / user viewing the public workflow) can remember what each group of entities refers to. It could also be used for example in the EMSynapseMapping case for naming the output entity created for each group (i.e. circuit). For these reasons, I would lean towards keeping it as a NamedTuple.

Indeed, having the possibility to give names to certain groups of entities would be very useful I think. Could use a default name if no specific name is provided. But should make sure that no duplicate names can be used though, if possible.

Just thinking, maybe it could even be extended to have a name + description (optional), since in some cases it could be even more useful to add not only a name but also a description to a specific group of entities containing information about why/how they have been selected etc.

Unrelated to that, I think the `gui-definition.md` file should be updated as well, containing a link to the new multiple entities design.

james-isbister:
A description is a nice idea but perhaps we can introduce that at a later date, as we're already doing a lot in the front and backend in this work?

Ok, so I think the NamedTuple would only be used in the last case D, if you agree @AurelienJaquier ?

AurelienJaquier:
>     1. For the NamedTuple let's also get @chr-pok 's opinion. We discussed it briefly last week. The advantage would be that the user (or a private collaborator / user viewing the public workflow) can remember what each group of entities refers to. It could also be used for example in the EMSynapseMapping case for naming the output entity created for each group (i.e. circuit). For these reasons, I would lean towards keeping it as a NamedTuple.

Actually I onder if namedtuples are the right structure for our use case here. I see that we are supposed to give field names for each entity (see [namedtuple's doc](https://docs.python.org/3/library/collections.html#collections.namedtuple)).
I think what we want is closer to a dict of tuples. A dict of tuples would have the names of the groups as the keys, and the entities as tuples inside.

Indeed that would only be for case D. B is a list of entities and C is a tuple of entities.

The description is a good idea, but I agree with James: we can postpone it to later.

>     2. Also perhaps we can document and unify the designs for the 4 similar cases we now have, each which have could be represented by a different UIElement:
>        A. Single entity (i.e. circuit simulation)
>        B. Scan over single entities (useful already for scanning over multiple EMCellMesh entities in skeletonization).
>        C. Multi entities no scan (not sure we have a use case atm, but this would just hide the "Add group" option).
>        D. Multi entities scan (EMSynapseMapping)

Ok. I can modify model_identifier for the single entity case, and add the other cases to my multiple_entities docs.

>     3. Also, did we agree to view the mini-detail view of each entity on the right hand side, when it's clicked in the central column? There was discussion of a way of toggling between the mini details view and a visualisation of same a morphology

For these I believe we said that we would stick to whatever display we currently have, and that we would implement the mini details+visualisation switch later.

AurelienJaquier:
@james-isbister I am thinking about the difference in design for cases B and C.
C would be the same but without the 'Add group' button and without the group name.
For B we would remove all of that plus the group rectangle, leaving only the list of neurons and the 'Add neuron' button.

Does that sound good?

james-isbister:
> C would be the same but without the 'Add group' button and without the group name.
> For B we would remove all of that plus the group rectangle, leaving only the list of neurons and the 'Add neuron' button.

This sounds good to me. We can see if it comes across clear that we are scanning in B. Maybe there is a hoverover on the "Add neuron" button or something like that

> I think what we want is closer to a dict of tuples. A dict of tuples would have the names of the groups as the keys, and the entities as tuples inside.

Hm, I think you're right that dictionaries would be the more natural representation. There are three issues/questions I see with it (maybe why we went with the named tuples to begin with):
1. We'd have to update the parameter scan logic to detect dictionaries as a scanning case. 
2. Maybe we don't always want this to be the case (dictionaries might mean something else in another context).
3. When the dictionary is detected, the generated SingleConfig(s) would need to hold a dictionary with a single element. We'd also need to enforce that there is only one tuple in the dictionary through the SingleCoordMixin.

Because of these, I'd lean towards the NamedTuple case? Maybe it reduces complexity in the scan logic for now. Any thoughts @chr-pok ?

AurelienJaquier:
One solution could be to just create our own custom class:

```
class EntitiesSet:
    def __init__(self, name: str, values: tuple):
        self.name = name
        self.values = values

entities = EntitiesSet(
  "neuron group 1",
  (
    MorphologyFromID(id1),
    MorphologyFromID(id2),
    MorphologyFromID(id3)
  )
)
```

AurelienJaquier:
@james-isbister @chr-pok I added the NamedTuple subclasses and a new SchemaKey as we discussed yesterday. Do you agree with my latest changes?

docs/gui-definition-spec/components/multiple_entities/multiple_entities.md
```diff
@@ -0,0 +1,100 @@
...
+    json_schema_extra={
+        SchemaKey.UI_ELEMENT: UIElement.MODEL_IDENTIFIER_MULTIPLE,
+    },
+)
+```
+
+## case D
+
+```py
+neurons: list[CellMorphologyAndMEModelNamedTuple, ...] = Field(
```
james-isbister: To capture the cast to SingleConfig, would the type be: 
`CellMorphologyAndMEModelNamedTuple | list[CellMorphologyAndMEModelNamedTuple]`

Also is the '...' necessary?

james-isbister: Could we also show how `CellMorphologyAndMEModelNamedTuple` is defined?

docs/gui-definition-spec/components/multiple_entities/multiple_entities.md
```diff
@@ -0,0 +1,100 @@
...
+    title="Neurons",
+    description="Neurons to include in the circuit (>= 1).",
+    min_length=1,
+    json_schema_extra={
+        SchemaKey.UI_ELEMENT: UIElement.MODEL_IDENTIFIER_MULTIPLE,
+        SchemaKey.ACCEPTED_TYPES: [CellMorphologyFromID, MEModelFromID],
```
james-isbister: Looks good. What does SchemaKey.ACCEPTED_TYPES show in the openapi.json for this? I wonder if they should be strings or classes.

obi_one/scientific/from_id/named_tuple_from_id.py
```diff
@@ -0,0 +1,17 @@
...
+class CellMorphologyNamedTuple(NamedTuple):
+    elements: tuple[CellMorphologyFromID, ...] = Field(min_length=1)
+
+
+class MEModelNamedTuple(NamedTuple):
+    elements: tuple[MEModelFromID, ...] = Field(min_length=1)
+
+
+class CellMorphologyAndMEModelNamedTuple(NamedTuple):
+    elements: tuple[CellMorphologyFromID | MEModelFromID, ...] = Field(min_length=1)
```
james-isbister, line 9: Is the '...' required?

docs/gui-definition-spec/components/multiple_entities/multiple_entities.md
```diff
@@ -0,0 +1,100 @@
...
+
++ buttons next to the entities that are not yet in the group let user add the entity to the group. - button next to the entities that are already in the group let the user remove an entity from the group.
+
+
+### Config examples
+
+## case B:
+
+```py
+cell_mesh: list[EMCellMeshFromID] = Field(
```
james-isbister: I think it should be:
`EMCellMeshFromID | list[EMCellMeshFromID]`

docs/gui-definition-spec/components/multiple_entities/designs/multiple_entities_case_B.png
james-isbister, line 1: Maybe "Add neuron to scan +"? What do you think?

docs/gui-definition-spec/components/multiple_entities/designs/multiple_entities_case_D.png
james-isbister, line 1: Maybe "Add group to scan +"?

james-isbister (commented):
Looking good

docs/gui-definition-spec/components/multiple_entities/multiple_entities.md
AurelienJaquier: Yes, I'll change it in my next commit.
Good catch for the '...'. Those are needed for tuple of variable length, but not needed for lists.

docs/gui-definition-spec/components/multiple_entities/multiple_entities.md
AurelienJaquier: Ah, good question. How do I get the openapi.json?
If needed, I can create a StrEnum to be used in SchemaKey.ACCEPTED_TYPES

obi_one/scientific/from_id/named_tuple_from_id.py
AurelienJaquier: For the tuple, yes, it is required to indicate that the tuple can have variable length.

docs/gui-definition-spec/components/multiple_entities/designs/multiple_entities_case_B.png
AurelienJaquier: Sounds good, I'll change it

AurelienJaquier:
Ok, I think this is ready to be reviewed by front-end engineers.
Can I ask you to review this PR @bilalesi ? This is about a new UI element for selecting multiple entities to be used first in the EM synapse mapping task.

I will be away until Wednesday, but @james-isbister should be able to answer your questions if you have any in the meantime.

AurelienJaquier:
@bilalesi could you review this when you have some time?

docs/gui-definition-spec/components/multiple_entities/designs/multiple_entities_case_B.png
bilalesi: what's the difference between CASE_B and CASE_C ?
there is "Add neuron to scan +" and "Add neuron" ?

docs/gui-definition-spec/components/multiple_entities/designs/multiple_entities_case_D.png
bilalesi: what's the group use case ?
and how in the UI we can distinguish between group and non group using the schema ?

docs/gui-definition-spec/components/multiple_entities/designs/multiple_entities_case_B.png
AurelienJaquier: In case B, we scan over single entities, meaning we will produce multiple single configs containing a single entity that will each run a job with a single entity.
In case C, we do not do any scan, but the single config contains multiple entities. So we will have one job that uses multiple entities as inputs.

This is why we have 'add neuron to scan' in case B, but just 'add neuron' in case C. Does it make sense?

docs/gui-definition-spec/components/multiple_entities/designs/multiple_entities_case_D.png
AurelienJaquier: So in case D, we can scan over groups of entities, meaning that we will produce multiple single configs (with the scan), that each use multiple entities.
For this case, the schema has a list of NamedTuple. For non-group case (so case B, scan over single entities), we have a list of entities in the schema.

bilalesi (approved):

PR #723, opened 2026-04-24
Add inter-spike-interval distribution spike stimulus

ilkilic:
Adds a mechanism for generating spike trains by sampling inter-spike intervals from probability distributions.

### What’s included

- Distribution blocks:
  - constant
  - exponential
  - gamma
  - uniform
- `    InterSpikeIntervalDistributionSpikeStimulus` for spike generation based on inter-spike interval (ISI) sampling
- Integration with existing stimulus and union systems
- Tests for distributions and spike stimulus behavior
- Example notebook demonstrating usage

### Notes
- This PR focuses on ISI-based spike generation (i.e. sampling inter-spike intervals and accumulating spike times)
- A complementary approach based on directly sampling spike times from a distribution will be added in a follow-up PR
- Designed to be extensible for additional distributions
- Integrates with existing stimulus infrastructure

### Acknowledgements

Parts of this work build on changes originally developed by @james-isbister in the `syn_param_task` branch (including initial distribution block implementations and related infrastructure) .


obi_one/scientific/blocks/stimuli/spike/distribution.py
```diff
@@ -0,0 +1,137 @@
...
+                msg = "Inter-spike intervals must be positive."
+                raise ValueError(msg)
+            t += interval
+            if t >= duration:
+                break
+            spikes.append(t)
+        return spikes
+
+    @staticmethod
+    def _validate_stimulus_windows(
```
james-isbister: Not sure we need this. We have removed this requirement in the higher classes, to give the user flexibility and because some edge cases were tricky to catch iirc

obi_one/scientific/blocks/distributions/gamma.py
```diff
@@ -0,0 +1,47 @@
...
+    random_seed: int | list[int] = Field(
+        default=1,
+        title="Random Seed",
+        description="Seed for drawing random values from the gamma distribution.",
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.INT_PARAMETER_SWEEP,
+        },
+    )
+
+    def _sample_generator(self, n: int = 1, rng: np.random.Generator | None = None) -> list[float]:
```
james-isbister: I think this will return list[PositiveFloat]

obi_one/scientific/blocks/distributions/exponential.py
```diff
@@ -0,0 +1,39 @@
...
+
+import numpy as np
+from pydantic import Field, PositiveFloat
+
+from obi_one.core.schema import SchemaKey, UIElement
+from obi_one.core.units import Units
+from obi_one.scientific.blocks.distributions.base import Distribution
+
+
+class ExponentialDistribution(Distribution):
```
james-isbister: I think we might want a NegativeExponentialDistribution rather than a positive one for an inter-spike interval distribution?

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
```diff
@@ -0,0 +1,137 @@
...
+    def _generate_spike_train_from_distribution(
+        distribution: Distribution,
+        duration: float,
+        rng: np.random.Generator | None = None,
+    ) -> list[float]:
+        spikes: list[float] = []
+        t = 0.0
+        while True:
+            interval = distribution.sample(1, rng=rng)[0]
+            if interval <= 0.0:
```
james-isbister, line 77: I need to think a bit more about how to deal with the possible negative case (i.e. possible with uniform and constant distributions atm).

obi_one/scientific/blocks/distributions/exponential.py
```diff
@@ -0,0 +1,39 @@
...
+    random_seed: int | list[int] = Field(
+        default=1,
+        title="Random Seed",
+        description="Seed for drawing random values from the exponential distribution.",
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.INT_PARAMETER_SWEEP,
+        },
+    )
+
+    def _sample_generator(self, n: int = 1, rng: np.random.Generator | None = None) -> list[float]:
```
james-isbister: I think this will return list[PositiveFloat]

james-isbister (commented):
Looks good! I'd like to think through the logic once or twice more but I think it's good - maybe we could handle the zero case differently, I'm not sure

Interestingly, it didn't occur to me that there are two forms of distribution based stimuli, which are both probably equally important:
1. Sampling inter-spike-intervals
2. Sampling spike times from a probability distribution

I realise we never specified which one - I was thinking of the latter! Still I think it's great to have both, so maybe we can keep this PR about the first case, and then implement the second case in a separate PR?

As such, could we rename the DistributionSpikeStimulus to InterspikeIntervalDistributionSpikeStimulus or something like that, and update the stimulus tests appropriately?

ilkilic:
Thanks @james-isbister ! I have addressed you comments. For the zero/negative case I kept the runtime check for now, happy to revisit once we decide how we want to handle this more generally. And yes, good point about the two types. This PR is really the inter-spike-interval one, I renamed it to InterSpikeIntervalDistributionSpikeStimulus and updated the tests and notebooks. I will add the spike-time version in a follow-up PR.

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
james-isbister: Thought about it again, and I think raising an error like this is a good way of deal with values drawn less than equal 0

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
```diff
@@ -0,0 +1,114 @@
...
+
+    distribution: AllDistributionsReference | None = Field(
+        default=None,
+        title="Interval Distribution",
+        description="Distribution used to sample inter-spike intervals in milliseconds.",
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.REFERENCE,
+            SchemaKey.REFERENCE_TYPE: AllDistributionsReference.__name__,
+        },
+    )
```
james-isbister, line 52: I wonder if we want to add a refractory period here or add an offset (which defaults to 0) to the exponential and gamma distributions?

james-isbister (commented):
Looks good, thanks! Just one more question about refractory periods

examples/F_single_cell_simulations/entitysdk_memodel_with_synapses_simulation.ipynb
james-isbister, line 1: Could we add some rasters of input spikes for the different ISI stimuli, and histogram plots of the ISIs as a visual validation?

james-isbister (commented):
And a comment about viewing the spikes. Then I think we're good to go!

ilkilic:
Good points! I have included both in my last commit.

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
```diff
@@ -0,0 +1,127 @@
...
+                msg = "Inter-spike intervals must be positive."
+                raise ValueError(msg)
+            t += interval + refractory_period
+            if t >= duration:
+                break
+            spikes.append(t)
+        return spikes
+
+    def generate_spikes_by_gid(self, source_gids: list[int]) -> dict[int, list[float]]:
+        if self.distribution is None:
```
james-isbister, line 87: Maybe here we could just set self.distribution= uniform distribution [0, 1]  in this case, as for ui schema reasons we need to support the none case.

Then we can also specify a string here for the distribution reference case, which will appear for the default case.

Perhaps "Default: Uniform [0, 1]"

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
```diff
@@ -0,0 +1,127 @@
...
+        default=None,
+        title="Interval Distribution",
+        description="Distribution used to sample inter-spike intervals in milliseconds.",
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.REFERENCE,
+            SchemaKey.REFERENCE_TYPE: AllDistributionsReference.__name__,
+        },
+    )
+
+    refractory_period: NonNegativeFloat | list[NonNegativeFloat] = Field(
```
james-isbister: Sorry! If we could move this to the distribution for gamma and exponential as a shift parameter, i think it's a useful addition to those distributions :)

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
james-isbister: The string can be added here:

https://github.com/openbraininstitute/obi-one/blob/cf9e45d9342f66c50927024ae447455455e9f14f/obi_one/scientific/tasks/generate_simulations/config/circuit.py#L60

And:
https://github.com/openbraininstitute/obi-one/blob/cf9e45d9342f66c50927024ae447455455e9f14f/obi_one/scientific/tasks/generate_simulations/config/circuit.py#L60

ilkilic: Makes sense for the UI side. I am just wondering if having a default might surprise users. Maybe better to have the user pick one explicitly?

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
ilkilic: Ah interesting. I was thinking of refractory_period more as part of the spike generation rather than the distribution itself, so I put it on the ISI stimulus. It also keeps it consistent across different distributions. But happy to move it if you feel strongly about it.

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
james-isbister: I kind of agreed with Gil and Bilal that we'd always support a default case for these references, so for now it would be simpler! We could always change it later to a more interesting default also, maybe uniform [0, 50] is better, so we don't produce too many spikes

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
james-isbister: Yes, I thought that too, but then I thought that its maybe confusing to use say a uniform or constant distribution for the ISI, and then on top of that add an additional offset. The uniform and constant distributions which the user provides would naturally include the refractory period in their mind

I also think its nice that the distributions also offer this shift, so I'd lean towards changing it

obi_one/scientific/blocks/stimuli/spike/isi_distribution.py
ilkilic: Makes sense. I was wondering if an exponential default might be more natural here, for example scale=50 (~20 Hz)?. But happy to go with uniform if we prefer to keep it simpler.

james-isbister: Yes, sounds good!

ilkilic:
Done! I added the default exponential distribution (scale=50) when none is provided and moved the refractory handling to the distributions as a shift parameter.

obi_one/scientific/blocks/distributions/exponential.py
```diff
@@ -0,0 +1,50 @@
...
+    scale: PositiveFloat | list[PositiveFloat] = Field(
+        default=1.0,
+        title="Scale",
+        description="Scale parameter of the exponential distribution in milliseconds.",
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.FLOAT_PARAMETER_SWEEP,
+            SchemaKey.UNITS: Units.MILLISECONDS,
+        },
+    )
+    shift: NonNegativeFloat | list[NonNegativeFloat] = Field(
```
james-isbister: I think these can be standard floats for flexibility?

obi_one/scientific/tasks/generate_simulations/config/base.py
```diff
@@ -48,6 +48,7 @@
 
 DEFAULT_NODE_SET_NAME = "Default: All Biophysical Neurons"
 DEFAULT_TIMESTAMPS_NAME = "Default: Simulation Start (0 ms)"
+DEFAULT_DISTRIBUTION_NAME = "Default: Exponential ISI, scale 50 ms (~20 Hz)"
```
james-isbister: Hm, we might have to rethink this when we come to the other types of distribution stimulus, but let's leave that to that future PR!

For now this could be slightly shorter to appear in the dropdown, like: "Default: Exp, scale 50 ms, 20 Hz"

james-isbister (commented):
A final two tiny comments, and then we can merge, I promise :D

ilkilic:
Done! :D

obi_one/scientific/blocks/distributions/gamma.py
```diff
@@ -0,0 +1,58 @@
...
+    scale: PositiveFloat | list[PositiveFloat] = Field(
+        default=1.0,
+        title="Scale",
+        description="Scale parameter of the gamma distribution in milliseconds.",
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.FLOAT_PARAMETER_SWEEP,
+            SchemaKey.UNITS: Units.MILLISECONDS,
+        },
+    )
+    shift: NonNegativeFloat | list[NonNegativeFloat] = Field(
```
james-isbister: And here, sorry :)

ilkilic: ah sorry! updated in my last commit :)

james-isbister (approved):
Great, thank you @ilkilic !

pyproject.toml
```diff
@@ -33,7 +33,7 @@ dependencies = [
     # WARNING: `make upgrade-deps` upgrades all dependencies (including optional) and may introduce breaking changes.
     # Add an upper bound version only if automatic upgrades must be prevented.
     "entitysdk>=0.13.3",
-    "pydantic>=2.10.6",
+    "pydantic>=2.10.6,<=2.12.5",
```
GianlucaFicarelli, line 36: @ilkilic why pydantic has been capped to <=2.12.5?

ilkilic: Sorry, I hit some errors with the connectome-analysis tests earlier but I just reran them after updating deps and they are passing now. I just opened a [PR](https://github.com/openbraininstitute/obi-one/pull/743) to uncap it.

GianlucaFicarelli: np, I've seen that the tests were passing with the latest pydantic so the upper bound has been reverted already in https://github.com/openbraininstitute/obi-one/pull/742
In general if an upper bound is set, it should be temporary and it would be better to add a comment to explain the reason, so it's easier to verify if it can be removed later.

ilkilic: I agree and thank you for taking care of it. I will close my PR

PR #726, opened 2026-04-28
Switch morphometrics computation from SWC to H5 assets

ilkilic:
## Summary

Switch morphometrics computation from SWC to H5 assets and add tests.

## Changes

- Use H5 morphology assets in computation pipeline
- Return `400` if no H5 asset is available
- Add golden test using existing SWC-based reference data
- Add endpoint test for missing H5

## Notes
- Golden reference is SWC-based, so it validates computation logic but not full H5 → metrics equivalence


dkeller9 (approved):
LGTM, good to be explicit that it is from swc.

PR #727, opened 2026-04-28
initial implementation of accounting for circuit extraction

jdcourcol:
Adds a new end point to estimate the cost of a circuit extraction.

This computes the size of the neuron set from the configuration file and forward it to the accounting service to get the credit estimate.


tests/obi_one/scientific/tasks/test_circuit_extraction_estimate.py
```diff
@@ -0,0 +1,123 @@
...
+        ),
+        patch(
+            "obi_one.scientific.tasks.circuit_extraction.estimate.CircuitExtractionSingleConfig.model_validate",
+            return_value=fake_config,
+        ),
+    ):
+        assert estimate_circuit_extraction_count(db_client=db_client, config_id=config_id) == 3
+
+
+def test_estimate_circuit_extraction_count_has_minimum_one_for_empty_set():
```
jdcourcol, line 46: To be decided @chr-pok . what do we do for empty set ?

chr-pok: I think it would be best to raise an early error for empty neuron sets before even launching the task, since such circuits don't make much sense (similar to circuit simulations with empty target neuron sets).

But also fine to leave it as is, since technically it will still create a circuit config but w/o any populations.

obi_one/scientific/tasks/circuit_extraction/estimate.py
```diff
@@ -0,0 +1,58 @@
...
+from entitysdk.client import Client
+from entitysdk.types import AssetLabel
+
+from obi_one import deserialize_obi_object_from_json_data
+from obi_one.scientific.from_id.circuit_from_id import CircuitFromID
+from obi_one.scientific.tasks.circuit_extraction.task import CircuitExtractionSingleConfig
+from obi_one.utils import db_sdk
+
+
+def estimate_circuit_extraction_count(*, db_client: Client, config_id: UUID) -> int:
```
chr-pok, line 18: Just that I understand correctly: Here, the actual count is return, but the actual mapping from count to cost (using tiers) is defined in the accounting service, right?

obi_one/scientific/tasks/circuit_extraction/estimate.py
```diff
@@ -0,0 +1,58 @@
...
+    config_dict = json.loads(config_bytes.decode("utf-8"))
+
+    single_config = CircuitExtractionSingleConfig.model_validate(
+        deserialize_obi_object_from_json_data(config_dict).model_dump()
+    )
+
+    parent_circuit = single_config.initialize.circuit
+    if isinstance(single_config.initialize.circuit, CircuitFromID):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            parent_circuit = single_config.initialize.circuit.stage_circuit(
```
chr-pok: I think in most cases it would not be required to stage the full circuit for resolving the neuron set, fetching the node file(s) alone would be sufficient (as we do for circuit metrics). But with file mount, this should not be an issue.

Also, when launching an extraction, the circuit is being fetched twice, once for resource estimation and once for accounting. But since these steps are independent and task-specific, I suppose there is no way around that (?).

chr-pok (dismissed):
Thanks for the implementation! Looks good to me and works nicely. Just had a few questions for better understanding.

obi_one/scientific/tasks/circuit_extraction/estimate.py
jdcourcol: exactly. We provide inputs (here the number of neurons) to accounting. and it uses that, depending on the subcription of the user and the product to define a cost.

obi_one/scientific/tasks/circuit_extraction/estimate.py
jdcourcol: yes, I notice that too. 
However, estimate endpoint is used standalone (to tell the user how much it is going to cost before launching the task).

Probably we should change resource estimate, so that we could reuse that information to avoid reloading the circuit.

tests/obi_one/scientific/tasks/test_circuit_extraction_estimate.py
jdcourcol: yeah, it is like launching a simulation with 0 neurons. It doesn't seem to make sense.

obi_one/scientific/tasks/circuit_extraction/estimate.py
chr-pok: Yeah, could be changed in a separate PR. Also, resource estimation could be refined using the output circuit size; I am using a coarse approximation atm since we did not have that count available before.

app/endpoints/circuit_extraction.py
```diff
@@ -0,0 +1,33 @@
...
+from app.mappings import TASK_DEFINITIONS
+from app.schemas.task import TaskAccountingInfo
+from app.services import accounting as accounting_service
+from app.types import TaskType
+
+router = APIRouter(prefix="/declared", tags=["declared"], dependencies=[Depends(user_verified)])
+
+
+@router.post(
+    "/circuit-extraction/estimate",
```
eleftherioszisis: Don't we already have an estimate endpoint for tasks?
https://github.com/openbraininstitute/obi-one/blob/e26af339f7aa70e20b71f352c4751d88af0e9715/app/endpoints/task.py#L122

obi_one/scientific/tasks/circuit_extraction/estimate.py
eleftherioszisis: Organization-wise, is cost-estimation a concept that should exist within the obi_one library? Isn't it only service specific?

app/endpoints/circuit_extraction.py
jdcourcol: indeed, should reuse that and have a switch inside rather than on the route.

obi_one/scientific/tasks/circuit_extraction/estimate.py
jdcourcol: it is task specific.

app/endpoints/circuit_extraction.py
jdcourcol: actually , I can just reuse that endpoint w/o any modification.

chr-pok (approved):
Finalized PR and ready to be merged, as discussed.

PR #736, opened 2026-04-30
bluerecording in obi-one

cattabiani:
WIP


Makefile
```diff
@@ -39,6 +39,68 @@ install-service:  ## Install with service dependencies (for running as a service
...
+	@echo 'export PATH="$(NEURODAMUS_NEOCORTEX_ROOT)/bin:$$PATH"' >> bluerecording_env.sh
+	@echo ""
+	@echo "=== BlueRecording installation complete ==="
+	@echo "To activate the environment, run:"
+	@echo ""
+	@echo "    source bluerecording_env.sh"
+	@echo ""
+	@echo "NOTE: Do not use 'uv run' for bluerecording commands — it will overwrite the MPI-enabled h5py."
+	@echo "      Use 'source bluerecording_env.sh' which activates the venv directly."
+	@echo ""
```
eleftherioszisis, line 102: obi-one should have no responsibility in installing bluerecording by compiling neuron/neurodamus/h5py-mpi etc., potentially breaking all other tasks, but rather declare the optional dependency. All this should be in bluerecording repo along with docs how to install it.

eleftherioszisis (changes requested):

Makefile
cattabiani: hoe exactly can I run [this](https://github.com/openbraininstitute/obi-one/blob/extracellular_electrode_array/examples/J_create_recording_array/create_recording_array.ipynb) locally? Please, 1 solution because I asked @james-isbister and he said this was ok. You come along and you tell me I need to redo everything. Who is in charge?

cattabiani: obi-one has a lot of responsibilities in installing other stuff. Why not bluerecording (which, in my branch, is an optional dep)

eleftherioszisis: Yes, it is called technical debt.

cattabiani: ah so we want to remove `make-install` and everything explained in the readme? Good to know

Should I just pip install obi-one?

This question still stands:

>how exactly can I run [this](https://github.com/openbraininstitute/obi-one/blob/extracellular_electrode_array/examples/J_create_recording_array/create_recording_array.ipynb) locally?

eleftherioszisis: You can do it by installing bluerecording in your venv and then obi-one with the fulfilled dep

cattabiani: that is not how everything works in obi-one but ok. I will do that if @james-isbister is fine with that. It actually simplifies a lot my job

james-isbister: That works for me, and sounds like a good approach. I just said do what’s simplest for you :D

james-isbister: Sorry for confusion. I didn’t fully think through the consequences of the two options. Just wanted something that worked for testing the task locally

cattabiani:
@eleftherioszisis does not like this approach. I will try something else

PR #737, opened 2026-04-30
EM synapse mapping - Advanced Block for population name overrides 

james-isbister:
This creates a new block for overriding the default population names as mentioned in the [em synapse mapping ticket](https://github.com/openbraininstitute/prod-build-circuit/issues/29#issuecomment-4630591755)

If the user leaves the fields blank, the default values will be used


obi_one/scientific/tasks/em_synapse_mapping/task.py
```diff
@@ -235,10 +236,20 @@ def execute(  # NOQA: PLR0914, PLR0915, C901, PLR0912
...
-        pop_virt = init.virtual_node_population
-        pop_edge_int = init.physical_edge_population_name
-        pop_edge_ext = init.virtual_edge_population_name
+
+        pop_edge_ext = "virtual_afferents"
+        pop_edge_int = "physical_connections"
+        pop_bio = "biophysical_neurons"
+        pop_virt = "virtual_afferent_neurons"
+
+        if advanced.custom_virtual_edge_population_name:
```
jplanasc, line 245: These `if`s are not checked by the tests in case they're `True`.

jplanasc (dismissed):
LGTM, just one minor comment about testing.

obi_one/scientific/tasks/em_synapse_mapping/task.py
james-isbister: Added tests for both test_task and test_task_multineuron

jplanasc (approved):
LGTM, thanks for the tests!

PR #740, opened 2026-04-30
Push obi-one to private ECR

danifr:


.github/workflows/publish.yml
GianlucaFicarelli: There are still some references to the public ECR.
Isn't possible to reuse the same job defined for launch-system [here](https://github.com/openbraininstitute/launch-system/blob/main/.github/workflows/publish.yml) without the Azure part?

danifr: yes, I was thinking of keeping both public and private repos for awhile until we are ready to sunset the public one, but if you want we can do it in one go

GianlucaFicarelli: Yes we should completely remove the logic that pushes to the public repository before merging and releasing https://github.com/openbraininstitute/obi-one/pull/649, so it's better to proceed with that change and ensure that the service can be deployed pulling the image from the private repository.
Optionally, we could remove the old images from the public ECR because they can be confusing, but this can be decided and done later.

danifr: @GianlucaFicarelli alright. I pushed again, this time the CI looks like the launch-system one and I also remove the public registry push steps. Give a go when you have some time, thanks

GianlucaFicarelli (approved):
LGTM!

PR #744, opened 2026-05-01
Support I sim configuration

james-isbister:
* Could be improved, but this allows us to test the simulator in the frontend

### DIFF
--- a/obi_one/scientific/blocks/neuron_sets/specific.py
+++ b/obi_one/scientific/blocks/neuron_sets/specific.py
@@ -32,6 +32,9 @@ class AllNeurons(AbstractNeuronSet):
     @staticmethod
     def check_node_set(circuit: Circuit, _population: str) -> None:
         if _ALL_NODE_SET not in circuit.node_sets:
+            L.info(_population)
+            if _population == "core":
+                return
             msg = (
                 f"Node set '{_ALL_NODE_SET}' not found in circuit '{circuit.name}'. "
                 f"Available node sets: {', '.join(circuit.node_sets)}"
@@ -41,6 +44,10 @@ class AllNeurons(AbstractNeuronSet):
     def _get_expression(self, circuit: Circuit, population: str) -> list:  # ty:ignore[invalid-method-override]
         """Returns the SONATA node set expression (w/o subsampling)."""
         self.check_node_set(circuit, population)
+
+        if population == "core":
+            return ["core"]
+
         return [_ALL_NODE_SET]
 
 
--- a/obi_one/scientific/blocks/neuron_sets/specific.py
+++ b/obi_one/scientific/blocks/neuron_sets/specific.py
@@ -32,7 +32,6 @@ class AllNeurons(AbstractNeuronSet):
     @staticmethod
     def check_node_set(circuit: Circuit, _population: str) -> None:
         if _ALL_NODE_SET not in circuit.node_sets:
-            L.info(_population)
             if _population == "core":
                 return
             msg = (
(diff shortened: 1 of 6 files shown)


obi_one/scientific/tasks/generate_simulations/config/base.py
```diff
@@ -61,7 +61,6 @@ class BlockGroup(StrEnum):
     EVENTS_GROUP = "Events"
 
 
-TARGET_SIMULATOR = "NEURON"
```
eleftherioszisis, line 64: You can use the TargetSimulator StrEnum from entitysdk:
https://github.com/openbraininstitute/entitysdk/blob/5730cfb2083db50e2d0d02959f1ddfca5fe737aa/src/entitysdk/types.py#L46

james-isbister: Nice, @mgeplf suggested I get rid of it for now from the simulation configs as the circuits should have it specified

We can put it back if we try to support switching between NEURON / CORENEURON but for now we just want to test the inait sims

eleftherioszisis: Note that inait sims need to have a "LearningEngine" target_simulator specified.

james-isbister: It's not read from the circuit_config?

obi_one/scientific/blocks/stimuli/stimulus.py
```diff
@@ -144,7 +144,8 @@ def config(
         self._default_timestamps = default_timestamps
 
         if (self.neuron_set is not None) and (
-            self.neuron_set.block.population_type(circuit, population) != "biophysical"  # ty:ignore[unresolved-attribute]
+            self.neuron_set.block.population_type(circuit, population)  # ty:ignore[unresolved-attribute]
+            not in {"biophysical", "inait_point_neuron_lif"}
```
jdcourcol, line 148: should we instead ignore the virtual ones.

obi_one/scientific/blocks/neuron_sets/specific.py
```diff
@@ -32,6 +32,8 @@ class AllNeurons(AbstractNeuronSet):
     @staticmethod
     def check_node_set(circuit: Circuit, _population: str) -> None:
         if _ALL_NODE_SET not in circuit.node_sets:
+            if _population == "core":
```
jdcourcol, line 35: what is "core" ? I don't know that concept.

james-isbister: That's the name of the node_set which seems to come as standard with the inait circuit

jdcourcol: I am not sure I understand what is going on. Why do we have to check the existence of a population named "core" ?

james-isbister: I was just aiming to get the simulation generation working asap with the time I had available on Friday, so we could test generating and launching with the frontend. 

Supporting multiple simulators properly, needs some careful consideration, which I'd suggest we leave to another PR

obi_one/scientific/blocks/stimuli/stimulus.py
james-isbister: Same as the other comment. I'd rather think more carefully about this, which I might not have time for before I go away on Wednesday, so maybe better to test as is?

obi_one/scientific/tasks/generate_simulations/config/base.py
jdcourcol: that is what I would expect. if not in simulation config then the value comes from the circuit config

obi_one/scientific/blocks/neuron_sets/specific.py
jdcourcol: I see. I understand as a lack in sonata to identify the node_set defining all simulatable neurons. although nothing prevents the simulation UI to create a nodeset including all populations that are not virtual.

jdcourcol (approved):
I am good. There is technical debt but if we can demo and get users to run simulation that is fine.

PR #747, opened 2026-05-01
Remove numpy constraint now that connectome-utilities 0.4.12 is out

danifr:
All tests should pass now with no warnings nor errors

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ dependencies = [
     "pylmesh",
     "python-multipart>=0.0.26",
     "obp-accounting-sdk>=0.6.0",
+    "numpy",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ dependencies = [
     "bluepysnap>=3.0.2",
     "brainbuilder>=0.20.2",
     "caveclient>=7.11.0",
-    "connectome-utilities>=0.4.11",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -53,7 +53,6 @@ dependencies = [
     "pylmesh",
     "python-multipart>=0.0.26",
     "obp-accounting-sdk>=0.6.0",
-    "numpy<2.4", # Needs https://github.com/openbraininstitute/ConnectomeUtilities/pull/5
```
GianlucaFicarelli, line 56: Although it wasn't listed before, I would keep the numpy dependency, without upper bound, as it's used explicitly in obi-one.

uv.lock
```diff
@@ -548,7 +548,7 @@ wheels = [
 
 [[package]]
 name = "connectome-utilities"
-version = "0.4.11"
```
GianlucaFicarelli, line 551: To update numpy and verify that tests pass, could you update the lockfile with:
```
make upgrade-deps
```
or, if there are too many packages being updated (I don't expect many, since they were updated yesterday), just numpy with:
```
uv lock --upgrade-package numpy
```

pyproject.toml
danifr: Cool, I added it back

uv.lock
danifr: Done too. thanks

GianlucaFicarelli (approved):

PR #748, opened 2026-05-04
Add spike-time distribution spike stimulus

ilkilic:
Adds a distribution-based spike stimulus that generates spike trains by sampling spike times directly from probability distributions.

This complements the existing inter-spike interval (ISI) based stimulus (https://github.com/openbraininstitute/obi-one/pull/723) by supporting direct spike-time sampling.

## What’s included

* `SpikeTimeDistributionSpikeStimulus` for direct spike-time sampling
* Default `FloatUniformDistribution(low=0.0, high=duration)` when no distribution is provided
* `mean_firing_rate` parameter to control expected spike count per neuron
* Spike counts sampled per neuron/repetition (Poisson process based on duration)
* Support for `random_seed` through the referenced distribution block
* Support for `resample_each_repetition`
* Filtering to ensure sampled spike times remain within the stimulus duration
* Tests for spike-time sampling behavior and generated spike files
* Example notebook demonstrating block composition and visual validation


james-isbister (commented):
Thanks @ilkilic 

I'm not sure about using `number_of_spikes` to define how spikes are drawn from each neuron. 

Possibly something like a mean_firing_rate over all neurons over the `duration` could make sense, with spikes for each neuron drawn independently so we have different numbers of spikes per neuron. 

What do you think of that?

ilkilic:
Good point, I have updated this to use mean_firing_rate instead of a fixed number_of_spikes.

james-isbister (approved):
Looks great. Thank you @ilkilic 

I think it would be great to add log-normal and gaussian distributions in a separate PR if/when you have time, as these are commonly used for this type of stimulus

ilkilic:
Thanks! Agreed, those distributions would be very useful. I will open a follow-up issue for it.

PR #751, opened 2026-05-05
calculate mesh during morphology registration

dkeller9:
The morphology regstration endpoint now uses the functions in convert_morphology_to_registered_mesh to compute and register the mesh during normal registration. Failure of the meshing is not considered a fatal error for the purposes of this endpoint, provided that everything else is properly registered.


app/endpoints/morphology_metrics_calculation.py
```diff
@@ -29,6 +30,12 @@
 import app.endpoints.useful_functions.useful_functions as uf
 from app.dependencies.auth import user_verified
 from app.dependencies.entitysdk import get_client
+from app.endpoints.convert_morphology_to_registered_mesh import (
+    HAS_MESHING,
+    _mesh_and_register,
```
eleftherioszisis: The protected name signifies that the function is local to a module, which is not the case here.

app/endpoints/morphology_metrics_calculation.py
eleftherioszisis: Isn't it possible to return a pydantic schema here which has the files information? For example:

```
class MorphologyFiled(BaseModel):
    swc: Path
    hdf5: Path
```

In addition, the convert_morphology() in app/services/morphology.py has all the information it needs to construct an 1-1 mapping schema from extensions to target files. Instead of returning a flat list of paths losing the information of which extension returns which file, return a schema model instead and then use the swc_file downstream without searching.

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -429,15 +471,30 @@ async def _run_pipeline(
...
         )
-        return entity_id, str(data2.id)  # ty:ignore[unresolved-attribute]
+
+        swc_bytes_for_mesh: bytes | None = await run_in_threadpool(
+            _resolve_swc_bytes_for_mesh,
+            converted_morphology_file1,
+            converted_morphology_file2,
+            file_extension,
+            content,
+        )
```
eleftherioszisis: See my comment above. This is not needed.

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -481,6 +538,7 @@ async def morphology_metrics_calculation(
         return {
             "entity_id": entity_id,
             "measurement_entity_id": measurement_entity_id,
+            "mesh_asset_id": mesh_asset_id,
             "status": "success",
             "morphology_name": morphology_name,
         }
```
eleftherioszisis: Endpoints should return pydantic schemas instead of raw data. See app/endpoints.task.py for examples.

app/endpoints/morphology_metrics_calculation.py
dkeller9: The protected name has been fixed.

app/endpoints/morphology_metrics_calculation.py
dkeller9: The endpoint now returns a Pydantic schema.

A new MorphologyFiles schema was added to app/services/morphology.py. It explicitly defines fields for swc and hdf5 paths rather than returning an unordered list.

app/endpoints/morphology_metrics_calculation.py
dkeller9: A pydantic schema is now returned.

app/endpoints/morphology_metrics_calculation.py
dkeller9: This has been refactored.

dkeller9:
@eleftherioszisis thanks for the comments. I've revised the code accodingly. Please let me know if that is what you had in mind.

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -356,36 +366,54 @@ def _register_assets_and_measurements(
     morphology_name: str,
     content: bytes,
     measurement_list: list[dict[str, Any]],
-    converted_morphology_file1: str,
-    converted_morphology_file2: str,
+    converted_files: MorphologyFiles,
 ) -> dict[str, Any]:
     with tempfile.TemporaryDirectory() as temp_dir_for_upload:
         temp_upload_path_obj = pathlib.Path(temp_dir_for_upload) / morphology_name
         temp_upload_path_obj.write_bytes(content)
```
eleftherioszisis: It seems this path is unused. I suppose it is implicitly used in the function below. This shouldn't be the case but rather pass the explicit path.

In addition, the content can be directly registered without an intermediate file using entitysdk Client's upload_content instead of upload_file.

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -268,35 +278,67 @@ def _get_entity(key_suffix: str, entity_class: type[T]) -> T | None:
     return registered
 
 
-def register_assets(
+EXTENSION_MIME_MAP: Final[dict[str, str]] = {
+    ".asc": "application/asc",
+    ".swc": "application/swc",
+    ".h5": "application/x-hdf5",
```
eleftherioszisis: We been over that in previous PRs, use entitysdk.types ContentType instead of raw strings. The same goes for all other entitysdk types like AssetLabel, etc.

app/endpoints/morphology_metrics_calculation.py
dkeller9: Good point, the code has been refactored accordingly.

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -268,35 +278,67 @@ def _get_entity(key_suffix: str, entity_class: type[T]) -> T | None:
...
 
-    if not file_path_obj.exists():
+    try:
+        asset = client.upload_content(
+            entity_id=entity_id,  # ty:ignore[invalid-argument-type]
+            entity_type=CellMorphology,
+            content=content,  # ty:ignore[invalid-argument-type]
+            file_name=morphology_name,  # ty:ignore[invalid-argument-type]
+            file_content_type=mime_type,  # ty:ignore[invalid-argument-type]
+            asset_label="morphology",  # ty:ignore[invalid-argument-type]
```
eleftherioszisis: Gianluca enabled static type checking so that we improve the quality and correctness of the code. Please remove the ignore directives and try addressing the issues. They should never be added unless there is an unfixable issue that cannot be addressed otherwise.

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -268,35 +278,67 @@ def _get_entity(key_suffix: str, entity_class: type[T]) -> T | None:
...
-        error_msg = f"Unsupported file extension: '{file_extension}'."
-        raise ValueError(error_msg)
+    mime_type = _get_mime_type(file_path.suffix)
 
     try:
         asset1 = client.upload_file(
             entity_id=entity_id,  # ty:ignore[invalid-argument-type]
             entity_type=CellMorphology,
-            file_path=file_path,  # ty:ignore[invalid-argument-type]
+            file_path=str(file_path),  # ty:ignore[invalid-argument-type]
```
eleftherioszisis: Does passing a Path not work here?

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -356,36 +402,68 @@ def _register_assets_and_measurements(
     morphology_name: str,
     content: bytes,
     measurement_list: list[dict[str, Any]],
-    converted_morphology_file1: str,
-    converted_morphology_file2: str,
+    converted_swc_path: str | None,
+    converted_hdf5_path: str | None,
```
eleftherioszisis: Use MorphologyFiles

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -395,28 +473,29 @@ async def _run_pipeline(
...
             output_stem=Path(morphology_name).stem,
             single_point_soma_by_ext=single_point_soma_by_ext,
         )
-        if converted_morphology_file1:
-            stack.callback(pathlib.Path(converted_morphology_file1).unlink, missing_ok=True)
-        if converted_morphology_file2:
-            stack.callback(pathlib.Path(converted_morphology_file2).unlink, missing_ok=True)
+
+        converted_swc = str(converted_files.swc) if converted_files.swc else None
+        converted_hdf5 = str(converted_files.hdf5) if converted_files.hdf5 else None
```
eleftherioszisis: The reason I suggested MorphologFiles is so that you use it downstream to simplify the multiple paths passing.

app/endpoints/morphology_metrics.py
```diff
@@ -33,12 +34,24 @@ def neuron_morphology_metrics_endpoint(
     cell_morphology_id: str,
     db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
     requested_metrics: Annotated[
-        list[Literal[*MORPHOLOGY_METRICS]] | None,  # type: ignore[misc]  # ty:ignore[invalid-type-form]
+        list[str] | None,
```
eleftherioszisis: Given that there is a list of MORPHOLOGY_METRICS to validate the allowed metrics, would it work if a StrEnum MorphologyMetric was used here as `list[MorphologyMetric]`so that it triggers pydantic validation that the fastapi level? You wouldn't need the manual validation below with the invalid metrics.

app/endpoints/morphology_metrics.py
```diff
@@ -74,7 +87,7 @@ def register_morphology_metrics(
     db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
 ) -> dict:
     # 1) fetch morphology and its H5 asset
-    morph = db_client.get_entity(entity_id=cell_morphology_id, entity_type=CellMorphology)  # ty:ignore[invalid-argument-type]
+    morph = db_client.get_entity(entity_id=UUID(cell_morphology_id), entity_type=CellMorphology)
```
eleftherioszisis: Seems like cell_morphology_id should have a UUID type to begin with above.

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -242,17 +256,34 @@ def _get_entity(key_suffix: str, entity_class: type[T]) -> T | None:
...
+        ),
+    ):
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST,
+            detail={
+                "code": ApiErrorCode.BAD_REQUEST,
+                "detail": "A valid cell_morphology_protocol_id is required.",
+            },
+        )
+    morphology_protocol: CellMorphologyProtocolUnion = raw_protocol
```
eleftherioszisis: I am curious, why is this needed? What would be an invalid case?

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -242,17 +256,34 @@ def _get_entity(key_suffix: str, entity_class: type[T]) -> T | None:
...
     repair_pipeline_state = new_item.get("repair_pipeline_state")
 
-    license = _get_entity("license", License)  # ty:ignore[invalid-argument-type]
+    license = _get_entity("license", License)
     name = new_item.get("name")
     description = new_item.get("description")
-    authorized_public = new_item.get("authorized_public")
+    authorized_public: bool = bool(new_item.get("authorized_public"))
```
eleftherioszisis: Casting None to bool is not very nice. Would `new_item.get("authorized_public", False)` work?

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -261,44 +292,78 @@ def _get_entity(key_suffix: str, entity_class: type[T]) -> T | None:
...
+    morphology_name: str,
+    content: bytes,
+) -> Asset:
+    file_extension = pathlib.Path(morphology_name).suffix
+    content_type = _get_content_type(file_extension)
+    try:
+        asset = client.upload_content(
+            entity_id=UUID(entity_id),
+            entity_type=CellMorphology,
+            file_content=io.BytesIO(content),
```
eleftherioszisis: This is a issue of entitysdk not having broad enough type to cover bytes.
Will be fixed with https://github.com/openbraininstitute/entitysdk/pull/229
If entitysdk is not released fast enough to cover this, please ignore it instead of converting bytes to a stream.

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -261,44 +292,78 @@ def _get_entity(key_suffix: str, entity_class: type[T]) -> T | None:
...
+            file_name=morphology_name,
+            file_content_type=content_type,
+            asset_label=AssetLabel.morphology,
+        )
+    except requests.exceptions.RequestException as e:
```
eleftherioszisis: Does this capture anything meaningful? I would expect entitysdk to throw an EntitySDKError for all request related issues.

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -261,44 +292,78 @@ def _get_entity(key_suffix: str, entity_class: type[T]) -> T | None:
...
+                "detail": f"Entity asset registration failed: {e}",
+            },
+        ) from e
+    else:
+        return asset
+
+
 def register_assets(
     client: Client,
     entity_id: str,
```
eleftherioszisis: Should be UUID

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -309,19 +374,19 @@ def register_assets(
             },
         ) from e
     else:
-        return asset1  # ty:ignore[invalid-return-type]
+        return asset1
 
 
 def register_measurements(
     client: Client,
     entity_id: str,
```
eleftherioszisis: Should be UUID

app/endpoints/morphology_metrics_calculation.py
```diff
@@ -309,19 +374,19 @@ def register_assets(
...
         measurement_annotation = MeasurementAnnotation(
-            entity_id=entity_id,  # ty:ignore[invalid-argument-type]
-            entity_type="cell_morphology",  # ty:ignore[invalid-argument-type]
-            measurement_kinds=measurements,  # ty:ignore[invalid-argument-type]
+            entity_id=UUID(entity_id),
+            entity_type=MeasurableEntity.cell_morphology,
+            measurement_kinds=measurements,
         )
         registered = client.register_entity(entity=measurement_annotation)
     except requests.exceptions.RequestException as e:
```
eleftherioszisis: Same here. Not sure if this does indeed capture anything. Please correct me if wrong.

app/endpoints/morphology_metrics.py
dkeller9: Yes, a StrEnum-based MorphologyMetric used as list[MorphologyMetric] lets Pydantic/FastAPI handle validation automatically, eliminating the manual invalid-metrics check.

app/endpoints/morphology_metrics.py
dkeller9: Agreed; typing it as UUID directly in the endpoint signature lets FastAPI validate and parse it automatically, removing the need for any manual UUID() cast downstream.

app/endpoints/morphology_metrics_calculation.py
dkeller9: `_get_entity `returns None when no id is supplied or on lookup failure, and may also return a bare CellMorphologyProtocol base-class instance if entitysdk behaves unexpectedly, so the check guards against both cases.

app/endpoints/morphology_metrics_calculation.py
dkeller9: Yes, `new_item.get("authorized_public", False)` is cleaner and makes the default explicit without silently coercing `None`.

app/endpoints/morphology_metrics_calculation.py
dkeller9: left the `BytesIO` wrapper in place for now but can be updated when entitysdk#229 is released

app/endpoints/morphology_metrics_calculation.py
dkeller9: No; entitysdk wraps its HTTP calls in EntitySDKError, so RequestException would never be reached and has been replaced with EntitySDKError throughout.

app/endpoints/morphology_metrics_calculation.py
dkeller9: Fixed, the signature now accepts UUID directly and the internal UUID(entity_id) conversion has been removed.

app/endpoints/morphology_metrics_calculation.py
dkeller9: Fixed.

app/endpoints/morphology_metrics_calculation.py
dkeller9: Fixed.

app/endpoints/morphology_metrics_calculation.py
dkeller9: Yes, I have now updated this file to use `ContentType` and `AssetLabel`.

app/endpoints/morphology_metrics_calculation.py
dkeller9: All of these types of ignore directive have been removed and the issues fixed with some additional refactoring.

app/endpoints/morphology_metrics_calculation.py
dkeller9: Changed to `path`.

app/endpoints/morphology_metrics_calculation.py
dkeller9: `MorphologyFiles` is now used.

app/endpoints/morphology_metrics_calculation.py
eleftherioszisis: v0.15.6 is released with the fix.

app/endpoints/morphology_metrics_calculation.py
eleftherioszisis:
```
and may also return a bare CellMorphologyProtocol base-class instance if entitysdk behaves unexpectedly,
```
This should not be possible. Pydantic type adapter for CellMorphologyProtocol ensures that the correct type is constructed based on the protocol type discriminated union. If there is a repro of such an issue we should fix it. The expectation is to always return a specific protocol type.

dkeller9: Okay, probably pydantic will always endeavor to return a type but I suspect there still may be edge cases when None is returned. So I would rather be on the safe side with checks.

app/endpoints/morphology_metrics_calculation.py
dkeller9: The `BytesIO` wrapper is now removed as the latest entitysdk is released.

app/endpoints/morphology_metrics_calculation.py
eleftherioszisis: Why does a function that is named "_get_entity()" returns the entity or None? That is quite unexpected given the name of the function. In any case, if you need to check if the entity is None, then check that. If there is an edge case where the deserialization is not working in entitysdk, then we should fix it there, not here.

dkeller9: The test is now removed entirely, since perhaps it is not needed.

eleftherioszisis (approved):

PR #752, opened 2026-05-05
bluerecording in obi-one (second take)

cattabiani:
## How to

This is mostly for @james-isbister 

Since @eleftherioszisis does not want to have a full bluerecording dependency chain here we proceed like this:
install `BlueRecording` on your own. You will need the `katta/circuit-config-support` branch. I suggest this:

```
git clone https://github.com/openbraininstitute/BlueRecording.git
cd BlueRecording
git checkout katta/circuit-config-support
./dev_setup.sh
source env.sh
```

This will activate an environment with bluerecording and everything else that is needed. After that you can go into your obi-one folder, with the bluerecording env active and:

```
pip install -e ".[notebooks]"
```

Now you are able to run the notebook:

```
jupyter lab examples/J_create_recording_array/create_recording_array.ipynb
```


james-isbister:
Got to the pip install step but then it seems to take a while / get stuck on: `Building wheel for pylmesh (pyproject.toml)`

The code looks good though and the instructions are clear.

If you could just add the install instructions to the notebooky for prosperit, then I can approve once the installation finishes and the notebook runs fully

cattabiani:
>If you could just add the install instructions to the notebooky for prosperit, then I can approve once the installation finishes and the notebook runs fully

you mean to the notebook for posterity right?

cattabiani:
@james-isbister when you mention the `pip install step` do you mean this:

```
pip install -e ".[notebooks]"
```

in obi-one? Or the internal pip install of bluerecording? because they are both pretty fast (<1min) in my installation. 

To be extra sure I will check with the docker image and add instructions based on that

what do you think?

james-isbister:
It worked! I was traveling, hence the dodgy spelling and impatience...

Indeed, if you can add the instructions to the "notebook for posterity" I'm happy for @chr-pok or @ilkilic to accept tomorrow on my behalf. Thanks

obi_one/scientific/tasks/create_recording_array/create_recording_array.py
```diff
@@ -192,24 +186,47 @@ def execute(
...
+            electrodes=electrodes,
+            replace_axons=True,
+        )
+        L.info("weights shape: %s", weights.shape if weights is not None else None)
+        L.info("positions_df shape: %s", positions_df.shape)
+        L.info("cols shape: %s", cols.shape)
+        L.info("neurite_types shape: %s", neurite_types.shape)
+        L.info("population_name: %s", population_name)
+
+        weights_output_path = Path(circuit_dest_dir) / "weights.h5"
```
james-isbister: The weights can be written to self.config.coordinate_output_root / “weight.h5”

cattabiani: done

cattabiani:
I tried to install in the docker but, locally, the browser auth did not work. I may retry in the future if you want

ilkilic (dismissed):
Thank you @cattabiani

pyproject.toml
```diff
@@ -88,6 +88,13 @@ all = [
     "obi-one[connectivity,notebooks,service]",
 ]
 
+# BLUERECORDING - optional, needed for extracellular recording array tasks.
+# Requires MPI-enabled h5py and NEURON built from source.
+# Use BlueRecording/dev_setup.sh for full setup, then pip install -e obi-one[bluerecording]
+bluerecording = [
```
james-isbister, line 102: Is this still needed now we install with the dev_setup.sh script?

james-isbister (commented):
I tried merging main into this branch and was going to fix the linting issues

I noticed there is still the bluerecording dependency group though. Do we still need it?

After that I think I can make the fixes and we can merge into main today so I can start on the next steps

pyproject.toml
cattabiani: @eleftherioszisis I think this is the preferred solution, yes. In the platform/docker, when everything is installed properly, you can just install the correct obi-one version with pip install and it should pick and install bluerecording. In that environment it should be pip installable. If it is already installed (as it may very well be), pip just skips (correct behavior)

eleftherioszisis: Although the name of the optional dependency may change depending on how we separate the dep groups, the dependency resolver should be aware that bluerecording is needed for the task. Depending on how the task is going to be used, this can be stated here or in the task's requirements.txt

james-isbister: Ok, we'll can get this running with the launch system in a follow up PR

james-isbister (approved):
Have merged main and fixed the linting erros

PR #757, opened 2026-05-07
Move distributions into separate GUI block group

ilkilic:
Moves distributions out of the “Stimuli & Recordings” GUI section into their own dedicated “Distributions” block group for clearer organization and reuse across features.

### DIFF
--- a/obi_one/scientific/tasks/generate_simulations/config/circuit.py
+++ b/obi_one/scientific/tasks/generate_simulations/config/circuit.py
@@ -53,6 +53,7 @@ class CircuitSimulationScanConfig(SimulationScanConfig):
         SchemaKey.GROUP_ORDER: [
             BlockGroup.SETUP_BLOCK_GROUP,
             BlockGroup.STIMULI_RECORDINGS_BLOCK_GROUP,
+            BlockGroup.DISTRIBUTIONS_BLOCK_GROUP,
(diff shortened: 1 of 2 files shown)


obi_one/scientific/tasks/generate_simulations/config/circuit.py
```diff
@@ -134,7 +135,7 @@ class Initialize(SimulationScanConfig.Initialize):
             SchemaKey.UI_ELEMENT: UIElement.BLOCK_DICTIONARY,
             SchemaKey.REFERENCE_TYPE: AllDistributionsReference.__name__,
             SchemaKey.SINGULAR_NAME: "Distribution",
-            SchemaKey.GROUP: BlockGroup.STIMULI_RECORDINGS_BLOCK_GROUP,
+            SchemaKey.GROUP: BlockGroup.DISTRIBUTIONS_BLOCK_GROUP,
             SchemaKey.GROUP_ORDER: 2,
```
chr-pok: If I am not mistaken, group order should be 0, since it is the only element in this new group.

ilkilic: Good catch! I have udpated it to 0.

chr-pok (approved):
Thanks, looks good!

PR #759, opened 2026-05-07
Isolate core dependencies from scientific

### LINKED ISSUE #696: Split core and scientific dependencies
## Description
OBI-ONE's core and scientific modules import classes from each other and their interdependencies are entangled. We should try to split them as much as possible to have a cleaner dependency schema, depending on which functionality we want to use.

## Acceptance criteria
- [ ] Core dependencies are extracted from scientific ones.

Time estimate:
- Most of the modules can be easily separated, this should take 2-3 days.
- There are a few modules (2-3) where separation will be more complicated and its feasibility needs to be evaluated.

jplanasc:
Fixes #696


obi_one/core/deserialize.py
```diff
@@ -3,21 +3,22 @@
...
-    return obi_object
+def deserialize_obi_object_from_json_data(json_dict: dict) -> OBIBaseModel:
+    cls = type_registry.get(json_dict["type"])
+    return cls.model_validate(json_dict)  # ty:ignore[unresolved-attribute]
 
 
-def deserialize_obi_object_from_json_file(json_path: Path) -> obi.OBIBaseModel:
+def deserialize_obi_object_from_json_file(json_path: Path) -> OBIBaseModel:
     with Path.open(json_path) as file:
         json_dict = json.load(file)
```
eleftherioszisis: Use `obi_one.utils.io.load_json`

obi_one/core/base.py
```diff
@@ -29,6 +31,9 @@ def __init_subclass__(cls, **kwargs) -> None:
         cls.__annotations__["type"] = Literal[cls.__qualname__]  # ty:ignore[invalid-type-form]
         cls.type = cls.__qualname__
 
+        # Auto-register for deserialization
+        type_registry.register(cls.__qualname__, cls)
```
eleftherioszisis: With very few exceptions I would always prefer a simple and explicit dictionary mapping stored in a module instead of dynamic registration of models using fancy design patterns.

Here you make the assumption that every single model inheriting this model should be registered, which may not be the case. In addition, does this solution work without importing all classes anyway?

Luigi task manager was notorious for its hard-to-debug dynamic task registry. Having PTSD from that I would prefer if we didn't do that here as well if possible. I asked chatgpt for a quick example of what I mean:

```python
TYPE_MAPPING = {
    "json": "myapp.parsers.json_parser.JsonParser",
    "yaml": "myapp.parsers.yaml_parser.YamlParser",
}
import importlib


from importlib import import_module


def load_class(type_name: str):
    path = TYPE_MAPPING[type_name]

    module_path, class_name = path.rsplit(".", 1)

    module = import_module(module_path)

    return getattr(module, class_name)
```

No fancy inheritance or complex mechanisms at play -> easy to read and debug. Also trivial to exhaustively test that all imports work in a parametrized unit test.

obi_one/core/deserialize.py
jplanasc: Done.

obi_one/core/base.py
jplanasc: Fair point, I've implemented the change in [7ab5f61](https://github.com/openbraininstitute/obi-one/pull/759/commits/7ab5f610f6eaa6b30b13e6facd41475c6de6d59f). I let you check and see if this is what you had in mind. Thanks!

obi_one/scientific/unions/config_task_map.py
```diff
@@ -71,57 +72,104 @@
...
+
+def _populate_task_registry() -> None:
+    """Populate the core TaskRegistry with all scientific task mappings.
+
+    This allows core/run_tasks.py to look up task classes, single configs,
+    and asset labels without importing scientific modules directly.
+    Called once at module load time.
+    """
+    # Config class -> Task class
+    task_registry.register_config_task(CircuitSimulationSingleConfig, GenerateSimulationTask)
```
james-isbister: Could we have a single function register_task(task, single_config, task_type, asset_label) where task_type and asset_label are optional?

james-isbister (commented):
Nice, looks good to me. Just one suggestion

obi_one/scientific/unions/config_task_map.py
jplanasc: @james-isbister I can merge everything into one function, but I'd find it a bit confusing in the sense that the `register_config_task` is called for all the tasks, but only a few (4 currently) use the `TaskType` registry. There will be "quite some" branching in the function to treat the multiple `None` values.

Maybe a middle ground would be leave `register_config_task` as is and then merge `register_task_type`, `register_task_type_single_config` and `register_task_type_config_asset_label` into a single one, since the latter 3 are tightly related?

I'm not pushing against your suggestion, just trying to think about pros and cons. Let me know what you finally think.

obi_one/core/deserializable_types.py
```diff
@@ -0,0 +1,71 @@
...
+Used by core/deserialize.py to resolve the concrete class for a given
+'type' field in serialized JSON without importing all scientific modules
+eagerly. All deserializable types must have an entry in this dictionary.
+"""
+
+# ruff: noqa: E501 — long lines are unavoidable in a path mapping dict
+
+from importlib import import_module
+
+TYPE_MAP: dict[str, str] = {
```
eleftherioszisis, line 15: Given that key is the class name couldn't this be simplified into having the module path as values?

e.g. 
```
"Block": "obi_one.core.block"
```

obi_one/core/registry.py
```diff
@@ -0,0 +1,101 @@
...
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from entitysdk.types import AssetLabel
+
+    from obi_one.types import TaskType
+
+
+class TaskRegistry:
```
eleftherioszisis, line 23: Is this needed now?

obi_one/scientific/unions/config_task_map.py
eleftherioszisis: I am confused. Why do we have dynamic registration here?

jplanasc: @eleftherioszisis : similarly to the `TYPE_MAPPING`, I need to break the core --> scientific dependency.

Either we do it dynamically, as it is done now, or we do it statically (following `TYPE_MAPPING` style) at the cost of losing the type checking/safety of the mappings.

Personally, I like this approach, but if you prefer everything static, I can also switch.

obi_one/core/deserializable_types.py
jplanasc: Good point, done in [ed8be31](https://github.com/openbraininstitute/obi-one/pull/759/commits/ed8be31e9c05cd95084d3456e344359bc8ae3cc9)

obi_one/core/registry.py
jplanasc: It depends :)

If we keep the dynamic registration, yes it's needed. If we switch to a fully static dictionary, I think we can remove it.

obi_one/scientific/unions/config_task_map.py
james-isbister: I think it wouldn't be so much branching to wrap the existing functions with a new one. I think it would be worth it, as there is a lot of repetition atm and it's too long to quickly get a mental overview. Also it will make it very easy to see which configs are missing task types, asset labels etc

james-isbister (commented):
Looks good. I still think my suggestion would be useful to go with as I've explained - the file is a bit overwhelming atm

obi_one/core/registry.py
eleftherioszisis: We shouldn't do dynamic registration. It will only complicate things and it is only needed if we provide a library to dynamically register tasks, which is not what obi-one is for.

obi_one/scientific/unions/config_task_map.py
eleftherioszisis: This is done statically, therefore it should be static. Dynamic registration patterns should be used when registration can only be done during runtime. Otherwise, the tradeoff of complexity vs. functionality just doesn't worth it.

In addition, wasn't the goal to avoid doing all these imports and only access what's needed for a given task?

jplanasc: @james-isbister , @eleftherioszisis : Ok, thanks for the feedback. I hope now this is what you had in mind? See [61c7be1](https://github.com/openbraininstitute/obi-one/pull/759/commits/61c7be1fb208999c886e7842593ed78454d9e9ca).

> @eleftherioszisis > wasn't the goal to avoid doing all these imports and only access what's needed for a given task?

The goal of this PR is to isolate core. The per-task split will come later, but first core needs to be independent.

obi_one/scientific/unions/config_task_map.py
```diff
@@ -75,58 +76,130 @@
...
-}
-_task_type_config_asset_label_map = {
-    TaskType.circuit_extraction: AssetLabel.task_config,
-    TaskType.morphology_skeletonization: AssetLabel.task_config,
-    TaskType.circuit_simulation: None,
-    TaskType.ion_channel_model_simulation_execution: None,
-    TaskType.em_synapse_mapping: AssetLabel.task_config,
+# Complete task registry: single_config_cls -> (task_cls, task_type, asset_label)
+# task_type and asset_label are None for tasks not launchable via the API.
+TASK_MAP: dict[type, tuple[type, TaskType | None, AssetLabel | None]] = {
```
eleftherioszisis: I am wondering, is this mapping desirable (in general not the specific changes here)? Shouldn't we have instead sth similar to how app tasks are mapped, i.e. task_type -> (config_cls, task_cls) ? @james-isbister

james-isbister: Yes, making the keys of TASK_MAP obi-one TaskTypes seems nice to me

jplanasc: @eleftherioszisis , @james-isbister : please, note:

- `TaskType` (e.g.: `TaskType.circuit_extraction`) is `None` for many tasks, so it can't be the key as it is right now.
- I can't use `type` (e.g.: `CircuitExtractionTask`) as the key, since `GenerateSimulationTask` is used in 4 task configs.

That being said, I'm happy to make the necessary changes to have one of the keys above if you want to.

james-isbister: That's a fair point.

I'd leave it as it is for now then. It's confined to one file so we could always change it later if we like

obi_one/core/deserializable_types.py
```diff
@@ -0,0 +1,73 @@
...
+    "MorphologyLocationsScanConfig": "obi_one.scientific.tasks.morphology_locations",
+    "MorphologyLocationsSingleConfig": "obi_one.scientific.tasks.morphology_locations",
+    "MorphologyMetricsScanConfig": "obi_one.scientific.tasks.morphology_metrics",
+    "MorphologyMetricsSingleConfig": "obi_one.scientific.tasks.morphology_metrics",
+    "SkeletonizationScanConfig": "obi_one.scientific.tasks.skeletonization",
+    "SkeletonizationSingleConfig": "obi_one.scientific.tasks.skeletonization",
+    # __init__.py aliases (class is re-exported under this name)
+    "CoupledScan": "obi_one",
+    "GridScan": "obi_one",
+    "SimulationsForm": "obi_one.scientific.unions.aliases",
```
eleftherioszisis, line 65: Do we need these aliases? I am not fond of importing from the root level as it may import other stuff too.

obi_one/scientific/unions/config_task_map.py
eleftherioszisis: > TaskType (e.g.: TaskType.circuit_extraction) is None for many tasks, so it can't be the key as it is right now.

Yeah, I was suggesting adding new task types for all tasks we have.

obi_one/core/deserializable_types.py
jplanasc: Yes, I understand your concerns. But they could have potentially been used as the `type` to serialize the "aliased" classes, so if we don't import them, it may break something?

I was being cautions, but if you're sure there's no serialization involving them, I can remove them.

obi_one/scientific/unions/config_task_map.py
jplanasc: > Yeah, I was suggesting adding new task types for all tasks we have.

It would definitely be more homogeneous. I can do it in this PR or I can create a follow-up issue to do it later, as @james-isbister was suggesting? Up to you.

eleftherioszisis: I would be ok if it is done here, given that the task maps are reorganized in this pr. However, I don't have a strong opinion.

obi_one/types.py
```diff
@@ -23,3 +24,20 @@ class TaskType(StrEnum):
...
+    electrophysiology_metrics = auto()
+    folder_compression = auto()
+    ion_channel_fitting = auto()
+    ion_channel_model_simulation = auto()
+    me_model_simulation = auto()
+    me_model_with_synapses_circuit_simulation = auto()
+    morphology_containerization = auto()
+    morphology_decontainerization = auto()
+    morphology_locations = auto()
+    morphology_metrics = auto()
```
eleftherioszisis, line 44: If these are not supported by remote task execution, would it make sense perhaps they are added to a different enum, e.g. LocalTaskType ?

Otherwise, we will allow users to specify these types in the launch endpoints even if they don't work.

tests/obi_one/core/test_scan_config_blocks.py
```diff
@@ -251,3 +251,30 @@ def test_json_round_trip_dict_blocks(self):
         json_str = config.model_dump_json()
         restored = DictBlockConfig.model_validate_json(json_str)
         assert restored.blocks["b1"].val_a == 10
+
+
+class TestAddBlockUnknownReferenceType:
```
eleftherioszisis, line 256: I usually ask LLMs to not do tests in classes (they seem to like them for some reason). It is minor but we should follow the style of the rest of the tests.

james-isbister: I liked some of the class based tests which an LLM produced for me. It made it a bit more readable for me. 

But indeed we should discuss the pros and cons and standardize at some point

eleftherioszisis: It's personal preference but I am in favor of self-contained tests + fixtures instead of a class with setup/teardown because there is state to keep in mind as well. And I am allergic to state. :D

It's debatable though if for classes function tests are more readable in the end, hence, my "personal preference" part.

obi_one/core/deserializable_types.py
eleftherioszisis: Frankly, I don't know. I would expect to have final types to serialize/deserialize in configs.  @james-isbister ?

obi_one/types.py
jplanasc: That was my concern in our chat from last Friday:
> I wanted to double-check if adding all the tasks here will have a collateral effect somewhere else, as this seems to enumerate the tasks supported for job submission but now I'd add here tasks that (I guess?) are not supported for job submission?

But from my understanding, @james-isbister said it was ok.

> @eleftherioszisis : Otherwise, we will allow users to specify these types in the launch endpoints even if they don't work.

The `TASK_MAPPINGS` holds the valid launchable tasks:

https://github.com/openbraininstitute/obi-one/blob/173fdc0d025ae95d0a81ac2243d36dffce1ca639/app/mappings.py#L26

So, would it be ok if we rely on this? Or do you want to force it by class type?

tests/obi_one/core/test_scan_config_blocks.py
jplanasc: I followed the style of each existing file: if the test file had classes, I put classes; otherwise, I didn't.

I can change it to the style you prefer, I don't mind.

obi_one/types.py
eleftherioszisis: Not without adding a field validator to check that the passed type is in the mappings here:
https://github.com/openbraininstitute/obi-one/blob/820703b0ff886ecefde8458673e4c8fbbea468f5/app/schemas/task.py#L73

Otherwise it would throw on dict lookup with an internal server error which is not very nice.

tests/obi_one/core/test_scan_config_blocks.py
james-isbister: I'd say we can leave it with what you have, and address homogenization separately at a later stage

obi_one/types.py
james-isbister: > The TASK_MAPPINGS holds the valid launchable tasks:

Yes seems ok to keep a single TaskType enum to me. As the TaskMappings defines what's runnable on the launch system as you say

obi_one/core/deserializable_types.py
jplanasc: Note those are referenced in the tests (in the `main` branch, I didn't touch it):
https://github.com/openbraininstitute/obi-one/blob/820703b0ff886ecefde8458673e4c8fbbea468f5/tests/test_data/model_dumps/grid_scan_simulations_form.json#L1-L14

obi_one/types.py
jplanasc: Ok, so I keep a single enum and I added this validation:

https://github.com/openbraininstitute/obi-one/blob/3cea0c43330f8864bf94c41a7f1587443fc4e3d4/app/schemas/task.py#L73-L87

obi_one/core/deserializable_types.py
james-isbister: This is for a work around to support duplication of old old configs in the frontend. At some point we will migrate or deprecate them, but we can't do that atm, so good if we can leave in the support for now

obi_one/core/serialization_constants.py
```diff
@@ -0,0 +1,4 @@
+"""Core constants used by scan serialization."""
+
+_SCAN_CONFIG_FILENAME = "obi_one_scan.json"
+_COORDINATE_CONFIG_FILENAME = "obi_one_coordinate.json"
```
eleftherioszisis: Minor: These are not local to this file, therefore they shouldn't have prepending underscores.

jplanasc: True, they had the underscore originally and I didn't pay attention.

Fixed.

eleftherioszisis (approved):

PR #760, opened 2026-05-07
Add validation on multiple state sharing targets

WonderPG:


jankrepl (dismissed):
LGTM:)

obi_one/scientific/library/ephys_extraction.py
```diff
@@ -252,6 +252,9 @@ def get_electrophysiology_metrics(  # noqa: PLR0912, PLR0914, PLR0915, C901
         for asset in trace_metadata.assets:
             logger.debug("Asset object: %s", asset)
             if asset.content_type == "application/nwb":
+                if asset.id is None:
```
james-isbister, line 255: I'm intrigued, were you getting errors which were returning assets without an id, in general?

app/endpoints/config_validation.py
```diff
@@ -21,9 +33,16 @@
 
 
 class SharedStatePartial(BaseModel):
-    """For validation and schema dumping."""
+    """All validatable config fields. Each is optional — validate whichever are present."""
 
-    smc_simulation_config: CircuitSimulationScanConfig | None = None
+    circuit_simulation_config: CircuitSimulationScanConfig | None = None
```
james-isbister, line 38: Was there a reason to have each ScanConfig as a separate variable. Could we use the scan config union? I don't have strong feelings though

james-isbister (commented):
Looks good, just two minor questions

obi_one/scientific/library/ephys_extraction.py
WonderPG: No but linting wouldn't go through because entitysdk has updated, and apparently it now returns the id with signature `UUID | None`. It's a pure linting change

app/endpoints/config_validation.py
WonderPG: Yeah because in the future it might be possible that multiple fields will be present at the same time. At least we make this option possible here. With only one field annotated as the union of `ScanConfig`, we force the state to have only one of these keys at the time.

james-isbister (approved):
Thanks for answering the two questions. All good on my side

PR #761, opened 2026-05-07
Update notebook simulation example to new entrypoint API

ilkilic:
Updates the E_run_small_microcircuit/run_circuit_simulation.ipynb notebook to use the new simulation entrypoint import path and explicit libnrnmech_path argument.


chr-pok (approved):
Notebook update looks good!

PR #765, opened 2026-05-11
Use number_neurons from Simulation for resource estimation

eleftherioszisis:


mgeplf (approved):

app/services/resource_estimation/circuit_simulation.py
```diff
@@ -31,11 +32,7 @@ def estimate_task_resources(
         entity_id=json_model.config_id,
         entity_type=models.Simulation,
     )
-    circuit = db_client.get_entity(
-        entity_id=config.entity_id,
-        entity_type=models.Circuit,
-    )
-    number_of_neurons = circuit.number_neurons
+    number_of_neurons = cast("int", config.number_neurons)
```
WeinaJi, line 35: Is `config.number_neurons` mandatory or optional ? If optional and unset, it may still need to be retrieved from the circuit.

eleftherioszisis: It seems to be mandatory: https://github.com/openbraininstitute/entitycore/blob/01a957a464c07ef83b9812f7eb6db29f232c04ba/app/db/model.py#L1582

WeinaJi: I thought we are using this one? 
```from entitysdk import models```
https://github.com/openbraininstitute/entitysdk/blob/main/src/entitysdk/models/simulation.py
But anyway, it seems that one will be non-optional.

PR #771, opened 2026-05-15
Add neuronal manipulation support for circuit simulations

darshanmandge:


obi_one/scientific/tasks/generate_simulations/config/base.py
```diff
@@ -87,6 +87,7 @@ class SimulationScanConfig(InfoScanConfig, abc.ABC):
         },
         SchemaKey.PROPERTY_ENDPOINTS: {
             MappedPropertiesGroup.CIRCUIT: "/mapped-circuit-properties/{circuit_id}",
+            "NeuronalManipulation": "/neuronal-manipulation-properties",
```
james-isbister: We should use a new MappedPropertiesGroup key

obi_one/scientific/tasks/generate_simulations/task/task.py
```diff
@@ -274,6 +274,12 @@ def _ensure_all_blocks_have_neuron_set_reference_if_neuron_sets_dictionary_exist
                     stimulus
                 )
```
james-isbister: What about other manipulations for single cell?

obi_one/scientific/blocks/neuronal_manipulations/neuronal_manipulations.py
```diff
@@ -227,3 +227,47 @@ def config(self, _default_population_name: str, default_node_set: str) -> list[d
...
+):
+    """Circuit variant of Full Neuron Variable Modification.
+
+    The neuron_set field is visible so the user can select which neurons
+    the manipulation targets. All logic is inherited from the parent class.
+    """
+
+    title: ClassVar[str] = "Full Neuron Variable Modification"
+
+    neuron_set: NeuronSetReference | None = Field(
```
james-isbister: I think we'll need to wait for the new neuron sets for this to be released, so it can use the logic to only allow biophysical neuron set or non virtual neuron sets i.e. [here](https://github.com/openbraininstitute/obi-one/blob/517902ea48cf9998407f469201810663935dbc94/obi_one/scientific/blocks/stimuli/stimulus.py#L112)

This doesn't block frontend development beginning though

We could re-base to the new_neuron_sets branch thought perhaps

app/endpoints/circuit_properties.py
```diff
@@ -193,3 +225,88 @@ def mapped_circuit_properties_endpoint(
...
+
+
+# --- Neuronal manipulation endpoints ---
+
+
+@router.post(
+    "/circuit/{circuit_id}/node-ids",
+    summary="Resolve neuron set to node IDs",
+    description="Returns the node IDs for a given neuron set selection in a circuit.",
+)
```
james-isbister, line 321: Could the name circuit_node_ids_endpoint be changed to neuron_set_node_ids

obi_one/scientific/blocks/neuronal_manipulations/neuronal_manipulations.py
```diff
@@ -227,3 +227,47 @@ def config(self, _default_population_name: str, default_node_set: str) -> list[d
...
+    ByNeuronMechanismVariableNeuronalManipulation,
+):
+    """Circuit variant of Full Neuron Variable Modification.
+
+    The neuron_set field is visible so the user can select which neurons
+    the manipulation targets. All logic is inherited from the parent class.
+    """
+
+    title: ClassVar[str] = "Full Neuron Variable Modification"
+
```
james-isbister, line 281: We'll need a way of signaling that the neuron_set reference and manipulation ui_elements are related, as we've linked population and property filter [here](https://github.com/openbraininstitute/obi-one/blob/517902ea48cf9998407f469201810663935dbc94/obi_one/scientific/blocks/neuron_sets_2/property.py#L98) with `property_filter_key` and `population_source_dropdown_key`

james-isbister (commented):
Looking good @darshanmandge 

I checked out the designs with @loris-olivier-obi which look good. We'll need to iron out the specification of the ui element in obi-one, so I suggest we address the third comment and write up the new ui_element in the schema-docs with screenshots so we're ready for the frontend implementation

obi_one/scientific/tasks/generate_simulations/config/base.py
darshanmandge: Done. Added NEURONAL_MANIPULATION enum member and used it.

obi_one/scientific/blocks/neuronal_manipulations/neuronal_manipulations.py
darshanmandge: alright. I think I can wait till the new_neuron_sets branch is merged.

obi_one/scientific/blocks/neuronal_manipulations/neuronal_manipulations.py
darshanmandge: Ok. We can follow the property_filter_key / population_source_dropdown_key pattern from new_neuron_sets branch once merged.

james-isbister: We will need to add this before the frontend implementation can start @darshanmandge @bilalesi 

And it would be good to update the schema docs describing what is expected

darshanmandge: Added `SchemaKey.PROPERTY_SOURCE_FIELD = "property_source_field"` to make the link between neuron_set and modification. The circuit manipulation blocks now emit:

```
"modification": {
  "property_group": "NeuronalManipulation",
  "property": "MechanismVariablesByIonChannel",
  "property_source_field": "neuron_set"
}
```
This tells the frontend: "to populate this field's options, call the endpoint from `PROPERTY_ENDPOINTS["NeuronalManipulation"]`, passing the selected neuron_set value as input.

obi_one/scientific/library/neuronal_manipulation_properties.py
```diff
@@ -0,0 +1,555 @@
...
+
+        # Stage the population's nodes.h5
+
+        snap_circuit = SnapCircuit(str(config_path))
+        nodes_h5_resolved = Path(snap_circuit.nodes[population].h5_filepath).resolve()
+        nodes_relative = str(nodes_h5_resolved.relative_to(temp_dir_path))
+        _stage_file(db_client, circuit_id, asset_id, temp_dir_path, nodes_relative)
+
+        # Resolve neuron set → node IDs
+        node_ids = neuron_set.get_neuron_ids(obi_circuit, population).tolist()
```
james-isbister: The key change for this PR from the new neuron sets is that a NeuronSet can have multiple populations - resolvable through `get_populations()`.

This affects [get_neuron_ids](https://github.com/openbraininstitute/obi-one/blob/251aed4c0afc73b4aa05b2ba0789f91cc8096ef1/obi_one/scientific/blocks/neuron_sets/base.py#L170) which:
1. No longer accepts a population parameter (as no longer needed)
2. Now returns a dictionary of neuron ids by population

The logic in the here and in the endpoint will need to be adjusted to take this into account

james-isbister (commented):
I merged main and updated the schema of the neuronal maniupulations blocks to only accept biophsyical neuron sets (we didn't have this specificity before).

An additional change will also be required following the merge. See comment below. Let me know if there are any questions

obi_one/scientific/library/neuronal_manipulation_properties.py
darshanmandge: Updated the code in the latest commits.

app/endpoints/circuit_properties.py
```diff
@@ -264,3 +294,79 @@ def mapped_circuit_properties_endpoint(
...
+                "detail": f"Internal error resolving node IDs for circuit {circuit_id}.",
+            },
+        ) from err
+    return NodeIdsResponse(node_ids_per_population=ids_per_population)
+
+
+@router.post(
+    "/neuronal-manipulation-properties",
+    summary="Neuronal manipulation properties",
+    description="Returns mechanism variables for neuronal manipulation blocks. "
```
james-isbister: This description should describe that endpoint returns the intersection of mechanism variables, in case other people or the agent want to use it in other contexts

obi_one/scientific/library/neuronal_manipulation_properties.py
```diff
@@ -0,0 +1,476 @@
...
+        return {}
+
+    groups = list(emodel_groups.values())
+
+    # Filter out groups that failed to fetch mechanism variables
+    groups_with_vars = [g for g in groups if g["mechanism_variables_by_ion_channel"]]
+    if not groups_with_vars:
+        return {}
+
+    # Single emodel: return its variables directly (no intersection needed)
```
ilkilic, line 318: If the other e-model fetches failed, `groups_with_vars` may contain only one group here. In that case, we return all its variables as common even though they were not verified for the other selected e-models.

darshanmandge: Yes, it is intentional. Variable fetch failures are logged with warnings. If the derivations exist, it will show all the variables of the single e-model.

ilkilic: Alright, that sounds reasonable if failed fetches are logged with a warning.

app/endpoints/circuit_properties.py
```diff
@@ -274,3 +304,79 @@ def mapped_circuit_properties_endpoint(
...
+# --- Neuronal manipulation endpoints ---
+
+
+@router.post(
+    "/circuit/{circuit_id}/node-ids",
+    summary="Resolve neuron set to node IDs",
+    description="Returns the node IDs for a given neuron set selection in a circuit.",
+)
+def neuron_set_node_ids(
+    circuit_id: str,
```
ilkilic, line 323: Could `circuit_id` be typed as UUID here?

app/endpoints/circuit_properties.py
```diff
@@ -274,3 +304,79 @@ def mapped_circuit_properties_endpoint(
...
+    db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
+) -> NodeIdsResponse:
+    try:
+        ids_per_population = get_circuit_node_ids(
+            db_client=db_client,
+            circuit_id=circuit_id,
+            neuron_set=request.neuron_set,
+        )
+    except ValueError as err:
+        raise HTTPException(status_code=HTTPStatus.BAD_REQUEST, detail=str(err)) from err
```
ilkilic, line 334: Are all ValueErrors here really client errors? _get_circuit_asset() also raises ValueError when the circuit asset is missing or invalid, which would return 400 even though the request itself may be valid. Could we use more specific exceptions?

app/endpoints/circuit_properties.py
```diff
@@ -274,3 +304,79 @@ def mapped_circuit_properties_endpoint(
...
+    summary="Neuronal manipulation properties",
+    description="Returns mechanism variables for neuronal manipulation blocks. "
+    "Supports both MEModel (single cell) and Circuit (multi-cell) entities.",
+)
+def neuronal_manipulation_properties_endpoint(
+    request: NeuronalManipulationPropertiesRequest,
+    db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
+) -> dict:
+    # Try MEModel path first
+    memodel_result = try_get_mechanism_variables(
```
ilkilic: One concern here: try_get_mechanism_variables() returns None both when the entity is not an MEModel and when it is an MEModel but retrieving or parsing its mechanism variables fails. Since this return value is used to choose between the MEModel and Circuit paths, the second case would also fall through to the Circuit path and could hide the original error.

app/endpoints/circuit_properties.py
```diff
@@ -274,3 +304,79 @@ def mapped_circuit_properties_endpoint(
...
+    return NodeIdsResponse(node_ids_per_population=ids_per_population)
+
+
+@router.post(
+    "/neuronal-manipulation-properties",
+    summary="Neuronal manipulation properties",
+    description="Returns mechanism variables for neuronal manipulation blocks. "
+    "Supports both MEModel (single cell) and Circuit (multi-cell) entities.",
+)
+def neuronal_manipulation_properties_endpoint(
```
ilkilic: Would it be possible to expose separate MEModel and Circuit endpoints here? I think it would avoid having to infer the entity type by trying one path and then falling back to the other.

app/endpoints/circuit_properties.py
darshanmandge: Other endpoints take circuit_id as `str`. We will have to change each one. Let me know if I should address it in this PR.

ilkilic: Alright, I think this can be addressed later.

app/endpoints/circuit_properties.py
darshanmandge: Thanks for the suggestion. Currently, obi-one defines a limited set of error types (see obi_one/core/exception.py), and the neuronal manipulation library functions raise ValueError for all failure cases. Introducing more specific exception classes would be a broader change that updates multiple files, endpoints, and tests. I think this, too, can be addressed in a separate PR to keep this one focused, if it's fine with you?

ilkilic: Sure! Thanks

app/endpoints/circuit_properties.py
darshanmandge: Thanks for the suggestion! I have separated the MEModel and Circuit endpoints. Could you have a look at it again?

app/endpoints/circuit_properties.py
darshanmandge: Since the endpoints are now separate, we no longer have this issue.

ilkilic (dismissed):
Thanks @darshanmandge ! LGTM

app/endpoints/circuit_properties.py
darshanmandge: @james-isbister I renamed the function and endpoint to use the form: `neuron_set_node_ids`. I also added a cell in the endpoints notebook to test it.

james-isbister (dismissed):
LGTM. Thanks for the changes @darshanmandge

app/endpoints/circuit_properties.py
darshanmandge: Done in the latest commit.

james-isbister (approved):
LGTM!

PR #774, opened 2026-05-15
Add support for extra morphology diagnostics

dkeller9:
**Summary**
When a morphology file fails validation, the API now runs a full suite of
NeuroM quality checks against the file and appends the results to the error
response. Callers receive a structured quality_checks object alongside the
existing error detail, listing exactly which checks passed and which failed.

**Motivation**
Validation errors previously returned only a generic message (e.g.
"Morphology validation failed: <SomaNonConform ...>"), giving no actionable
information about what was structurally wrong with the file. Users had no way
to know whether the issue was missing neurite types, zero-length segments, fat
ends, or something else entirely. The quality check output makes the error
immediately actionable.

**Changes**
app/services/morphology.py

Added _QUALITY_CHECK_CONFIG — a module-level constant defining the 12
standard NeuroM checks and their tolerance options, extracted directly from
the reference notebook (conduct_morphology_quality_checks.ipynb).
Added _quality_check_runner — a single CheckRunner instance cached at
module level, consistent with other module-level singletons in the file.
Added run_quality_checks(file_path: Path) -> dict[str, Any] — runs the
checks against a morphology file and returns ran_to_completion,
failed_checks, and passed_checks. Exceptions inside the function are
caught and logged as warnings so a secondary diagnostic failure can never
mask the original validation error.

app/endpoints/morphology_validation.py

Wrapped the three validation/conversion steps (load_morphio_morphology,
validate_soma_diameter, convert_morphology) in a try/except HTTPException block.
On failure, run_quality_checks is called against the saved input file
(still on disk at that point) and its results are merged into the
HTTPException detail under a quality_checks key before re-raising with
the original status code.
Added Any to the from typing import line; annotated detail explicitly
as dict[str, Any] to satisfy the type checker (was inferred as
dict[str, str]).

**Error response shape**
A 422 response now looks like:
json{
  "detail": {
    "code": "INVALID_REQUEST",
    "detail": "Morphology validation failed: ...",
    "quality_checks": {
      "ran_to_completion": true,
      "failed_checks": ["Has no jumps", "Has no fat ends"],
      "passed_checks": [
        "Has axon", "Has basal dendrite", "Has apical dendrite",
        "Has nonzero soma radius", "Has all nonzero neurite radii",
        "Has all nonzero section lengths", "Has all nonzero segment lengths",
        "Has no flat neurites", "Has no narrow start", "Has no dangling branch"
      ]
    }
  }
}


ilkilic (approved):
LGTM

Issue #699, opened 2026-04-17
Refactor/unify circuit registration & upload functionality


chr-pok:
## Description
Refactor/unify circuit registration & upload functionality to be used by several obi-one tasks. For internal use only (w/o circuit validation). Including automatic main image generation (unless provided by the user). Incl. refactored functions for additional/optional assets generation.

## Acceptance criteria
- [ ] Refactored circuit registration & upload functionality

chr-pok:
Different code locations to be refactored/unified:
- **[circuits_hardcoded](https://github.com/openbraininstitute/circuits_hardcoded) repo:** [register_circuit.ipynb](https://github.com/openbraininstitute/circuits_hardcoded/blob/main/register_circuit.ipynb) + [register_circuit_helpers.py](https://github.com/openbraininstitute/circuits_hardcoded/blob/main/register_circuit_helpers.py)
- **[obi-one](https://github.com/openbraininstitute/obi-one) repo:**
  - `_create_circuit_entity()` and `_generate_additional_circuit_assets()` in [tasks/circuit_extraction.py](https://github.com/openbraininstitute/obi-one/blob/main/obi_one/scientific/tasks/circuit_extraction.py)
  - `register_output()` in [tasks/em_synapse_mapping/register.py](https://github.com/openbraininstitute/obi-one/blob/main/obi_one/scientific/tasks/em_synapse_mapping/register.py)

chr-pok:
Refactoring started in the branch [circuit_registration](https://github.com/openbraininstitute/obi-one/tree/circuit_registration).

PR #776, opened 2026-05-18
Circuit registration refactoring/unification

chr-pok:
Refactored and unified the circuit registration and upload functionality.

**Summary:**
- Unified circuit registration into a single shared utilities module to be used stand-alone or by obi-one tasks
- Integrated unified circuit registration consistently into circuit extraction and EM synapse mapping tasks (thereby removed duplicated registration code from both tasks)
- Unified circuit registration in EM synapse mapping tasks:
  - Compressed circuit asset contains now full circuit folder (before: selected files only)
  - mapping_stats.png also contained in circuit folder --> Could be a separate asset, if needed
  - node_sets.json added, since required by SNAP (before: was missing)
  - Has now an overview image from a template (before: "Error loading thumbnail")
- Scale, neuron/synapse/connection counts, and circuit properties (morphologies, point neurons, e-models, spines) are now determined automatically from the circuit files and no longer passed manually
- The SONATA circuit folder asset is uploaded automatically during registration
- Additional assets (compressed circuit, connectivity matrices, connectivity plots, overview figures) are generated automatically and uploaded as well
- Derivation links, contributions, and publication links are registered as well
- A user-facing function for notebook workflows resolves all entity references from string names (subject, brain region, license, etc.)
- Added an example notebook demonstrating the user-facing registration workflow
- Added comprehensive unit tests for the new module

**Important:**
- Added coordinate directory option "NONE" (required to write output of sub-tasks directly in output folder w/o subfolders)
- Removed AWS S3 path support (since does not work with automated circuit size/property retrieval)
- Renamed `check_only` to `dry_run` (which is more commonly used)

**Example outputs for validation:**
- Circuit entity registered using the new example notebook: [Circuit ID 123b820c-bd53-4100-a0b4-0a218cf51619](https://staging.openbraininstitute.org/app/entity/123b820c-bd53-4100-a0b4-0a218cf51619)
- Circuit entities registered by EM synapse mapping task:
  - Before refactoring/unification: [Circuit ID 4fae051c-f134-4b64-bd63-81f8a1e00a75](https://staging.openbraininstitute.org/app/entity/4fae051c-f134-4b64-bd63-81f8a1e00a75)
  - After refactoring/unification: [Circuit ID 7d812b76-126e-4aa5-9fe0-f1b8133c6c95](https://staging.openbraininstitute.org/app/entity/7d812b76-126e-4aa5-9fe0-f1b8133c6c95)

Related issue: https://github.com/openbraininstitute/obi-one/issues/699


james-isbister (commented):
Looks good. I haven't yet tested it but my initial reaction is the following:
- Could we add to the example notebook, examples of how to register contributors, publications etc?
- It maybe seems a little strange that we specify contributors, publications etc as dictionaries rather than the database ids. Perhaps there is a reason? Perhaps we want to support both?
- Could the main file utils/circuit_extraction.py be split up into several files within a new directory (perhaps around 5 or 6, so we're not more than low hundreds of lines per file). For a file with many functions like this, I find it hard to parse.

obi_one/scientific/tasks/circuit_extraction/task.py
```diff
@@ -225,11 +208,11 @@ def _resolve_circuit(self, *, db_client: Client, entity_cache: bool) -> None:
             msg = "Failed to resolve circuit!"
```
james-isbister: Worth noting that we now have three very similar versions of _resolve_circuit:
- Here
- [Create recording array](https://github.com/openbraininstitute/obi-one/blob/16c242c372c0ceebb0828bc647ea3caadca3682f/obi_one/scientific/tasks/create_recording_array/create_recording_array.py#L152-L181)
- [Generate simulations](https://github.com/openbraininstitute/obi-one/blob/16c242c372c0ceebb0828bc647ea3caadca3682f/obi_one/scientific/tasks/generate_simulations/task/task.py#L119)

Would be good to unify at some point

obi_one/utils/circuit_registration/register.py
```diff
@@ -0,0 +1,297 @@
...
+    Args:
+        client: The entitycore SDK client.
+        circuit_path: Path to circuit_config.json (or the folder containing it).
+        name: Circuit name.
+        description: Circuit description.
+        build_category: Build category (computational_model, em_reconstruction).
+        brain_region: Resolved brain region entity.
+        subject: Resolved subject entity.
+        target_simulator: Target simulator for the circuit (optional,
+            defaults to NEURON if not provided).
```
james-isbister, line 103: Maybe better to make it non-optional as the simulation generation configuration which is displayed in the UI for the circuit depends on this property.

Perhaps we should also do a validation that the target_simulator corresponds with the corresponding SONATA target_simulator in the circuit_config

Keeping this necessary duplication was of concern to some of the engineers, so here we have the opportunity to catch mismatches

james-isbister: It would of course be set to NEURON in the em_synapse_mapping task

examples/L_utils/register_circuit.ipynb
```diff
@@ -0,0 +1,327 @@
...
+   "source": [
+    "circuit_metadata = {\n",
+    "    # Name of the circuit\n",
+    "    \"name\": \"EXAMPLE_01__N_10__top_nodes_dim6\",\n",
+    "\n",
+    "    # Description of the circuit\n",
+    "    \"description\": \"Example circuit with 10 neurons\",\n",
+    "    \n",
+    "    # Circuit build category (computational_model, em_reconstruction)\n",
+    "    \"build_category\": \"computational_model\",\n",
```
james-isbister: Perhaps the values of build_category and target_simulator in this dictionary could use the EntitySDK types so that it is clear to anyone using this, what is expected for these fields

james-isbister (commented):
The notebook works well for me. Just some very small comments

Also, we might also add to this error message [here](https://github.com/openbraininstitute/obi-one/blob/a08ced6e2d3b913f57c8942986825f3a0bad4e6e/obi_one/scientific/tasks/basic_connectivity_plots.py#L385) to note that the plotting can also be installed with`make install-connectivity` in development

obi_one/utils/circuit_registration/register.py
chr-pok: > Perhaps we should also do a validation that the target_simulator corresponds with the corresponding SONATA target_simulator in the circuit_config

I am not aware that the circuit_config has such field. Has it been added recently?

> It would of course be set to NEURON in the em_synapse_mapping task

Hmm, I don't understand, why should it be set to NEURON? The output circuit is not simulatable, so we should not set any target simulator I believe. This is the main why I had to leave it optional in the first place.

chr-pok: Would be even better to introduce a "None" type in entitycore for such cases.

james-isbister: Maybe it's recent I'm not sure, but it is in the [specs](https://sonata-extension.readthedocs.io/en/latest/sonata_config.html) if you search "target_simulator"

Ah yes, you're right about the em synapse mapping task

Indeed, maybe a None type in entitycore is useful for anatomical circuits. Or perhaps "nullable" could be set to True [here](https://github.com/openbraininstitute/entitycore/blob/01236cecd7b515f666ce984c0593a2ae176a21cf/app/db/model.py#L1827)

cc. @eleftherioszisis

chr-pok: > Maybe it's recent I'm not sure, but it is in the [specs](https://sonata-extension.readthedocs.io/en/latest/sonata_config.html) if you search "target_simulator"

Thanks! I was not aware. Seems to be optional, but I will add a consistency check in case the field exists in the config.

chr-pok: I added a consistency check in case the target simulator is specified by the user and in the circuit config, or using it from the config if not specified by the user.

examples/L_utils/register_circuit.ipynb
chr-pok: Good idea, I added the values from the corresponding enums.

obi_one/scientific/tasks/circuit_extraction/task.py
chr-pok: Yes indeed. I unified two of them which were really identical, the unified function is now in db_sdk utilities. The third one (generate simulations) is more complex and has some additional logic and there is no clean way to unify, so this should be tackled separately if needed.

james-isbister (approved):
Very nice. All comments addressed. Now its time to add some circuits to the platform!

PR #784, opened 2026-05-20
Reorganize + refactor existing Simulation Generation ScanConfigs 

james-isbister:
Before refactoring the SimulationGenerationTask to work for Learning Engine, Neuron + Brian2, I needed to refactor and re-organize the existing Brain2 and Neuron ScanConfigs - Adding the LearningEngine ones should then be trivial

I have added a [readme](https://github.com/openbraininstitute/obi-one/tree/move_simulation_generation_files/obi_one/scientific/tasks/generate_simulations/config) which describes the approach and choice made

The inheritance provides a logical re-use of components whilst maintaining consistency across simulators and scales. If we find it difficult to manage we can always write the concrete classes without inheritance later, but I would find this much harder to manage and maintain consistency across different configs, so would like to try with this approach first

Please note that the details of the Brian2 configs will be finalised in a separate PR - this is just to get the organization with what we currently have


obi_one/scientific/tasks/generate_simulations/config/README.md
```diff
@@ -0,0 +1,201 @@
...
+│   └── me_model.py    MEModelBaseSimulationScanConfig    (adds ME-model `circuit`)
+├── brian2/
+│   ├── brian2_base.py    Brian2SimulationScanConfig      (Brian2 Initialize fields)
+│   └── brian2_circuit.py Brian2CircuitSimulationScanConfig + …SingleConfig
+└── neuron/
+    ├── neuron_base.py                Neuron equivalent of brian2_base
+    ├── neuron_circuit.py             CircuitSimulationScanConfig + …SingleConfig
+    ├── neuron_me_model.py            MEModelSimulationScanConfig + …SingleConfig
+    ├── neuron_me_model_with_synapses.py
+    └── neuron_ion_channel_models.py
```
eleftherioszisis: There is a conceptual separation here that seems strange to me. Why is `neuron` at the same level as `brian2`? In one case we separate by target simulator and in the other case by cell family? These two are orthogonal in my mind. wdyt?

eleftherioszisis: Scratch that.. That is NEURON isn't it? #facepalm

eleftherioszisis: Are the prefixes in the modules necessary?

james-isbister: Haha, NEURON indeed.

Yep, I'd prefer not to have the prefixes but the reason I kept them was to make working across multiple files a bit easier. Otherwise there could end up being 4,5,6... files same name, which wouldn't be fun when all 4 files are open. When things stabalise and we're not editing these files so often we could remove the prefixes though perhaps

eleftherioszisis: May I divert the question to a broader one?

Should it be the case that you need to import all of them at once? GenerateSimulationTask is mixing all the responsibilities for all configs in one place. Would it be cleaner if there were in a similar fashion dedicated modules with tasks that imported only the dedicated types for the specific target simulator / circuit type classes?

The fact that there is so much branching in GenerateSimulationTask is a telltale that class is having an identity crisis :D

james-isbister: Haha, the task will be the subject of one of the next PRs in the series. Definitely needs tidying up in some way. This is just the first step reorganizing the configs

obi_one/scientific/tasks/generate_simulations/config/circuit.py
```diff
@@ -0,0 +1,60 @@
...
+)
+from obi_one.scientific.unions.unions_neuron_sets import (
+    NeuronSetReference,
+    SimulationNeuronSetUnion,
+)
+
+CircuitDiscriminator = Annotated[Circuit | CircuitFromID, Field(discriminator="type")]
+
+
+class CircuitBaseSimulationScanConfig(BaseSimulationScanConfig, abc.ABC):
```
eleftherioszisis: Does it worth it adding the whole MRO complexity just to avoid adding a circuit and node_set in neuron_circuit.py? In brian2 you override everything anyway.

obi_one/core/block.py
```diff
@@ -23,6 +29,27 @@ class Block(OBIBaseModel, extra="forbid"):
...
+    @staticmethod
+    def order_schema_properties(schema: dict) -> None:
+        properties = schema.get("properties")
+        if not properties:
+            return
+
+        def priority(prop_schema: dict) -> float:
+            return prop_schema.get("order_priority", 0) or 0
+
+        schema["properties"] = dict(sorted(properties.items(), key=lambda item: -priority(item[1])))
```
eleftherioszisis: Move this function to `obi_one/utils/pydantic.py` and add/move tests that check the correct behavior.

obi_one/scientific/tasks/generate_simulations/config/brian2/brian2_base.py
```diff
@@ -0,0 +1,86 @@
...
+
+from pydantic import Field, NonNegativeFloat, PositiveFloat
+
+from obi_one.core.schema import SchemaKey, UIElement
+from obi_one.core.units import Units
+from obi_one.scientific.library.constants import (
+    _DEFAULT_SIMULATION_LENGTH_MILLISECONDS,
+    _MAX_SIMULATION_LENGTH_MILLISECONDS,
+    _MIN_SIMULATION_LENGTH_MILLISECONDS,
+    _SIMULATION_TIMESTEP_MILLISECONDS,
```
eleftherioszisis: Prepending underscores should only be used for local to the module constants.

obi_one/scientific/tasks/generate_simulations/config/neuron/neuron_base.py
```diff
@@ -0,0 +1,101 @@
...
+
+from pydantic import Field, NonNegativeFloat, PositiveFloat
+
+from obi_one.core.schema import SchemaKey, UIElement
+from obi_one.core.units import Units
+from obi_one.scientific.library.constants import (
+    _DEFAULT_SIMULATION_LENGTH_MILLISECONDS,
+    _MAX_SIMULATION_LENGTH_MILLISECONDS,
+    _MIN_SIMULATION_LENGTH_MILLISECONDS,
+    _SIMULATION_TIMESTEP_MILLISECONDS,
```
eleftherioszisis: Same here. Remove underscore.

obi_one/scientific/tasks/generate_simulations/task/task.py
```diff
@@ -86,7 +87,7 @@ def _initialize_sonata_simulation_config(self) -> dict:  # ty:ignore[invalid-ret
             self._sonata_config["target_simulator"] = BRIAN2_TARGET_SIMULATOR
 
         self._sonata_config["run"] = {}
-        self._sonata_config["run"]["dt"] = self.config.initialize.timestep
+        self._sonata_config["run"]["dt"] = self.config.initialize.timestep  # ty:ignore[unresolved-attribute]
```
eleftherioszisis: Is this ty error expected?

obi_one/utils/pydantic.py
```diff
@@ -0,0 +1,12 @@
...
+
+def order_schema_properties(schema: dict) -> None:
+    properties = schema.get("properties")
+    if not properties:
+        return
+
+    def priority(prop_schema: dict) -> float:
+        return prop_schema.get(SchemaKey.PARAMETER_ORDER_PRIORITY, 0) or 0
+
+    schema["properties"] = dict(sorted(properties.items(), key=lambda item: -priority(item[1])))
```
eleftherioszisis: Is would be better if the function had no side-effects, i.e. returned a copy of the schema instead of mutating it in-place.

obi_one/utils/pydantic.py
```diff
@@ -0,0 +1,19 @@
...
+def order_schema_properties(schema: dict) -> dict:
+    """Return a copy of `schema` with its `properties` ordered by priority.
+
+    Properties are sorted by descending `SchemaKey.PARAMETER_ORDER_PRIORITY`
+    (absent or falsy priorities default to 0, ties keep their original order).
+    The input `schema` is not modified.
+    """
+    properties = schema.get("properties")
+    if not properties:
+        return dict(schema)
```
eleftherioszisis: In this case, it is ok to return the initial input, given that you do not modify it, no?

james-isbister: Hm, I'm not keen that it would return the original in one case and a copy in the other. Could lead to unexpected results, no?

eleftherioszisis: If the function was used in different places, maybe, but I doubt it in this case.

eleftherioszisis (approved):
Looks good. I assume you have tested the GUI generation and that the sorting functionality generates the desired results.

PR #785, opened 2026-05-21
Neurodamus-fueled beta simulations

eleftherioszisis:
## Summary

Adds Neurodamus as a simulation backend and introduces a unified simulation execution workflow for running staged simulations and registering their outputs.

## What's new
New simulation_execution task framework with a shared base (SimulationExecutionTask) and specialized tasks for:
* Circuit simulation
* Ion channel model simulation
* Single neuron simulation
* Single neuron synaptome simulation

Mechanism compilation via neurodamus-compile-mods (neurodamus), MPI launch with the NEURON special binary, and result collection

Backend-aware simulation staging and parameter extraction (SimulationBackend.neurodamus vs bluecellulab)

Launch System integration: task types, config→task mappings, and Neurodamus Docker image for machine/cluster execution

## Refactor
* Replaces the old ion-channel simulation execution task with the new unified simulation execution module
* Extends the simulation library (process, staging, schemas) to support both Bluecellulab and Neurodamus backends


obi_one/scientific/library/simulation/neuron/process.py
james-isbister, line 1: There is now library/simulation/brain2 so maybe these files could go in library/simulation/NEURON

obi_one/scientific/library/simulation/neuron/staging.py
```diff
@@ -1,16 +1,42 @@
...
+from obi_one.types import SimulationBackend
 from obi_one.utils.io import load_json
 
+if TYPE_CHECKING:
+    from entitysdk.models import MEModel
+
 L = logging.getLogger(__name__)
 
 
+def stage_circuit(*, client: Client, model: models.Circuit, output_dir: Path) -> Circuit:
```
james-isbister, line 30: There are already stage_circuit functions for CircuitFromID, MEModelFromID and CircuitFromIonChannelModels. CircuitFromID.stage_circuit is used in circuit extraction and simulation generation. Not saying its better, just I think it would be good to unify across tasks with the a single well tested implementation either here or as a separate PR to avoid duplication

obi_one/scientific/tasks/simulation_execution/__init__.py
james-isbister, line 1: Maybe this directory could be simulation_execution/NEURON to be specific

james-isbister (commented):
LGTM, just a few very small comments

obi_one/scientific/library/simulation/neuron/process.py
eleftherioszisis: Sure, however it is surprising that there is a task in library/simulation/brian2

james-isbister: True, I'm working on a PR actually where I could also move the brain2 script

obi_one/scientific/library/simulation/neuron/staging.py
eleftherioszisis: I prefer using the entitysdk `stage_circuit` function directly because its behavior is explicit, well understood, and already tested. CircuitFromID staging is stateful and also performs unrelated operations, such as skipping staging if the target directory already exists and caching, neither of which is applicable to the remote task. Reusing it here could introduce unnecessary complexity and subtle bugs that are easy to avoid by calling `stage_circuit` directly.

obi_one/scientific/tasks/simulation_execution/__init__.py
eleftherioszisis: done

obi_one/scientific/library/simulation/neuron/staging.py
james-isbister: entitysdk.stage_circuit we should definitely use - CircuitFromID.stage_circuit also uses this - although it is from a while back and has some legacy stuff which is no longer used and complicates it, certainly. I'm not saying we should use the exact version of CircuitFromID.stage_circuit, but eventually it would be good to have a single wrapper of entitysdk.stage_circuit in obi-one rather than one for each task. For now, we can keep things as is though I think

tests/obi_one/scientific/library/simulation/neuron/test_entrypoint.py
james-isbister, line 1: Could these test files also go in library/simulation/neuron?

eleftherioszisis: Good catch. Yeah, will move them, thanks

tests/obi_one/scientific/tasks/simulation_execution/neuron/test_ion_channel_simulation_execution.py
GianlucaFicarelli, line 1: Since this file is moved, `tests/obi_one/scientific/tasks/ion_channel_model_simulation_execution/__init__.py` can be removed.

GianlucaFicarelli (dismissed):
LGTM

GianlucaFicarelli (approved):

PR #786, opened 2026-05-26
Multi-part circuit upload

chr-pok:
Enabled multi-part directory upload for circuit and matrix folder assets.

Required for: https://github.com/openbraininstitute/prod-build-circuit/issues/28

### DIFF
--- a/obi_one/utils/db_sdk.py
+++ b/obi_one/utils/db_sdk.py
@@ -10,7 +10,6 @@ from entitysdk.exception import EntitySDKError
 from entitysdk.models import Entity, TaskActivity, TaskConfig
 from entitysdk.models.activity import Activity
 from entitysdk.models.asset import Asset
-from entitysdk.schemas.asset import MultipartDirectoryUploadTransferConfig
 from entitysdk.types import ActivityStatus, AssetLabel, ContentType, ExecutorType, TaskActivityType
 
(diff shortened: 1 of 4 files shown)


obi_one/utils/db_sdk.py
```diff
@@ -367,12 +370,14 @@ def add_connectivity_matrix_asset(
     L.info(f"{len(matrix_files)} files in '{matrix_dir}'")
 
     # Upload directory asset
+    transfer_config = MultipartDirectoryUploadTransferConfig()
```
mgeplf: if you're not changing the default, you don't need this

chr-pok: Oh, maybe I misunderstood, but this was exactly what I asked on the other ticket. So if I don't provide `transfer_config=...` it will still use large file upload (whenever I use entitysdk >= 0.16.0)?

mgeplf: I thought you were asking if the signature had changed - it has, but just to add that parameter; otherwise it's the same call, but under the hood it does the big file upload

chr-pok: Ok, thanks for clarifying. I reverted my changes, just kept the version update.

GianlucaFicarelli (approved):

PR #797, opened 2026-06-03
LE Generation + base sim config refactor

james-isbister:
- Add Learning Engine ScanConfig, endpoint etc
- Move initialisation of sonata sim config into config functions
- Reshuffle several properties in config inheritance hierarchy
- Add brian2_point to set of acceptable simulation targets (this is going to change with the new neuron sets, so just a temporary solution)

Note, this is meant only as a first step of refactoring the GenerateSimulationTask


obi_one/scientific/tasks/generate_simulations/config/brian2/brian2_base.py
```diff
@@ -37,48 +37,9 @@ class Brian2SimulationScanConfig(BaseSimulationScanConfig, abc.ABC):
...
-            title="Random Seed",
-            description="Random seed for the simulation.",
-            json_schema_extra={
-                SchemaKey.UI_ELEMENT: UIElement.INT_PARAMETER_SWEEP,
-            },
-        )
+    def base_sonata_config(self, sonata_config: dict | None = None) -> dict:
+        """Returns the base SONATA configuration for the simulation campaign."""
+        sonata_config = super().base_sonata_config(sonata_config)
+        return sonata_config
```
eleftherioszisis: Doe this override have a purpose? It seems it only calls the parent method the exact same way.

obi_one/scientific/tasks/generate_simulations/config/learning_engine/le_base.py
```diff
@@ -0,0 +1,33 @@
...
+            json_schema_extra={
+                SchemaKey.UI_ELEMENT: UIElement.INT_PARAMETER_SWEEP,
+                SchemaKey.UI_HIDDEN: True,
+            },
+        )
+
+    def base_sonata_config(self, sonata_config: dict | None = None) -> dict:
+        """Returns the base SONATA configuration for the simulation campaign."""
+        sonata_config = super().base_sonata_config(sonata_config)
+        return sonata_config
```
eleftherioszisis: Same here. Why not remove this and use the parent method?

obi_one/scientific/tasks/generate_simulations/config/neuron/neuron_base.py
```diff
@@ -24,6 +22,10 @@
 class NeuronSimulationScanConfig(BaseSimulationScanConfig, abc.ABC):
     """Abstract base class for neuron-based simulation scan configurations."""
 
+    _target_simulator: ClassVar[SimulatorType] = SimulatorType.NEURON
+    _spike_location: ClassVar[str] = "soma"
```
eleftherioszisis: Minor: Does this "soma" belong to a controlled vocabulary that we could either use as an Enum or a list of allowed values?

mgeplf:
> Note I removed the spike stimuli as @mgeplf mentioned that these are only supported between biophys populations atm. 

I should be able to add it in a reasonable amount of time.
Do you know what sort of virtual nodes you will need?
I guess we could make a task for the next sprint?

> Better to add the virtual neuron support in learning engine than find a work around for representing that in obi-one I think - we could do it with the new neuron sets but ideally not needed

Agreed.

> one of you could try with one of the sim configs generated by the notebook?

I should be able to do that, I will get back to you.

eleftherioszisis (approved):

PR #800, opened 2026-06-05
Make target simulator non-optional for circuit registration

chr-pok:


james-isbister (approved):
Thanks for the change. Looks good!

PR #803, opened 2026-06-05
update virtual lab compute_cell extraction

bilalesi:


chr-pok (approved):
Thanks! LGTM and works again with this update!

PR #804, opened 2026-06-08
copy `_write_indexes` from brainbuilder

mgeplf:


chr-pok (commented):
Thanks, LGTM! But @MWolfR may want to have a look as well, I think he originally implemented that code.

MWolfR (approved):

PR #806, opened 2026-06-08
Dispatch LOD mesh generation as an asynchronous launch-system task

dkeller9:
Refactors the mesh registration endpoint to submit LOD generation as a background task via the launch-system, matching the pattern established by `circuit_extraction` and `morphology_skeletonization`.

### Motivation

Previously, the mesh registration endpoint ran the full pipeline — OBJ upload, ultraliser LOD generation, and LOD directory upload — synchronously in a single request. This is slow and fragile for large meshes. LOD generation is now dispatched as an async task job, returning immediately with a `task_job_id` and `status: "pending"`.

### Changes

**`endpoints/mesh_registration.py`**
- Removed synchronous `_run_registration_pipeline`, `_generate_lods`, and `_register_lod_directory` helpers
- OBJ asset upload still happens synchronously (fast)
- Creates a `TaskConfig` entity encoding `entity_id` and `obj_asset_id`
- Submits a `mesh_lod_generation` task job through the existing accounting + launch-system infrastructure
- Response schema updated: `lod_directory_asset_id` and `lod_file_count` replaced by `task_job_id` and `status: "pending"`
- the endpoint also accepts glb files for registration, and converts obj files to glb

**`mappings.py`**
- Added `TaskType.mesh_lod_generation` to `TASK_DEFINITIONS` with `private_packages=True`, 4 cores / 8 GB / 1-hour timelimit

**`services/accounting.py`**
- Added `case TaskType.mesh_lod_generation` in `_evaluate_accounting_parameters`, returning `count=1` and `ServiceSubtype.NEURON_MESH_SKELETONIZATION`

**`obi_one/scientific/tasks/mesh_lod_generation/`** — new task module
- `config.py` — `MeshLodGenerationScanConfig` with `entity_id` and `obj_asset_id` fields
- `estimate.py` — `estimate_mesh_lod_generation_count` returning `1`
- `task.py` — `run_mesh_lod_generation`: downloads OBJ, runs `ultraliser.LODGenerator.generate_web_lods`, uploads resulting directory block
- `__init__.py` — exports `MeshLodGenerationScanConfig`

### Dependencies
- Requires the entitysdk `_server_schemas.py` update adding `mesh_lod_generation__config` / `mesh_lod_generation__execution` to `TaskConfigType` / `TaskActivityType`


app/mappings.py
```diff
@@ -64,7 +64,7 @@
         config_type=TaskConfigType.mesh_lod_generation__config,
         activity_type=TaskActivityType.mesh_lod_generation__execution,
         code=PythonRepositoryCode(
-            location=settings.OBI_ONE_REPO,
+            location="obi-one @ git+https://github.com/openbraininstitute/obi-one.git@mesh_registration_endpoint",
```
danifr: I don't think pointing the repo to a feature branch is the best idea in the long run

dkeller9: This is just a temporary measure for additional testing.

obi_one/scientific/tasks/mesh_lod_generation/task.py
```diff
@@ -0,0 +1,105 @@
...
+    )
+    asset_id = str(result.id) if hasattr(result, "id") else str(result[0].id)
+    return asset_id
+
+
+class MeshLODGenerationTask(Task):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    config: MeshLodGenerationScanConfig
+    client: entitysdk.Client
```
jplanasc: @dkeller9 are you sure this is going to work? I thought the client needed to be passed as a parameter in `execute`

obi_one/scientific/tasks/mesh_lod_generation/config.py
```diff
@@ -0,0 +1,32 @@
...
+        description="The unique identifier of the target EMCellMesh entity.",
+    )
+    obj_asset_id: UUID = Field(
+        ...,
+        description="The specific asset ID corresponding to the source OBJ payload data.",
+    )
+
+
+# Alias single configuration mapping for task registry deserialization matching
+MeshLodGenerationScanConfig = MeshLodGenerationSingleConfig
```
jplanasc: @dkeller9 why are you doing this alias?

obi_one/scientific/tasks/mesh_lod_generation/task.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def execute(self) -> str:
```
jplanasc: I think the minimal params for `execute` are:
`def execute(self, *, db_client, entity_cache, execution_activity_id)`

Have you tried to execute the task? Does it work?

app/endpoints/mesh_registration.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def _create_lod_task_config(
+    client: entitysdk.client.Client,
+    entity_id: UUID,
+    obj_asset_id: str,
+) -> UUID:
+    """Create a TaskConfig entity encoding the LOD generation inputs."""
+    L.info(f"Creating LOD generation TaskConfig for entity {entity_id} …")
+    config_payload = json.dumps(
```
jplanasc: I think this is going to fail... Check how the download function accesses the uploaded data:

https://github.com/openbraininstitute/obi-one/blob/d0ca3f71b59836da6b1a8e032acc67c14d15aa96/obi_one/core/run_tasks.py#L109-L118

obi_one/scientific/tasks/mesh_lod_generation/task.py
dkeller9: It worked on my local testing setup. I've now refactored it with the client though.

obi_one/scientific/tasks/mesh_lod_generation/config.py
dkeller9: The alias was just there for convenience but is not needed. I've removed it.

obi_one/scientific/tasks/mesh_lod_generation/task.py
dkeller9: Execute now has these fields.

app/endpoints/mesh_registration.py
dkeller9: I've added extra checks to make sure that the method works.

obi_one/scientific/tasks/mesh_lod_generation/estimate.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def estimate_mesh_lod_generation_count(
+    *,
+    db_client: entitysdk.Client,
+    config_id: UUID,
+) -> int:
+    _ = db_client
+    _ = config_id
+    return 1
```
jdcourcol: We don't need any of this. This is not going to accounting database.

We just need to launch something from launch-system. No need for ScanConfig and this kind of thing for that use case.

jdcourcol (changes requested):

obi_one/scientific/tasks/mesh_lod_generation/task.py
```diff
@@ -0,0 +1,139 @@
...
+
+def _upload_lod_directory(
+    client: entitysdk.Client,
+    entity_id: UUID,
+    lod_files: dict[os.PathLike, os.PathLike],
+) -> str:
+    client.upload_directory(
+        entity_id=entity_id,
+        entity_type=EMCellMesh,
+        name="lod-mesh-directory3",
```
jdcourcol: lod-mesh-directory3 ? I don't think we want to name it like that.

jdcourcol (dismissed):

obi_one/scientific/tasks/mesh_lod_generation/task.py
jdcourcol: can we settle on something ? because Fabien needs to read this.

dkeller9: This was a temporary renaming, should be lod-mesh-directory.

dkeller9: Now changed back to `lod-mesh-directory`

obi_one/scientific/tasks/mesh_lod_generation/estimate.py
dkeller9: `estimate.py` is now removed as well as `ScanConfig`

obi_one/scientific/tasks/mesh_lod_generation/task.py
dkeller9: @jplanasc I have tested that the endpoint can be called and works as intended.

jplanasc (approved):
LGTM.

PR #808, opened 2026-06-08
EM synapse mapping task: fail fast if duplicate morphologies are passed (same `pt_root_id`)

jplanasc:
Fixes #805


obi_one/scientific/tasks/em_synapse_mapping/task.py
```diff
@@ -121,6 +129,7 @@ def execute(  # NOQA: PLR0914, PLR0915, C901, PLR0912
             merge_spiny_morphologies(
                 source_files=[out_root / rn.fn_morph_h5 for rn in resolved_neurons],
                 output_path=out_root / fn_merged_h5,  # ty:ignore[unsupported-operator]
+                neuron_id_strings=[rn.name_in_circuit for rn in resolved_neurons],
```
jplanasc, line 132: @MWolfR : I haven't checked it thoroughly, but would it be safer to make pairs of (`source_file`, `neuron_id_string`) instead of passing 2 different parameters?

MWolfR: Mhm, that is right. 
But both inputs are assembled right here from `resolved_neurons`, not much that can go wrong.
I can change it if you want, but I think it's tolerable.

jplanasc: Yes, I prefer having them grouped. In any case, you're `zip`ping them in the function, so that will be already done once the function is called.

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
```diff
@@ -24,10 +24,52 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
```
jplanasc: Notebook execution should be clean.

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
```diff
@@ -166,7 +167,9 @@
    "outputs": [],
    "source": [
     "import os\n",
-    "os.environ[\"CAVE_TOKEN\"] = \"PASTE_YOUR_TOKEN_HERE\""
+    "import caveclient\n",
+    "os.environ[\"CAVE_TOKEN\"] = \"dac7bc9fc59ac3555044d4c9d878927e\"\n",
```
jplanasc: Tokens should not get into the public repo.

Why do you need both variables defined?

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
```diff
@@ -184,92 +187,46 @@
...
-    "    print(\"=\" * 60)\n",
-    "    for name, pt_id in no_int_conns:\n",
-    "        print(f\"  {name} (pt_root_id={pt_id})\")"
+    "\n",
+    "assert cave.materialize.version == 1507\n",
+    "\n",
+    "syns = cave.materialize.synapse_query(post_ids=pt_root_ids, pre_ids=pt_root_ids)\n",
+    "raw_mat = syns.groupby([\"pre_pt_root_id\", \"post_pt_root_id\"])[\"valid\"].count().unstack(\"post_pt_root_id\", fill_value=0)\n",
+    "mat = raw_mat.reindex(columns=pt_root_ids, index=pt_root_ids).fillna(0).to_numpy()\n",
+    "mat"
```
jplanasc: I think this matrix lacks information on which neuron is which in rows/columns. Since this is an explanatory notebook for new users, everything should be explained and very explicit.

I'd add something like this:
```
print(f"Neuron order (pt_root_id):\n{pt_root_ids}")
print(f"Degrees:\n{mat}")
```

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
```diff
@@ -299,8 +255,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from obi_one.scientific.tasks.em_synapse_mapping.config import EMSynapseMappingSingleConfig\n",
+    "from obi_one.scientific.tasks.em_synapse_mapping.config import EMSynapseMappingSingleConfig, AdvancedEMSynapseMappingOptions\n",
     "from obi_one.scientific.tasks.em_synapse_mapping.task import EMSynapseMappingTask\n",
+    "from obi_one.scientific.from_id.named_tuple_from_id import EMSynapseMappingInputNamedTuple\n",
     "from obi_one.core.info import Info\n",
     "\n",
     "# We take the last 3 digits of pt_root_id to easily identify the neurons\n",
```
jplanasc: @MWolfR : what do you think? Should we take the last 4 digits of `UUID` instead of the `pt_root_id`? Isn't it more representative now?

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
```diff
@@ -184,92 +187,46 @@
...
+    "mat"
    ]
   },
   {
    "cell_type": "markdown",
    "id": "c95b24fe",
    "metadata": {},
    "source": [
-    "Now we pick the pair of neurons with the highest connectivity between them."
+    "Now we pick the five neurons with the highest degrees."
```
jplanasc: I think 5 neurons is too much for an illustrative example. I suggest to keep 2, or maximum 3, so it runs faster and it's lighter.

Is there any added value between 2-3 to 5?

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
MWolfR: Yes. Sorry.

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
MWolfR: Just remove the last `to_numpy`. Then the rows / columns are labelled by pt_root_id.

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
jplanasc: Only `CAVECLIENT_MICRONS_API_KEY` is needed.

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
jplanasc: We'll do 3.

examples/H_map_em_synapses/map_em_synapses_multineuron.ipynb
jplanasc: Will use UUID[-3].

obi_one/scientific/tasks/em_synapse_mapping/task.py
jplanasc: Thinking better about it... I think we should delegate to `morph-spines` the file merge. I'll implement this first, then come back to this code.

As we discussed if we rename morphologies, we need to update the links in the spine table. I don't think EM synapse mapping task should be aware of all these implications. Better to offload to morph-spines, since it knows the format and all the intricacies.

jplanasc: This point will be tackled in a subsequent PR.

jplanasc (commented):
LGTM, but I can't approve, as I'm author of this PR.

MWolfR (approved):

PR #810, opened 2026-06-09
Circuit registration with optional validation

chr-pok:
Made SNAP validation for internal circuit registration optional and skipped validation for EM synapse mapping.


MWolfR (approved):
Thank you, this will be useful

PR #814, opened 2026-06-10
Add target simulator classvar to ion channel sim config

james-isbister:
* Add target simulator classvar to ion channel sim config

Note that ICM could eventually inherit from the NEURON Base but then we'd have to create a new intermediate class for scale neuron and above which has a soma location


AurelienJaquier (approved):
thanks!

PR #816, opened 2026-06-10
accounting for em synapse mapping

jdcourcol:
This is for https://github.com/openbraininstitute/prod-build-circuit/issues/33


james-isbister (approved):
Not so familiar with this accounting stuff, but LGTM

Issue #770, opened 2026-05-13
Add interactive morphology location selection in morphology viewer


ilkilic:
## Description

Add morphology-viewer support for creating deterministic morphology locations.

## Scope

- select one or multiple morphology points in the viewer
- highlight selected points/sections
- populate `ExplicitMorphologyLocations.locations`
- coordinate OBI-One enablement of `ExplicitMorphologyLocations` in the normal `MorphologyLocations` config/materialization path

## Backend Representation

```python
ExplicitMorphologyLocations(
    locations=[
        MorphologyLocationPoint(section_id=1, offset=0.5),
    ]
)
```

## Acceptance Criteria

- [x] Users can select morphology locations in the viewer
- [x] Selected locations are highlighted
- [x] Multiple locations can be selected
- [x] Viewer selections create/update `ExplicitMorphologyLocations`
```

james-isbister:
I tried this as part of reviewing the obi-one branch. It works very nicely. I noticed a few things and had the following comments:
- Explicit Morhology Locations is currently exposed for Single Neuron and Synaptome but the UI just works for Synaptome. I guess this requires homogenisation of the two 3D views. We should support both or remove the block from Single Neuron in obi-one for now @ilkilic 
- The text "Click a basal or apical dendrite in the 3D viewer to add a location. Click a location again to remove it." is so small that I kind of ignored it. Maybe something visually stronger with an arrow pointing to the 3D circuit or dynamic animation such as the 3D view zooms in slightly, would help the user realise the two are connected
- If I rotate the 3D view, sometimes a location is added. Maybe if the click and swipe are too close together we don't add a location.
- The "Click to add this location" label is very small when hovering over a neurite. This could be made more prominent? Also maybe "Add to Explicit Morphology Locations" is better - I'm not sure? 
- When hovering over a neurite when an ExplicitMorphologyLocations is not selected, it would also be cool to show a label like "Add to New Explicit Morphology Locations" I think?
- There is now a file browser opened when you click on circuit_config.json for viewing the circuit components in the "Simulations" tab. Not sure if that's meant to be in this PR? Something like that for the "Circuit" entity would be cool though, with a selector for Preview/File Browser, for example. A lot of the file types currently aren't supported even though they're really just text (hoc, mod) **Addition:** The other input files (simulation_config.json, compartment_sets.json, node_sets.json, obi-one-coordinate, could be grouped under a Simulation pill with file browser.

bilalesi:
[...]

Issue #769, opened 2026-05-13
Add deterministic MorphologyLocations blocks for explicit morphology targeting


ilkilic:
## Description

Add a backend representation for deterministic morphology locations.

This introduces explicit morphology-location points defined by:

```python
section_id: int
offset: float
```

and supports one or multiple points through a single block:

```python
ExplicitMorphologyLocations(
    locations=[
        MorphologyLocationPoint(section_id=1, offset=0.5),
        MorphologyLocationPoint(section_id=2, offset=0.25),
    ]
)
```

## Scope

This PR adds:

- `MorphologyLocationPoint`
- `ExplicitMorphologyLocations`
- validation for:
  - `section_id >= 0`
  - `0.0 <= offset <= 1.0`
  - non-empty `locations`
- conversion to the existing morphology-location dataframe format
- tests for single/multiple locations and serialization

## Out Of Scope

`ExplicitMorphologyLocations` is not exposed in the generic UI/config union yet.

Raw `section_id` values are not user-friendly as a manual input. UI exposure should be handled by the morphology viewer/editor work, where selected morphology points can populate `ExplicitMorphologyLocations.locations`.

## Follow-Up

The morphology viewer/editor [ticket](https://github.com/openbraininstitute/obi-one/issues/770) will:

- expose/create `ExplicitMorphologyLocations` from the UI
- support selecting one or multiple morphology points

## Acceptance Criteria

- [x] Explicit morphology locations can be represented in the backend
- [x] Single explicit location is supported
- [x] Multiple explicit locations are supported
- [x] Validation rejects invalid section IDs, offsets, and empty location lists
- [x] Existing morphology-location dataframe output format is preserved

PR #818, opened 2026-06-11
Add deterministic explicit morphology locations

ilkilic:
## Summary

Implement https://github.com/openbraininstitute/obi-one/issues/769

Add deterministic morphology targeting through a new `ExplicitMorphologyLocations` block.

Locations are represented as a non-empty collection of:

```python
MorphologyLocationPoint(section_id=1, offset=0.5)
```

The block supports both single and multiple explicit locations.

## Implementation

- Add `MorphologyLocationPoint` as a nested model.
- Add `ExplicitMorphologyLocations`
- Validate:
  - `section_id` is a non-negative integer.
  - `offset` is between `0.0` and `1.0`.
  - At least one location is provided.
  - Referenced neurite sections exist when `points_on()` is called.
- Update the morphology locations example notebook.

## Section ID convention

Locations use SONATA global section IDs:

- `0`: soma
- `1..N`: neurite sections in MorphIO `nrn_order`

Therefore, SONATA neurite section ID `N` maps to MorphIO section `N - 1`.

The generated dataframe is compatible with existing random morphology locations and contains `section_id` and `normalized_section_offset` for compartment-set materialization.

## Scope note
ExplicitMorphologyLocations is intentionally not exposed in the generic MorphologyLocationUnion yet.
Raw section IDs are not a good default manual UI. A follow-up frontend [ticket](https://github.com/openbraininstitute/obi-one/issues/770) will expose this block in the normal MorphologyLocations config/materialization path and populate it from selected morphology points.

## Additional fix

Fix normalized section-offset/path-distance calculations for random morphology locations

## Tests

Added coverage for:

- Single neurite location
- Soma location
- Multiple locations
- Invalid section IDs and offsets
- Empty location lists
- Missing morphology sections

Compartment-set materialization is provided by the dependent infrastructure in PR #498.


obi_one/scientific/blocks/morphology_locations/explicit.py
```diff
@@ -0,0 +1,136 @@
...
+
+    distance_on_section = offset * section_length
+    cumulative_lengths = np.cumsum(segment_lengths)
+    segment_id = min(
+        int(np.searchsorted(cumulative_lengths, distance_on_section, side="right")),
+        len(segment_lengths) - 1,
+    )
+    segment_start = 0.0 if segment_id == 0 else float(cumulative_lengths[segment_id - 1])
+
+    path_distance = distance_on_section
```
james-isbister: Probably we can reuse path_distance_all_segments_from in the library for calculating the path distance

james-isbister (commented):
Looks good. Just one suggestion

obi_one/scientific/blocks/morphology_locations/explicit.py
ilkilic: Good point, I reused path_distances directly since we only need the distance for one segment here.

obi_one/scientific/blocks/morphology_locations/explicit.py
```diff
@@ -0,0 +1,158 @@
...
+        description="Normalized location along the section.",
+    ),
+]
+
+
+class MorphologyLocationPoint(BaseModel):
+    """A SONATA global section ID and normalized offset defining one location."""
+
+    model_config = ConfigDict(extra="forbid")
+
```
james-isbister, line 66: Probably 'population' and 'node_id' are required if we plan to move to circuits? Could the list of valid populations be fetched from the endpoint using the same mechanics used [here](https://github.com/openbraininstitute/obi-one/blob/d68d612a436e5d501565a1d83272eede49e1e28b/obi_one/scientific/blocks/neuron_sets/population.py#L152)

Not sure what we would fill for the single neuron case

james-isbister (commented):
A comment on the >1 neuron case. Maybe worth considering now?

obi_one/scientific/blocks/morphology_locations/explicit.py
ilkilic: Good point. I think I would still keep population and node_id on the neuron-set side rather than on each location point. `ExplicitMorphologyLocations` already inherits neuron_set from `MorphologyLocationsBlock` and materialization resolves that to the population and node ids when writing the compartment-set rows.
For different explicit locations on different nodes/populations, we can use separate `ExplicitMorphologyLocations` blocks with different neuron sets. What do you think?

james-isbister: I guess there are two use cases:

**1) The user wants to stimulate/record from the same section/offset combinations across all neurons in a neuron set**
- This seems more useful say if the user wants to record from specific sections/offsets which are meaningful across neurons? Say the AIS, but I suppose there could be other use cases in theory

**2) The user wants to click on different locations on different neurons to apply the same recordings/stimuli** 
- Creating a separate ExplicitMorphologyLocations, NeuronSet and Stimulus/Recording block for each neuron seems a bit tedious

Not sure how much we need 1) but 2) definitely seems useful. If I had to choose one of them it would be 2). But maybe we really need both?

ilkilic: Yes, I see your point. Use case 2 is fairly cumbersome to configure with the current block. Would it be okay to keep this PR focused on the existing ExplicitMorphologyLocations semantics and introduce a separate per-neuron explicit-locations block in a follow-up?

james-isbister: Yes, sounds good to me!

ilkilic: I created an [issue](https://github.com/openbraininstitute/obi-one/issues/948).

ilkilic:
@james-isbister this should be ready now if you want to test it with core-web-app: https://github.com/openbraininstitute/core-web-app/pull/1863

james-isbister (approved):
LGTM!

PR #819, opened 2026-06-11
Accounting simulation

jdcourcol:
accounting for microcircuit simulation was not taking duration into account


james-isbister (approved):
LGTM

PR #821, opened 2026-06-11
New neuron sets - Functional implementation

chr-pok:
Functional implementation of the new neuron set classes (in `/neuron_sets_2`) added according to [this proposal](https://openbraininstitute.sharepoint.com/:w:/s/OBI-Scientificstaff/IQBjp78-IWL8R74aVpqTIjdNAY24O5CXqmFWwoSTM3F_yKo?e=MAEs0x).

- Typed neuron set classes: Population, ID, Property, Predefined, Combined, Specific (AllNeurons)
- Each type has Biophysical/Point/Virtual/NonVirtual/Any variants with runtime type enforcement
- Combined neuron sets (union/intersect/diff) incl. nesting with cycle detection
- Currently limited to two neuron sets (not arbitrary list) for easier UI integration, since all UI elements exist; list support will require a new UI element (will create a separate ticket for it) --> https://github.com/openbraininstitute/prod-circuit-simulation/issues/210
- Symbolic SONATA expressions preserved when possible, explicit IDs resolved when needed
- Example notebook and unit tests added

Related issue: https://github.com/openbraininstitute/prod-circuit-simulation/issues/112

### DIFF
--- a/examples/B_blocks/neuron_set_2_example.ipynb
+++ b/examples/B_blocks/neuron_set_2_example.ipynb
@@ -68,7 +68,7 @@
    "outputs": [],
    "source": [
     "from obi_one.scientific.blocks.neuron_sets_2.combined import BiophysicalCombinedNeuronSet, SetOperation\n",
-    "from obi_one.scientific.blocks.neuron_sets_2.id import IDNeuronSet, BiophysicalIDNeuronSet\n",
+    "from obi_one.scientific.blocks.neuron_sets_2.id import BiophysicalIDNeuronSet\n",
     "from obi_one.scientific.blocks.neuron_sets_2.population import PopulationNeuronSet, BiophysicalPopulationNeuronSet\n",
     "from obi_one.scientific.blocks.neuron_sets_2.predefined import PredefinedNeuronSet, PredefinedPopulationNeuronSet, PredefinedVirtualPopulationNeuronSet\n",
     "from obi_one.scientific.blocks.neuron_sets_2.property import PropertyNeuronSet\n",
--- a/examples/B_blocks/neuron_set_2_example.ipynb
+++ b/examples/B_blocks/neuron_set_2_example.ipynb
@@ -67,12 +67,14 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from obi_one.scientific.blocks.neuron_sets_2.population import PopulationNeuronSet, BiophysicalPopulationNeuronSet\n",
-    "from obi_one.scientific.blocks.neuron_sets_2.predefined import PredefinedNeuronSet, PredefinedPopulationNeuronSet\n",
+    "from obi_one.scientific.blocks.neuron_sets_2.combined import BiophysicalCombinedNeuronSet, SetOperation\n",
     "from obi_one.scientific.blocks.neuron_sets_2.id import IDNeuronSet, BiophysicalIDNeuronSet\n",
+    "from obi_one.scientific.blocks.neuron_sets_2.population import PopulationNeuronSet, BiophysicalPopulationNeuronSet\n",
+    "from obi_one.scientific.blocks.neuron_sets_2.predefined import PredefinedNeuronSet, PredefinedPopulationNeuronSet, PredefinedVirtualPopulationNeuronSet\n",
     "from obi_one.scientific.blocks.neuron_sets_2.property import PropertyNeuronSet\n",
     "from obi_one.scientific.blocks.neuron_sets_2.property import NeuronPropertyFilter\n",
     "from obi_one.scientific.blocks.neuron_sets_2.specific import AllNeurons, AllBiophysicalNeurons\n",
+    "from obi_one.scientific.unions.unions_neuron_sets_2 import BiophysicalNeuronSetReference\n",
     "from obi_one.core.tuple import NamedTuple"
    ]
   },
@@ -175,6 +177,24 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Wrong population type\n",
+    "nset = PredefinedVirtualPopulationNeuronSet(\n",
+    "    node_set=\"Layer6\", population=\"S1nonbarrel_neurons\", sample_percentage=50, sample_seed=42\n",
+    ")\n",
+    "\n",
+    "# This should raise a ValueError about wrong population type\n",
+    "try:\n",
+    "    nset_def, combined = nset.get_node_set_definition(circuit)\n",
+    "except ValueError as e:\n",
+    "    print(f\"Caught expected error: {e}\")"
+   ]
+  },
+  {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
@@ -250,11 +270,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from obi_one.scientific.blocks.neuron_sets_2.combined import (\n",
-    "    BiophysicalCombinedNeuronSet, SetOperation,\n",
-    ")\n",
-    "from obi_one.scientific.unions.unions_neuron_sets_2 import BiophysicalNeuronSetReference\n",
-    "\n",
     "# Create two base neuron sets to combine\n",
     "nset_a = PredefinedPopulationNeuronSet(node_set=\"All\", population=\"S1nonbarrel_neurons\", sample_percentage=50, sample_seed=0)\n",
     "nset_a.set_block_name(\"nset_a\")\n",
@@ -362,6 +377,29 @@
     "    combined_x.get_neuron_ids(circuit)\n",
     "except ValueError as e:\n",
     "    print(f\"Caught expected error: {e}\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Create two base neuron sets to combine\n",
+    "nset_a = PredefinedPopulationNeuronSet(node_set=\"All\", population=\"S1nonbarrel_neurons\", sample_percentage=50, sample_seed=0)\n",
+    "nset_a.set_block_name(\"nset_a\")\n",
+    "\n",
+    "nset_b = PredefinedPopulationNeuronSet(node_set=\"All\", population=\"S1nonbarrel_neurons\", sample_percentage=50, sample_seed=1)\n",
+    "nset_b.set_block_name(\"nset_b\")\n",
+    "\n",
+    "# Wrap in BlockReferences (as required by the combined neuron set fields)\n",
+    "ref_a = BiophysicalNeuronSetReference(block_dict_name=\"neuron_sets\", block_name=\"nset_a\")\n",
+    "ref_a.block = nset_a\n",
+    "ref_b = BiophysicalNeuronSetReference(block_dict_name=\"neuron_sets\", block_name=\"nset_b\")\n",
+    "ref_b.block = nset_b\n",
+    "\n",
+    "print(f\"Set A (50%, seed 0): {nset_a.get_neuron_ids(circuit)}\")\n",
+    "print(f\"Set B (50%, seed 1): {nset_b.get_neuron_ids(circuit)}\")"
    ]
   },
   {
--- a/examples/B_blocks/neuron_set_2_example.ipynb
+++ b/examples/B_blocks/neuron_set_2_example.ipynb
@@ -138,6 +138,7 @@
    "outputs": [],
    "source": [
     "nset = PredefinedNeuronSet(node_set=\"AllPopulations\")\n",
+    "nset.set_block_name(\"nset\")\n",
     "\n",
     "ids = nset.get_neuron_ids(circuit)\n",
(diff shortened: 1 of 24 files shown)


examples/B_blocks/neuron_set_2_example.ipynb
james-isbister, line 1: IDNeuronSet is imported but not used

obi_one/scientific/blocks/neuron_sets_2/population.py
```diff
@@ -74,61 +53,52 @@ class PopulationNeuronSet(NeuronSet, abc.ABC):
...
         if len(ids) == 0:
             L.warning("Neuron set empty!")
 
-        return ids
+        return {self.population: ids.tolist()}
+
+    def get_node_set_definition(
+        self, circuit: Circuit, *, force_resolve_ids: bool = False
+    ) -> tuple[dict | list, dict]:
+        """Returns the SONATA node set definition, optionally forcing to resolve individual IDs.
```
james-isbister, line 98: I found the logic of the return statement very confusing at first, but it makes sense now:

If it's compound, the first value of the tuple is a list of the node_set names to be combined, and the second value is a dictionary where the keys are the names of the node_sets to be combined with values equal to the expressions

Otherwise, you just have the expression in the first element

Perhaps this can be better explained. An alternative might have been to make the first element always a dictionary and the second element a bool which illustrates how to interpret the dictionary (i.e. is_compound)

obi_one/scientific/blocks/neuron_sets_2/base.py
```diff
@@ -61,6 +207,14 @@ def _get_output_file(circuit: Circuit, file_name: str | None, output_path: str)
...
 
+    @staticmethod
+    def _check_existing(new_node_sets: dict, existing_node_sets: dict) -> None:
+        """Checks if new names already exist."""
+        existing = [f"'{n}'" for n in new_node_sets if n in existing_node_sets]
+        if existing:
+            msg = f"Node set(s) {', '.join(existing)} already existing!"
+            raise ValueError(msg)
+
     def to_node_set_file(
```
james-isbister, line 237: A key function. Perhaps some precise documentation of what the parameters do would be useful, as I'm left guessing what exactly they're impact will be

obi_one/scientific/blocks/neuron_sets_2/base.py
```diff
@@ -15,41 +17,185 @@
...
+        """Returns population names and types included in the neuron set."""
+        self.check_populations_in_circuit(circuit=circuit)
+
+        popul_types = {}
+        for pname in self.get_populations(circuit):
+            if circuit.sonata_circuit.nodes[pname].type == "biophysical":
+                ptype = SonataPopulationType.BIOPHYSICAL
+            elif circuit.sonata_circuit.nodes[pname].type == "virtual":
+                ptype = SonataPopulationType.VIRTUAL
+            elif circuit.sonata_circuit.nodes[pname].type.startswith("point_"):
```
james-isbister, line 108: There are two other types of point neuron we now support

See [here](https://github.com/openbraininstitute/obi-one/blob/7cab6f0f3cad3a7fdf9412300f59f555e3c782ec/obi_one/scientific/library/circuit_metrics.py#L33)

> TYPES_OF_POINT_NODES = ["point_process", "point_neuron", "brian2_point", "inait_point_neuron_lif"]

I think we'll need to update this statement. Also `Circuit.get_node_population_names` will need to be updated with this also. Are there other places too? Perhaps we can use a single list of constants across these different places

obi_one/scientific/unions/unions_neuron_sets_2.py
```diff
@@ -3,23 +3,95 @@
...
+    | NonVirtualPropertyNeuronSet
+    | PredefinedNonVirtualPopulationNeuronSet
+    | NonVirtualCombinedNeuronSet
+    | AllNonVirtualNeurons
+)
+_ALL_NEURON_SETS = (
+    _BIOPHYSICAL_NEURON_SETS
+    | _VIRTUAL_NEURON_SETS
+    | _POINT_NEURON_SETS
+    | _NONVIRTUAL_NEURON_SETS
```
james-isbister, line 69: Is there really a need for these 4 neuron sets:
- PopulationNeuronSet
- IDNeuronSet
- PropertyNeuronSet
- PredefinedPopulationNeuronSet

Which if I understand correctly, allow you to choose any population. I think we can achieve the same functionality with the more specific types (i.e. Biophysical..., NonVirtual...., etc). I think removing these 4 classes would reduce complexity both by the reduction in total classes (in the code + UI) and in the management of reference types etc.

Maybe I have missed a possible use case though

There is also inconsistency in the naming. Some have "Population" in the name, whilst other do not.

obi_one/scientific/blocks/neuron_sets_2/predefined.py
```diff
@@ -0,0 +1,261 @@
...
+                node_ids = circuit.sonata_circuit.nodes[npop].ids(node_set)
+            except snap.BluepySnapError:
+                # In case of an error (e.g., "No such attribute"), return empty list
+                node_ids = []
+            if len(node_ids) > 0:
+                node_populations.append(npop)
+        return node_populations
+
+
+class PredefinedNeuronSet(PredefinedBaseNeuronSet):
```
james-isbister: Hm, I'm not sure how we'll be able to use this in different contexts based on the static typing

obi_one/scientific/blocks/neuron_sets_2/combined.py
```diff
@@ -1,71 +1,369 @@
...
+
+    json_schema_extra_additions: ClassVar[dict] = {
+        SchemaKey.BLOCK_USABILITY_DICTIONARY: {
+            SchemaKey.PROPERTY_GROUP: MappedPropertiesGroup.CIRCUIT,
+            SchemaKey.PROPERTY: CircuitUsability.SHOW_NEURON_SETS,
+            SchemaKey.FALSE_MESSAGE: "This circuit has no populations.",
+        },
+    }
+
+    base_neuron_set: BlockReference | None = Field(
```
james-isbister, line 218: Through-out the different CombinedNeuronSets, we should be using the appropriate reference unions and reference type lists defined [here](https://github.com/openbraininstitute/obi-one/blob/4d4c16681f3e38a782c88d0c5084a721c6e4550c/obi_one/scientific/unions/unions_neuron_sets_2.py#L153)

Such as:
`base_neuron_set: ALL_NEURON_SETS_REFERENCE_UNION | None `

and:
`SchemaKey.REFERENCE_TYPES: ALL_NEURON_SETS_REFERENCE_TYPES`

In this case

obi_one/scientific/unions/unions_neuron_sets_2.py
```diff
@@ -3,23 +3,95 @@
...
+    | PopulationNeuronSet
+    | IDNeuronSet
+    | PropertyNeuronSet
+    | PredefinedNeuronSet
+    | PredefinedPopulationNeuronSet
+    | CombinedNeuronSet
```
james-isbister, line 71: If I understand correctly, the only place the general CombinedNeuronSet could be used would be as the source of a spike replay stimulus as that is the only context where biophys, virtual and point neurons can all be used.

It is a similar problem to the "PredefinedNeuronSet" which could be biophys, virtual or point, and therefore only usable in the spike replay context. The only awkward context would be if there was a predefined neuron set which combined all three types (biophys, virtual or point) and the user wanted to use this as, say, the target of the simulation. I think this is such a rare case though? Is it even common to find multi-population node_sets of multiple different types (including virtual)? I think not

If this logic is correct, we might even consider either:
1) Only allowing the PredefinedNeuronSet and CombinedNeuronSet to be used as the source of spike replay
2) Allowing these neuron sets to be used in any context where a NonVirtual NeuronSet can be used (i.e. simulation target, etc). But I don't know what would happen if the target of the simulation included virtual nodes

I'd probably lean towards 1.

obi_one/scientific/blocks/neuron_sets_2/combined.py
james-isbister: Although we could do this after merging into the new_neuron_sets branch if you prefer

examples/B_blocks/neuron_set_2_example.ipynb
chr-pok: Removed unused import from notebook

obi_one/scientific/blocks/neuron_sets_2/population.py
chr-pok: Yeah, this is complex indeed. I added detailed explanation to the docstring.

> An alternative might have been to make the first element always a dictionary and the second element a bool which illustrates how to interpret the dictionary (i.e. is_compound)

I think this would not work for all cases, e.g., for compound expressions (i.e., list) but where the second dict is still empty because only existing node sets are referenced in the compound. So I would rather leave it as is but with better explanations.

obi_one/scientific/blocks/neuron_sets_2/base.py
chr-pok: Added more details to the docstring

obi_one/scientific/blocks/neuron_sets_2/base.py
chr-pok: Will handle this in a separate PR, since we need to rebase to main to have `TYPES_OF_POINT_NODES` available.

obi_one/scientific/unions/unions_neuron_sets_2.py
chr-pok: Yes, you are right! I thought more about it and I think that even the NonVirtual population types are not really needed, since also these cases are covered by more specific type (biophysical or point). So I removed 8 population neuron sets in total (4 any and 4 non-virtual).

Also, I renamed the "Population" classes and used a consistent naming schema:
- Base classes: {Method}PopulationBaseNeuronSet
- Concrete classes: {Type}Population{Method}NeuronSet
(We can still rename again if you have a better schema in mind.)

obi_one/scientific/blocks/neuron_sets_2/combined.py
chr-pok: Yeah, this is a bit of a problem. `base_neuron_set: ALL_NEURON_SETS_REFERENCE_UNION | None` is not possible because of circular imports (because the unions import the combined neuron sets).

There is no clean solution I could find, so I followed Kiro's suggestion to leave it as generic `BlockReference`. On the UI side, this should not be a problem since the correct types are specified in `json_schema_extra`.

We can think about it more and fix it later, if needed.

obi_one/scientific/unions/unions_neuron_sets_2.py
chr-pok: > If I understand correctly, the only place the general CombinedNeuronSet could be used would be as the source of a spike replay stimulus as that is the only context where biophys, virtual and point neurons can all be used.

Yes indeed.

> ... I think this is such a rare case though? Is it even common to find multi-population node_sets of multiple different types (including virtual)? I think not

Exactly, this is a very rare case which could happen more like by accident (e.g., when keeping external populations as virtual ones but which still have certain node properties like layer, such that they would be included in node sets based on layer. But this is usually not intended and we should not even support this I think.

If we think we need something like this in the future, there is one type of neuron sets which I haven't thought of before, namely a `NonVirtualPredefinedNeuronSet` (i.e., not a population neuron set) which would resolve an existing node set within all biophysical and point neuron populations (similar to `AllNonVirtualNeurons`).

> 1. Only allowing the PredefinedNeuronSet and CombinedNeuronSet to be used as the source of spike replay

Yes, this makes sense. But I am not even sure we should support spike replay from multiple populations within one stimulus block? Currently, it uses `ALL_NEURON_SETS_REFERENCE_UNION` which does not include the general `CombinedNeuronSet` and `PredefinedNeuronSet` types, but we can add them later of course.

obi_one/scientific/blocks/neuron_sets_2/combined.py
james-isbister: I see. Yes, let's find a solution later. Just to make a note for later, two other possible solutions might be to:
1) Create a union within this class, which includes the class itself. Not sure if that's possible
2) We can't use CombinedNeuronSets within a CombinedNeuronSet - Hopefully we can avoid this

obi_one/scientific/unions/unions_neuron_sets_2.py
james-isbister: I agree. I think we can come back to NonVirtualPredefinedNeuronSet in the future if we need it, and consider which neuron sets are supported by the different stimuli when we integrate with the generate_simulations task

obi_one/scientific/unions/unions_neuron_sets_2.py
james-isbister: Sounds good! 

I noticed that there are two neuron sets which are defined and potentially useful but not in unions_neuron_set2.py, namely:
- NonVirtualCombinedNeuronSet - Useful as the target of the simulation, and potentially as the target of non virtual supporting stimuli + recordings? Related to your comment below: "not even sure we should support spike replay support spike replay from multiple populations within one stimulus block", but I think we can deal with this when we integrate with the simulation generation task?
- AllNonVirtualNeurons - Also seems useful as the default target of the simulation?

PR #824, opened 2026-06-12
Robust circ plots

MWolfR:
Made automatically generated circuit plots more robust.
No longer assumes the values of "synapse_class" must be "EXC" or "INH".

### DIFF
--- a/obi_one/scientific/library/basic_connectivity_plots_helpers.py
+++ b/obi_one/scientific/library/basic_connectivity_plots_helpers.py
@@ -16,10 +16,10 @@ import numpy as np
 import pandas as pd
 from conntility import ConnectivityMatrix
 from matplotlib import gridspec
+from matplotlib.colors import Colormap
 from matplotlib.gridspec import GridSpec
 from matplotlib.patches import Ellipse, FancyArrow
 from matplotlib.ticker import FuncFormatter
-from matplotlib.colors import Colormap
 from mpl_toolkits.axes_grid1.inset_locator import inset_axes
 
 # Connectivity dependencies (optional) - check for networkx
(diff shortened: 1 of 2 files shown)


obi_one/scientific/tasks/basic_connectivity_plots.py
```diff
@@ -20,6 +20,12 @@
 from obi_one.core.scan_config import ScanConfig
 from obi_one.core.single import SingleConfigMixin
 from obi_one.core.task import Task
+from obi_one.scientific.library.basic_connectivity_plots_helpers import (
+    _CANONICAL_EXC,
+    _CANONICAL_INH,
```
chr-pok: Should not import private constants; the underscore should be dropped if used in other modules.

obi_one/scientific/library/basic_connectivity_plots_helpers.py
```diff
@@ -46,10 +47,39 @@
...
 
 # Stats functions
 
 
+def find_canonical_synapse_classes(prop_values: list[str]) -> dict[str, str]:
+    canon_mapping = {}
+    for val in prop_values:
+        if _CANONICAL_EXC in val.lower():
+            canon_mapping[_CANONICAL_EXC] = val
+        if _CANONICAL_INH in val.lower():
```
chr-pok: In theory, there could be cases where the value matches both canonical values, in which case an incorrect mapping would be made. Consider using `elif` to avoid this.
```suggestion
        elif _CANONICAL_INH in val.lower():
```

obi_one/scientific/library/basic_connectivity_plots_helpers.py
```diff
@@ -200,7 +230,7 @@ def make_pie_plot(  # noqa: PLR0914
     cmap = cmaps[grouping_prop]
     if grouping_prop == "synapse_class":
         # Fix red/blue for EXC/INH, if NA make it gray
-        color_map = {"EXC": cmap(cmap.N), "INH": cmap(0), "NA": "gray"}
+        color_map = assemble_property_colormapping(conn, cmap, color_property=grouping_prop)
```
chr-pok, line 245: Looks like the "NA" case is no longer assigned to gray in the new mapping. I think we have such cases in MICrONS and H01.

chr-pok (commented):
Works fine, just a few things I noticed. Also, requires rebasing to main and fixing the CI tests.

obi_one/scientific/library/basic_connectivity_plots_helpers.py
MWolfR: That is intentional.

obi_one/scientific/library/basic_connectivity_plots_helpers.py
MWolfR: In the update, "NA" will be assigned to the middle value of the RedBlue colormap, which is sort of grey-ish. Not the exact same color. Is that similar enough?

obi_one/scientific/library/basic_connectivity_plots_helpers.py
MWolfR: Still fixed it :)

obi_one/scientific/library/basic_connectivity_plots_helpers.py
chr-pok: Sounds good

obi_one/scientific/library/basic_connectivity_plots_helpers.py
```diff
@@ -46,10 +47,49 @@
...
 
+def find_canonical_synapse_classes(prop_values: list[str]) -> dict[str, str]:
+    canon_mapping = {}
+    for val in prop_values:
+        if CANONICAL_EXC in val.lower():
+            canon_mapping[CANONICAL_EXC] = val
+        elif CANONICAL_INH in val.lower():
+            canon_mapping[CANONICAL_INH] = val
+        else:  # If both EXC and INH are found any additional value is interpreted as NA
+            canon_mapping[CANONICAL_NA] = val
```
chr-pok: To be on the safe side, could add an extra warning (or error) in case multiple different non-EXC/non-INH values would exist for some reason, since otherwise only the last one would be used in the mapping and the others silently ignored.
```suggestion
        elif CANONICAL_NA not in canon_mapping:
            canon_mapping[CANONICAL_NA] = val
        else:
            # Warning or error about multiple NA values
```

obi_one/scientific/library/basic_connectivity_plots_helpers.py
```diff
@@ -46,10 +47,49 @@
...
+    if (CANONICAL_EXC not in canon_mapping) or (CANONICAL_INH not in canon_mapping):
+        err_str = "No canon E/I mapping found!"
```
chr-pok:
```suggestion
        err_str = "No canonical E/I mapping found!"
```

chr-pok (approved):
LGTM!

PR #826, opened 2026-06-12
Circuit registration scale_override option

james-isbister:
* Override the calculated scale
* Needed for FlyWire which is whole-brain


obi_one/utils/circuit.py
```diff
@@ -105,6 +107,10 @@ def get_circuit_size(c: Circuit) -> tuple[types.CircuitScale, int, int, int]:
     # TODO: Add support for other scales as well
     # https://github.com/openbraininstitute/obi-one/issues/463
 
+    # Override scale if provided
+    if scale_override is not None:
```
chr-pok: Just wondering if this is the best place for the override. Perhaps we should only allow an override within the above `else` branch, since the scales "single", "pair", and "small" are exactly defined and should never be overridden I believe (?).

chr-pok (commented):
Looks good! Just one small question/concern regarding the location of the override.

chr-pok (approved):
Thanks for the change, looks all good to me now!

PR #832, opened 2026-06-16
duration should be read from sonata config.

jdcourcol:
it seems that the duration is not always present (only if we scan over duration), so we need to read it from the sonata configuration.


james-isbister (dismissed):
Yes, the scan_parameters are only for the scan parameters. I wouldn't rely on that at all. I'd even remove it, but that's not essential.

The only other place they would be accessible is the single_config. 

Better to get it from the sonata simulation config like you do here, because the simulation is useless without that anyway

james-isbister (approved):

PR #833, opened 2026-06-16
move logic to obi_one to remove dependency to app

jdcourcol:
Hello, I let the LLM move file around as with morphology registration we need:
- functions used in the end point to register morphologies
- functions used in task to register morphologies

and we don't want the library to import app packages.

so the functions are in obi_one now


james-isbister (commented):
Nice to separate this out. 

I'll let someone more familiar with morphologies and endpoint error handling check the details but I would mention that we already have two competing places for upload functions:
- [utils/db_sdk.py ](https://github.com/openbraininstitute/obi-one/blob/main/obi_one/utils/db_sdk.py)
- [utils/circuit_registration/register.py](https://github.com/openbraininstitute/obi-one/blob/main/obi_one/utils/circuit_registration/register.py)

This would be a third, so we might converge at some point but I don't think it's essential for getting what we need from these functions quickly

james-isbister (approved):

PR #835, opened 2026-06-16
clamp node positions to 0 when they exceed 4 times the stddev

mgeplf:


james-isbister (approved):
That catches the two problem neurons, thanks

PR #837, opened 2026-06-17
Circuit customization new emodels

AurelienJaquier:
This contain:

Validation of file (hoc and nodes) modifications:
- checking that the hoc files are valid hoc templates
- checking that the new hoc files are not using mechanisms that are not present in the circuit
- checking that only model_template and/or etype columns have been changed in the new nodes file
- checking that all hoc files used in the new nodes file exist, either from the new hoc e-models, or from the existing circuit e-models
- checking that all new hoc file are used in the new nodes file
- checking that the hoc files can be initialized by bluecellulab. This ensure that they have the expected structure and that they can be run by our platform.

circuit creation:
- replacing the old nodes file with the new nodes file
- adding new e-models to the circuit e-models and removing the ones that are not used by the circuit anymore
- re-computing the dynamics params, i.e. the holding current, the resting potential, the input resistance and the threshold current
- re-writing me-combos

circuit upload using `upload_customized_circuit`

example in `examples/L_utils/customiza_circuit_new_emodels.ipynb`


examples/L_utils/customiza_circuit_new_emodels.ipynb
chr-pok, line 1: Typo in the file name: Should be "customiz**e**_circuit_new_emodels.ipynb‎"

examples/L_utils/customiza_circuit_new_emodels.ipynb
```diff
@@ -0,0 +1,259 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "6e3decd9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Download existing circuit config\n",
```
chr-pok, line 146:
```suggestion
    "# Download circuit's existing node population file\n",
```

examples/L_utils/customiza_circuit_new_emodels.ipynb
```diff
@@ -0,0 +1,259 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d6a7fd91",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "modified_hoc_ids = [1]\n",
```
chr-pok, line 174: I know this is only a simple example, but would be nicer not to use a hard-coded ID but to select the neurons based on the actual template name, i.e., search for all neurons with `hoc:cADpyr_L2TPC` template and replace them by `hoc:cADpyr_L2TPC_v2`.

examples/L_utils/customiza_circuit_new_emodels.ipynb
```diff
@@ -0,0 +1,259 @@
...
+    "        nodes[\"nodes/S1nonbarrel_neurons/0/model_template\"][id_] = name\n",
+    "        nodes[\"nodes/S1nonbarrel_neurons/0/etype\"][id_] = etype"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c9d0e1f2",
+   "metadata": {},
+   "source": [
+    "## Validate modifications, Create modified circuit, and upload it\n",
```
chr-pok, line 188:
```suggestion
    "## Validate modifications, create modified circuit, and upload it\n",
```

examples/L_utils/customiza_circuit_new_emodels.ipynb
```diff
@@ -0,0 +1,259 @@
...
+   "id": "d0e1f2a3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "modified_circuit_name = \"nbS1-O1-E2PV-maxNsyn-HEX0-L2-modified\"\n",
+    "modified_circuit_description = \"nbS1-O1-E2PV-maxNsyn-HEX0-L2 with reduced sodium in soma and axon\"\n",
+    "run(\n",
+    "    client=client,\n",
+    "    parent_circuit_id=PARENT_CIRCUIT_ID,\n",
+    "    new_node_file_path=nodes_path,\n",
```
chr-pok, line 219: Just wondering, can `new_node_file_path` be `None` in case the hoc file name does not change, i.e., when replacing a hoc file by one with the same name?

examples/L_utils/customiza_circuit_new_emodels.ipynb
```diff
@@ -0,0 +1,259 @@
...
+    "    new_hoc_files_paths=MODIFIED_EMODELS_PATHS,\n",
```
chr-pok, line 220: Also the other way round, could `new_hoc_files_paths` be `None`, i.e., no new hoc files are provided but only their assignment to neurons is changed within the nodes file?

examples/L_utils/customiza_circuit_new_emodels.ipynb
chr-pok, line 219: How do you know which node population to update? There could be multiple biophysical ones, all with the same file name but within different folders. Also the corresponding hoc files may be in different folders depending on the population.

examples/L_utils/customiza_circuit_new_emodels.ipynb
```diff
@@ -0,0 +1,259 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d0e1f2a3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "modified_circuit_name = \"nbS1-O1-E2PV-maxNsyn-HEX0-L2-modified\"\n",
+    "modified_circuit_description = \"nbS1-O1-E2PV-maxNsyn-HEX0-L2 with reduced sodium in soma and axon\"\n",
+    "run(\n",
```
chr-pok, line 216: This function produces lots of log output (which is not easy to hide I guess) so it is not easy to understand what happened. So would be nice to get the `registered_circuit = run(...)` returned and print the circuit information (similar to what I do in my example notebook).

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+    if new_circuit_path is None:
+        # TODO: might want to change that in case users modify multiple times their circuits
+        new_circuit_path = f"{parent_circuit_path}-updated"
+    if new_circuit_path is not None and Path(new_circuit_path).exists():
+        msg = (
+            f"New circuit path {new_circuit_path} already exists. Please provide a different "
+            "path for the new circuit or delete the existing one."
+        )
+        raise ValueError(msg)
+    shutil.copytree(parent_circuit_path, new_circuit_path)
```
chr-pok, line 165: Would it be possible to modify the circuit in-place, i.e., without staging first and then copying the whole circuit folder? Because a circuit may be huge.

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+
+    Args:
+        parent_circuit_path: path to the parent circuit folder
+        new_nodes_file_path: path to the new nodes file
+        new_emodels_file_paths: list of paths to the new emodel hoc files
+        new_circuit_path: path to the new circuit folder to be created.
+            If None, a new folder will be created with name {parent_circuit_path}-updated.
+    """
+    from bluecellulab.circuit.circuit_access import EmodelProperties  # noqa: PLC0415
+    from bluecellulab.tools import (  # noqa: PLC0415
```
chr-pok, line 150: Any reason why these imports are within the function?

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+        calculate_SS_voltage,
+        compute_memodel_properties_v2,
+    )
+
+    # - copy old circuit into new circuit directory
+    if new_circuit_path is None:
+        # TODO: might want to change that in case users modify multiple times their circuits
+        new_circuit_path = f"{parent_circuit_path}-updated"
```
chr-pok, line 158: I think this could lead to unexpected behavior, because the `parent_circuit_path` is just a temp dir somewhere on the file system. So creating this folder my be fine, but it could remain there forever or might be cleaned up by the OS at some point. Maybe it would be better that the user is always required to provide a valid path (instead of accepting None)?

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+            node_file_path=new_node_file_path,
+            old_hoc_dir=parent_circuit_emodels_dir,
+            new_hoc_file_paths=new_hoc_files_paths,
+        )
+        check_new_hoc_in_nodes_file(
+            new_node_file_path=new_node_file_path,
+            new_hoc_paths=new_hoc_files_paths,
+        )
+
+        # TODO: try to have consitent namings
```
chr-pok, line 348: Still to be done, or can this comment be removed?

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+    model_template_names = {temp.split(":", 1)[1] for temp in model_templates}
+    return {Path(circuit_emodels_dir) / f"{name}.hoc" for name in model_template_names}
+
+
+def get_biophysical_population(circuit_path: str | Path) -> bluepysnap.nodes.NodePopulation:
+    """Open circuit and returns the biophysical population."""
+    circuit_config_path = Path(circuit_path) / "circuit_config.json"
+    circuit = bluepysnap.Circuit(circuit_config_path)
+    biophysical_pops = [pop for pop in circuit.nodes.values() if pop.type == "biophysical"]
+    if len(biophysical_pops) != 1:
```
chr-pok, line 126: Hmm, I think we should not rely on circuits having only one biophysical population. Maybe it would be better if the user would specify the population they want to customize, i.e., only one at a time. Then, instead of this function, there should be an additional validation that the specified population exists and is indeed "biophysical".

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+            emodel_properties=emodel_properties_filler,
+        )
+
+    # get mtype and layer for mecombo
+    mtype_layer_df = pop.get(properties=["layer", "mtype"])
+
+    # - update nodes file
+    # base path is defined in SONATA as /nodes/<population_name>/<index>
+    nodes_base_path = f"nodes/{pop.name}/0"
+    dynamics_params_base_path = f"{nodes_base_path}/dynamics_params"
```
chr-pok, line 261: Could there be additional dynamic parameters in this path other than the ones updated below? Maybe it would be safer to add a check and raise an error in case additional parameters exist for some reason but which are not updated by the code below.

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+
+    # - remove unused (old) emodel files if any
+    all_emodels_in_parent_circuit = set(Path(circuit_emodels_dir).glob("*.hoc"))
+    for emodel_fpath in all_emodels_in_parent_circuit:
+        if emodel_fpath not in all_emodel_file_paths:
+            emodel_fpath.unlink()
+
+    # - get holding current from parent circuit
+    ihold_column_name = "@dynamics:holding_current"
+    old_holding_currents = parent_pop.get(properties=[ihold_column_name])
```
chr-pok, line 208: What if the holding current does not exist in the parent? Should there be a dedicated check perhaps?

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+        memodels_dynamics_params[memodel] = compute_memodel_properties_v2(
+            hoc_fpath,
+            morph_fpath,
+            BCL_TEMPLATE_FORMAT,
+            holding_voltage,
+            emodel_properties=emodel_properties_filler,
+        )
+
+    # get mtype and layer for mecombo
+    mtype_layer_df = pop.get(properties=["layer", "mtype"])
```
chr-pok, line 256: Here again, what if layer or mtype properties don't exists?

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+    # - copy old circuit into new circuit directory
+    if new_circuit_path is None:
+        # TODO: might want to change that in case users modify multiple times their circuits
+        new_circuit_path = f"{parent_circuit_path}-updated"
+    if new_circuit_path is not None and Path(new_circuit_path).exists():
+        msg = (
+            f"New circuit path {new_circuit_path} already exists. Please provide a different "
+            "path for the new circuit or delete the existing one."
+        )
+        raise ValueError(msg)
```
chr-pok, line 164:
```suggestion
        raise FileExistsError(msg)
```

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+) -> list[dict[str:str]]:
+    """Returns a list of dict containing the hoc and morphology path for each memodel."""
+    hoc_morph_paths = []
+    _, new_hoc_morph_names_mapping, _ = map_ids_to_updated_memodel(
+        old_nodes_file_path=old_nodes_file_path,
+        new_nodes_file_path=new_node_file_path,
+        new_emodels_file_paths=new_hoc_files_paths,
+    )
+
+    for memodel in new_hoc_morph_names_mapping.values():
```
chr-pok, line 95: If I understand correctly, in `new_hoc_morph_names_mapping` multiple IDs could map to the same ME-model. Which means that this loop may iterate over duplicate values and will create duplicates in `hoc_morph_paths`.

Could iterate over unique values instead:
```suggestion
    for memodel in set(new_hoc_morph_names_mapping.values()):
```

obi_one/utils/circuit_customization/new_emodels.py
```diff
@@ -0,0 +1,382 @@
...
+    return old_id_mapping, new_id_mapping, new_to_old_memodel_mapping
+
+
+def get_hoc_morph_paths(
+    old_nodes_file_path: str | Path,
+    new_node_file_path: str | Path,
+    new_hoc_files_paths: list[str | Path],
+    parent_circuit_emodels_dir: str | Path,
+    available_parent_morph_dirs: dict[str, str],
+) -> list[dict[str:str]]:
```
chr-pok, line 86:
```suggestion
) -> list[dict[str, str]]:
```

obi_one/utils/circuit_customization/validations/new_emodels.py
```diff
@@ -0,0 +1,237 @@
...
+        # remove previously compiled mechanisms for each hoc
+        clean_compiled_mechanisms()
+        mech_dir = Path("mechanisms")
+        if mech_dir.exists():
+            shutil.rmtree(mech_dir)
+
+        _ = download_mechanisms(
+            circuit_id=str(circuit_id), db_client=db_client, dest_dir=Path(mech_dir)
+        )
+        compile_mechanisms(mechanisms_dir=mech_dir)
```
chr-pok, line 74: Looks like this downloads and compiles the same mechanisms in every loop iteration. Not sure how expensive this is. Could it be done only once outside the loop perhaps?

chr-pok (commented):
Nice work! Here a few questions, comments, and suggestions.

In general, I was wondering how expensive these customizations are. Have you already tested with larger circuits?

examples/L_utils/customiza_circuit_new_emodels.ipynb
AurelienJaquier: I thought that we would only have one biophysical population and one emodels folder per circuit. We will definitely have to take that into account in the new PR

obi_one/utils/circuit_customization/new_emodels.py
AurelienJaquier: Yes! At import, bluecellulab will import some NEURON stuff. I think it is loading the mechanisms at import, before we can compile them within the code if I recall correctly.
So it is better to import bluecellulab inside of functions, so that we don't encounter this issue simply by importing obi-one.

obi_one/utils/circuit_customization/new_emodels.py
AurelienJaquier: I actually had some cases of circuit having `threshold_current_legacy` entries in the dynamics params.
And that brought validation errors because the nodes file from the whole circuit and the nodes file when downloaded individually were different: One had the extra legacy params, and the other had not.
This is why I was a little bit lenient on the presence of other dynamics params, as, as far as I know, they do not have any influence on the circuit.

obi_one/utils/circuit_customization/new_emodels.py
AurelienJaquier: I thought we would always have a holding current?
If that is the case, does that mean that we do not compute the dynamics params at all? Or do we assume a default holding current of 0.0 nA and compute the dynamics params nonetheless (knowing that the threshold current depends on the holding current value)?

obi_one/utils/circuit_customization/new_emodels.py
AurelienJaquier: I did not know that that was a possibility. Then if layer and mtype are not present, I assume that we do not re-compute the mecombos?

obi_one/utils/circuit_customization/validations/new_emodels.py
AurelienJaquier: You are correct, we should do it outside of the loop

chr-pok:
Superseded by https://github.com/openbraininstitute/obi-one/pull/953.

PR #838, opened 2026-06-17
EMSynapseMapping Fix: Remove Initialize from SingleConfig

james-isbister:
* Removing as this caused a deserialization error
* The ty lint errors for the use of single config parameters in a task (which are typed for parameter scans is a larger question we don't need to worry about at this point imo)


AurelienJaquier (approved):
Ah yes indeed it is already in the scan config. Looks good then, thanks!

PR #843, opened 2026-06-18
Additional connectivity plots optimizations

james-isbister:
Additional speed up, when you have time to look @chr-pok 

> Two optimizations, both addressing the same root pattern — connalysis functions that don't scale to
>   whole-brain (127k-node) circuits:
> 
>   1. Pathway connection probability (the asked-for bottleneck)
> 
>   - What: connectivity_pathway computed within-distance connection probability per pathway via connalysis's
>   connection_probability_within, which materializes the full within-distance pair mask per submatrix.
>   - Fix: Swapped in an edge-based source for conntility's conn.analyze
>   (_connection_probability_within_pathway_source → directed_connection_probability_within), counting over
>   existing edges. All of conntility's pathway grouping/unstacking is untouched.
>   - Subtlety handled: conntility always passes v as a (pre, post) tuple (even for diagonal pathways) and
>   connalysis skips self-pair removal in that case — so I generalized the helper to do the same for tuples.
>   - Result: 297.9 s → 8.2 s (~36×); output byte-for-byte identical to connalysis across both test matrices,
>   synapse_class/mtype/layer, including cross-population EXC→INH pathways.
> 
>   2. Node degree densification (found while verifying end-to-end)
> 
>   - What: node_degree(adj) calls adj.toarray(), densifying the 127k×127k matrix to ~16 GB — twice (for adj
>   and adj_er), OOM-crashing the notebook on 24 GB RAM (this runs before my code).
>   - Fix: Replaced the two task-level calls with a sparse in_out_degree (axis sums; node_degree itself just
>   sums the densified matrix — exactly equivalent). Dropped the connalysis node_degree import, moved the
>   dependency sentinel to ER_model. Small-circuit node_degree uses elsewhere left as-is.
>   - Result: removes the OOM; equivalence validated (match=True).


chr-pok (commented):
Works nicely. Only the example notebook requires some cleanup again ;-)

chr-pok (approved):

PR #844, opened 2026-06-19
circuit customization or registration

jdcourcol:
Add circuit customization & unified async validation for registration

  Adds a POST /declared/circuit/customize endpoint and refactors circuit registration to follow the same async validation pattern: entities are created in draft state
  and transition to active after async validation passes, which then triggers additional asset generation.

  Circuit registration (refactored)

  The existing registration endpoint now follows the same lifecycle pattern as customization:

  1. Extracts the uploaded .tar.gz archive, locates circuit_config.json
  2. Computes metadata (get_circuit_size, get_circuit_properties)
  3. Creates the Circuit entity with lifecycle_status=draft
  4. Uploads the extracted directory as sonata_circuit asset + original archive as compressed_sonata_circuit
  5. Generates sync visualization assets (node_stats, network_stats_a/b, circuit_visualization, simulation_designer_image)
  6. Triggers an async validation job via launch-system

  Previously, registered circuits were immediately active. Now they go through the same validation gate as customized circuits.

  Circuit customization (new)

  POST /declared/circuit/customize lets users create a customized circuit from a parent by uploading override files (edges, HOC, MOD, nodes, node_sets,
  circuit_config) in a single multipart request.

  Upload validations (synchronous, at request time):

  - Edges: valid HDF5, edges group present, required columns (source_node_id, target_node_id, edge_type_id), no NaN/Inf in float columns
  - HOC: .hoc extension, valid template structure (begintemplate X with matching endtemplate X)
  - MOD: .mod extension, contains NEURON block
  - Nodes: valid HDF5, nodes group present
  - Node sets: valid JSON, SONATA expression syntax (attribute filters, compound expressions)
  - HOC↔MOD cross-check: mechanisms referenced via insert in HOC must exist as built-in NEURON mechanisms, in provided MOD files (SUFFIX declarations), or in the
  parent circuit's mechanisms directory
  - New MOD guard: new MOD files (names not in parent) must not contain NET_RECEIVE — new synapse mechanisms are not supported
  - Nodes↔HOC consistency: uploaded HOC files must be referenced by at least one node's model_template

  If any check fails → HTTP 422 with specific error messages. No entity is created.

  Staging and merge (synchronous, at request time):

  1. Stages the parent circuit from EFS (symlinks via entitysdk.stage_circuit)
  2. Overlays the user's overrides into the staged directory (paths resolved using libsonata.CircuitConfig)
  3. Computes circuit metadata
  4. Uploads the complete merged directory as sonata_circuit asset
  5. Generates sync visualization assets
  6. Creates Circuit entity with lifecycle_status=draft, root_circuit_id → parent
  8. Creates derivation link (type=circuit_customization)
  9. Submits async validation job to launch-system

  The customized circuit is immediately browsable — it has a full sonata_circuit directory and visualization assets.

  Async validation (shared by registration & customization)

  A validation task runs that:

  1. Stages the circuit from EFS
  2. Compiles MOD files with nrnivmodl (if present)
  3. Validates HOC loading with bluecellulab.Cell (per template + morphology)
  4. Validates morphology/emodel directory paths exist (respects alternate_morphologies H5 format)
  5. Runs bluepysnap.circuit_validation.validate() — full SONATA conformance
  6. For customizations: checks new populations are not biophysical, content is subset of parent

  Transitions draft → active on success, draft → disqualified on failure.

  Asset generation (triggered after successful validation)

  On validation success, a launch-system callback triggers POST /declared/circuit/{id}/generate-assets which submits an async asset generation job that produces:

  - circuit_connectivity_matrices — connectivity matrix H5 files
  - compressed_sonata_circuit — .tar.gz of the full circuit (skipped if already uploaded at registration)

  This runs for both registered and customized circuits. Asset generation does not affect lifecycle_status — the circuit remains active regardless of whether asset generation succeeds. The job is re-launchable with ?force=true.

  Simulation gate

  The task launch endpoint checks lifecycle_status == active before allowing simulation. Draft or disqualified circuits are rejected with HTTP 422.


obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+        fatal_errors.extend(hoc_errors)
+
+        # bluepysnap structural validation
+        L.info("Running circuit validation on %s", circuit_config_path)
+        snap_errors = circuit_validation.validate(str(circuit_config_path), skip_slow=False)
+        fatal_errors.extend(str(e) for e in snap_errors if e.level == "FATAL")
+        warning_messages.extend(str(e) for e in snap_errors if e.level == "WARNING")
+
+        # Subset checks: morphologies and emodels must exist in parent
+        if is_customization and circuit.root_circuit_id:
```
chr-pok: The parent circuit ID should be used here, not the root (the root is the parent of the whole hierarchy, but not necessarily the direct predecessor). The parent is not an attribute of the circuit entity but given by the "used" entity of the derivation link.

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+            with tempfile.TemporaryDirectory() as parent_tmp:
+                parent_dir = Path(parent_tmp) / "parent"
+                parent_dir.mkdir()
+                try:
+                    parent = db_client.get_entity(
+                        entity_id=circuit.root_circuit_id, entity_type=models.Circuit
```
chr-pok: Same here, should be parent, not root.

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+    component_morph_dir = cfg.get("components", {}).get("morphologies_dir", "")
+    errors = []
+
+    for entry in cfg.get("networks", {}).get("nodes", []):
+        for pop_name, pop_cfg in entry.get("populations", {}).items():
+            if pop_cfg.get("type") == "virtual":
+                continue
+
+            # Skip if alternate_morphologies (H5-based) is defined
+            if pop_cfg.get("alternate_morphologies"):
```
chr-pok: Why do we skip here? Shouldn't we validate all swc/asc/h5/containerized h5 paths that are defined?

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+def _validate_morphology_paths(circuit_config_path: Path) -> list[str]:
+    """Verify that the morphologies_dir referenced in the circuit config exists on disk.
+
+    Resolves per-population morphologies_dir (takes precedence over the component-level
+    default). Fails if the directory is missing or not a directory.
+    Skips populations that use alternate_morphologies (H5-based) instead.
+    """
+    config = libsonata.CircuitConfig.from_file(str(circuit_config_path))
+    cfg = json.loads(config.expanded_json)
+    component_morph_dir = cfg.get("components", {}).get("morphologies_dir", "")
```
chr-pok: Maybe better to rely on SNAP's population.morph (which automatically resolved the global components directory) instead of manually parsing the json. See, e.g., https://github.com/openbraininstitute/obi-one/blob/984ffe178295706d86c329b43906f93c02fbcf99/obi_one/utils/circuit.py#L218

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+
+def _validate_emodel_paths(circuit_config_path: Path) -> list[str]:  # noqa: C901
+    """Check that all HOC template files referenced by biophysical populations exist.
+
+    Resolves biophysical_neuron_models_dir per population (population-level override
+    takes precedence over the component-level default).
+    """
+    config = libsonata.CircuitConfig.from_file(str(circuit_config_path))
+    cfg = json.loads(config.expanded_json)
+    component_hoc_dir = cfg.get("components", {}).get("biophysical_neuron_models_dir", "")
```
chr-pok: Again, maybe better to reuse the existing implementation from SNAP's populations, in which case there is no need to read the global `component_hoc_dir` here. Even if defined globally under components, the path will show up under population.config.

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+                        f" not found in {hoc_dir}"
+                    )
+
+    return errors
+
+
+def _read_model_templates(nodes_file: str, pop_name: str) -> set[str]:
+    """Read unique model_template values for a population from its H5 file."""
+    try:
+        with h5py.File(nodes_file, "r") as f:
```
chr-pok: Again, why parsing the h5 file directly instead of relying on SNAP functionality? I think this is much more difficult to maintain in case of changes in the standard.

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+            " regenerate with brainbuilder or remove manually"
+        )
+    L.warning(msg)
+    return [msg]
+
+
+def _get_population_sizes(cfg: dict) -> dict[str, int]:
+    """Build per-population node count from the H5 files referenced in the config."""
+    pop_sizes: dict[str, int] = {}
+    for entry in cfg.get("networks", {}).get("nodes", []):
```
chr-pok: Again, I would rely on SNAP population.size instead of parsing the nodes file directly.

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+
+    if not stale_populations:
+        return []
+
+    detail = "; ".join(stale_populations)
+    msg = f"id_mapping.json is stale after nodes replacement ({detail})"
+    if id_mapping_path.is_symlink():
+        id_mapping_path.unlink()
+        msg += " — file removed from the circuit (regenerate with brainbuilder if needed)"
+    else:
```
chr-pok: I don't understand what this distinction with symlinks is about? Shouldn't this file be removed from the circuit asset in any case?

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+
+    return errors
+
+
+def _collect_hoc_files(cfg: dict, working_dir: Path) -> list[Path]:
+    """Collect all HOC files from per-population biophysical_neuron_models_dir."""
+    component_hoc_dir = cfg.get("components", {}).get("biophysical_neuron_models_dir", "")
+
+    hoc_dirs: list[Path] = []
+    for entry in cfg.get("networks", {}).get("nodes", []):
```
chr-pok: Again, why not using SNAP functionality here?

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+        L.info("Circuit %s lifecycle_status -> %s", circuit_id, status)
+    except Exception:  # noqa: BLE001
+        L.warning("Failed to update lifecycle_status for circuit %s", circuit_id, exc_info=True)
+
+
+def _find_mod_dir(circuit_config_path: Path) -> Path | None:
+    """Find the mechanisms directory from circuit_config.json using libsonata."""
+    config = libsonata.CircuitConfig.from_file(str(circuit_config_path))
+    cfg = json.loads(config.expanded_json)
+    mod_dir = cfg.get("components", {}).get("mechanisms_dir", "")
```
chr-pok: Better to rely on SNAP's "mechanisms_dir" in population.config. "mechanisms_dir" may be defined either globally under components or under a specific population (see https://sonata-extension.readthedocs.io/en/latest/sonata_config.html#components) which is resolved automatically by SNAP.

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+        if pop_name in entry.get("populations", {}):
+            return entry["populations"][pop_name]
+    return None
+
+
+def _resolve_node_morphology(pop: NodePopulation, node_id: int) -> Path | None:
+    """Resolve the morphology path for a node, trying swc then asc extensions."""
+    for ext in ("swc", "asc", "h5"):
+        try:
+            morph_path = pop.morph.get_morphology_path(node_id, extension=ext)
```
chr-pok, line 593: Hmm, if I am not mistaken, this function name does not exist in SNAP. It should be either `.get` or `.get_filepath` (?). BUT: I think the second one will fail with a `BluepySnapError` for h5 morphology containers (see https://github.com/openbraininstitute/snap/blob/727feefa64df98ed18b589054321cb0069fdb049/bluepysnap/morph.py#L71).

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+    else:
+        group.create_dataset(name, data=arr)
+
+
+# ---------------------------------------------------------------------------
+# Subset checks: customized circuit must not introduce new morphologies or
+# emodels that don't exist in the parent (adopted from PR #829).
+# ---------------------------------------------------------------------------
+
+_NON_BIOPHYSICAL_POPULATION_TYPES = {"virtual", "point_neuron"}
```
chr-pok: Should rely on `TYPES_OF_VIRTUAL_NODES` and `TYPES_OF_POINT_NODES` defined in circuit_metrics.py instead of using hard-coded constants: https://github.com/openbraininstitute/obi-one/blob/ab678dc0a7e02798e3943c0ae5d8fb0b3887a289/obi_one/scientific/library/circuit_metrics.py#L33

obi_one/scientific/tasks/circuit_validation/task.py
```diff
@@ -0,0 +1,828 @@
...
+
+def _check_new_populations_not_biophysical(
+    child_config_path: Path, parent_config_path: Path
+) -> list[str]:
+    """Reject new node populations that are biophysical.
+
+    New populations may only be virtual or point neurons — biophysical populations
+    require morphologies and HOC files that must come from the parent.
+    """
+    child_cfg = json.loads(libsonata.CircuitConfig.from_file(str(child_config_path)).expanded_json)
```
chr-pok: Again, maybe safer to rely on SNAP instead of parsing the dict directly.

obi_one/scientific/tasks/circuit_validation/task.py
chr-pok: Actually, in case of point neurons we should add another check that the type of point neurons matches the target simulator. E.g., we should not allow adding a "brian2_point" population to a circuit that is simulated with NEURON.

obi_one/utils/circuit_customization/staging.py
```diff
@@ -0,0 +1,280 @@
...
+        L.info("Replaced circuit_config.json with override")
+
+    # 4. Remove network files from the parent that the override config no longer references
+    if circuit_config_override:
+        _remove_stale_network_files(circuit_dir, circuit_config_path, parent_config)
+
+    return circuit_config_path
+
+
+def _apply_file_overrides(
```
chr-pok: If I understand correctly, this is only for overriding existing node/edge populations. What about adding new populations that are referenced in the overridden circuit config?

launch_scripts/launch_circuit_validation/main.py
```diff
@@ -0,0 +1,102 @@
...
+            project_context=project_context,
+            token_manager=token_manager,
+            local_store=LocalAssetStore(prefix=local_store_prefix),
+        )
+
+        circuit = db_client.get_entity(entity_id=circuit_id, entity_type=models.Circuit)
+        result = run_circuit_validation(
+            db_client=db_client,
+            circuit_id=circuit_id,
+            is_customization=circuit.root_circuit_id is not None,
```
chr-pok: `is_customization` should be given by checking if a derivation link of type "circuit_customization" exists.

tests/app/endpoints/test_task_circuit_gate.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class TestCheckCircuitIsActive:
+    def test_active_circuit_passes(self):
+        """Active customized circuit should not raise."""
+        db_client = MagicMock()
+        sim = MagicMock()
+        sim.entity_id = uuid4()
+        circuit = MagicMock()
+        circuit.root_circuit_id = uuid4()  # customized circuit
```
chr-pok: Root circuit ID is not relevant for customization. Only derivation of type "circuit_customization" is.

tests/app/endpoints/test_task_circuit_gate.py
```diff
@@ -0,0 +1,67 @@
...
+        db_client.get_entity.side_effect = [sim, circuit]
+        _check_circuit_is_active(db_client, uuid4())  # should not raise
+
+    def test_draft_circuit_raises(self):
+        """Draft customized circuit should be rejected."""
+        db_client = MagicMock()
+        sim = MagicMock()
+        sim.entity_id = uuid4()
+        circuit = MagicMock()
+        circuit.root_circuit_id = uuid4()  # customized circuit
```
chr-pok: Again, root circuit ID is not relevant for customization. Only derivation of type "circuit_customization" is.

tests/app/endpoints/test_task_circuit_gate.py
```diff
@@ -0,0 +1,67 @@
...
+        with pytest.raises(ApiError, match="not ready for simulation"):
+            _check_circuit_is_active(db_client, uuid4())
+
+    def test_disqualified_circuit_raises(self):
+        """Disqualified circuit should be rejected."""
+        db_client = MagicMock()
+        sim = MagicMock()
+        sim.entity_id = uuid4()
+        circuit = MagicMock()
+        circuit.root_circuit_id = uuid4()
```
chr-pok: Same here

tests/app/endpoints/test_task_circuit_gate.py
```diff
@@ -0,0 +1,67 @@
...
+        with pytest.raises(ApiError, match="not ready for simulation"):
+            _check_circuit_is_active(db_client, uuid4())
+
+    def test_non_customized_circuit_passes(self):
+        """Circuit without root_circuit_id (not customized) should pass."""
+        db_client = MagicMock()
+        sim = MagicMock()
+        sim.entity_id = uuid4()
+        circuit = MagicMock()
+        circuit.root_circuit_id = None  # not a customized circuit
```
chr-pok: Same here

app/endpoints/task.py
```diff
@@ -25,6 +28,30 @@
...
+
+def _check_circuit_is_active(db_client: entitysdk.client.Client, config_id: UUID) -> None:
+    """Block simulation launch if the circuit is a customized circuit not yet validated."""
+    try:
+        simulation = db_client.get_entity(entity_id=config_id, entity_type=models.Simulation)
+        circuit = db_client.get_entity(entity_id=simulation.entity_id, entity_type=models.Circuit)
+    except entitysdk.exception.EntitySDKError:
+        return
+
+    if circuit.root_circuit_id is None:
```
chr-pok: Root circuit ID is not relevant for customization. Only derivation of type "circuit_customization" is.

chr-pok: But why is this even relevant? I would say that all circuit that are not "active" for whatever reason should not be simulatable. This also applies to new user-uploaded circuits which must undergo validation first.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+    """Validate a dict-based nodeset expression."""
+    VALID_OPERATORS = {"$regex", "$gt", "$lt", "$gte", "$lte"}  # noqa: N806
+
+    for k, v in expr.items():
+        if k == "population":
+            if not isinstance(v, str | list):
+                msg = f"'{path.name}' nodeset '{key}': 'population' value must be a string or list"
+                raise NodeSetsValidationError(msg)
+        elif k == "node_id":
+            if not isinstance(v, list | int):
```
chr-pok: I think it must be a list, single ints are not supported if I am not mistaken: https://sonata-extension.readthedocs.io/en/latest/sonata_nodeset.html

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+                msg = (
+                    f"'{path.name}' nodeset '{key}': 'node_id' value must be an int or list of ints"
+                )
+                raise NodeSetsValidationError(msg)
+        elif k in VALID_OPERATORS:
+            pass  # operator values are unconstrained at this layer
+        elif not isinstance(v, str | int | float | list):
```
chr-pok: Looks like `bool` should be also supported: https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#node-sets-file

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+                f"'{path.name}' nodeset '{key}': "
+                f"attribute filter value for '{k}' must be a scalar or list"
+            )
+            raise NodeSetsValidationError(msg)
+
+
+def _validate_nodeset_list(path: Path, key: str, expr: list) -> None:
+    """Validate a list-based compound nodeset expression."""
+    for item in expr:
+        if not isinstance(item, str | dict):
```
chr-pok: If I am not mistaken, compound node sets must be lists of `str` only, i.e., no `dict` supported.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+
+    try:
+        # Stage parent config to get mechanism dir path (EFS-backed, fast)
+        from entitysdk.staging.circuit import stage_circuit  # noqa: PLC0415
+
+        with tempfile.TemporaryDirectory() as ptmp:
+            config_path = stage_circuit(db_client, model=parent, output_dir=Path(ptmp))
+            config = libsonata.CircuitConfig.from_file(str(config_path))
+            cfg = json.loads(config.expanded_json)
+            mech_dir = cfg.get("components", {}).get("mechanisms_dir", "")
```
chr-pok: "mechanisms_dir" can be defined at the population level as well. May use SNAP's population.config to resolve them for each population (also works if defined under global components).

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+                f"/{circuit_id}/generate-assets?force=true"
+            ),
+            "method": "POST",
+        },
+    }
+    job_data = {
+        "code": {
+            "type": "python_repository",
+            "location": settings.OBI_ONE_REPO,
+            "ref": f"commit:{settings.COMMIT_SHA}" if settings.COMMIT_SHA else "HEAD",
```
chr-pok: Should use an app tag (as we have for other tasks) once deployed.

launch_scripts/launch_circuit_validation/main.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+        db_client = Client(
+            environment=deployment,
+            project_context=project_context,
+            token_manager=token_manager,
+            local_store=LocalAssetStore(prefix=local_store_prefix),
+        )
+
+        circuit = db_client.get_entity(entity_id=circuit_id, entity_type=models.Circuit)
+        result = run_circuit_validation(
```
chr-pok, line 105: Should we check that the circuit is really in "draft" status? And skip otherwise? Could even introduce an overwrite flag when manually triggering the validation.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+                "inconsistent staging. Provide the config override alone, or provide file "
+                "overrides without replacing the config."
+            ),
+        )
+
+    pop_map = _parse_population_manifest(emodel_population_manifest)
+
+    # 1. Fetch parent circuit
+    try:
+        parent = db_client.get_entity(entity_id=parent_circuit_id, entity_type=models.Circuit)
```
chr-pok, line 641: Should check that parent circuit is in "active" state to prevent creating customizations of "draft" circuits.

launch_scripts/launch_circuit_asset_generation/main.py
```diff
@@ -0,0 +1,97 @@
...
+            staged_dir.mkdir()
+            circuit_config_path = stage_circuit(db_client, model=circuit, output_dir=staged_dir)
+
+            # Determine edge population for matrix extraction
+            from obi_one.scientific.library.circuit import Circuit as OBICircuit
+
+            c = OBICircuit(name=circuit.name, path=str(circuit_config_path))
+            edge_pop = c.default_edge_population_name if c.sonata_circuit.edges.population_names else None
+
+            generate_additional_circuit_assets(
```
chr-pok, line 86: In `generate_additional_circuit_assets`, we need to handle the case where the additional assets already exist, i.e., deleting and re-generate the assets.

Or skipping the ones that already exist, e.g., we should not re-compress the circuit if only one plot needs to be re-generated. Especially if the compressed file is already provided by the user upon upload.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+            "timelimit": "00:30",
+            "compute_cell": "local",
+        },
+        "inputs": [
+            f"--circuit_id {circuit_id}",
+            f"--virtual_lab_id {virtual_lab_id}",
+            f"--project_id {project_id}",
+        ],
+        "project_id": str(project_id),
+        "callbacks": [asset_gen_callback],
```
chr-pok: Do we need a specific failure callback (in case of timeout, OOM, etc.)? Currently, if I understand correctly, the circuit will just remain in "draft" status in this case (which I think is fine).

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+        raise HTTPException(
+            status_code=409,
+            detail=f"Circuit lifecycle_status is '{status}'. "
+            "Asset generation requires an active circuit.",
+        )
+
+    # Check if assets already exist (unless force)
+    if not force:
+        existing_labels = {a.label for a in (circuit.assets or [])}
+        needed = {"compressed_sonata_circuit", "circuit_connectivity_matrices"}
```
chr-pok, line 217: There are more assers needed: "node_stats", "network_stats_a", "network_stats_b" (basic connectivity plots)

May also consider: "circuit_visualization" and "simulation_designer_image" but could leave them optional

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+                f"/{circuit_id}/generate-assets?force=true"
+            ),
+            "method": "POST",
+        },
+    }
+    job_data = {
+        "code": {
+            "type": "python_repository",
+            "location": settings.OBI_ONE_REPO,
+            "ref": "commit:1ed9e654c7f51f971b5f1162b89d5766cbbb9d43",  # TODO: use tag after merge
```
chr-pok: Should use tag once deployed.

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+    if tarfile.is_tarfile(archive_path):
+        with tarfile.open(archive_path, "r:*") as tar:
+            tar.extractall(path=extract_dir, filter="data")
+    else:
+        raise HTTPException(status_code=422, detail="circuit_archive must be a .tar.gz file")
+
+    return extract_dir
+
+
+def _trigger_validation_task(
```
chr-pok: Isn't this a duplication of the same function in circuit_customization.py?

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+
+    return {"circuit_id": str(circuit_id), "status": "generation_triggered"}
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _extract_archive(archive_path: Path, dest: Path) -> Path:
```
chr-pok: We have a function for that already in utils/io.py: https://github.com/openbraininstitute/obi-one/blob/ab678dc0a7e02798e3943c0ae5d8fb0b3887a289/obi_one/utils/io.py#L20

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+            status_code=422, detail="derivation_type required when parent_circuit_id is set"
+        )
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp = Path(tmp_dir)
+
+        # 1. Extract archive
+        archive_path = tmp / circuit_archive.filename  # ty:ignore[unsupported-operator]
+        archive_path.write_bytes(circuit_archive.file.read())
+        circuit_dir = _extract_archive(archive_path, tmp)
```
chr-pok: Duplication of code. Should use `extract_tar_gz`: https://github.com/openbraininstitute/obi-one/blob/ab678dc0a7e02798e3943c0ae5d8fb0b3887a289/obi_one/utils/io.py#L20

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+        )
+
+        # 4. Resolve entities
+        brain_region = db_client.get_entity(
+            entity_id=brain_region_id, entity_type=models.BrainRegion
+        )
+        subject = db_client.get_entity(entity_id=subject_id, entity_type=models.Subject)
+
+        # 5. Create circuit entity
+        circuit_model = models.Circuit(
```
chr-pok: Need to set the lifecycle status to "draft".

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+from obi_one.scientific.library.circuit import Circuit as OBICircuit
+from obi_one.utils.circuit import get_circuit_properties, get_circuit_size
+
+L = logging.getLogger(__name__)
+
+router = APIRouter(prefix="/declared", tags=["declared"], dependencies=[Depends(user_verified)])
+
+
+@router.post("/circuit/register")
+def register_circuit_endpoint(  # noqa: PLR0913, PLR0917, PLR0914
```
chr-pok: Again, this endpoint duplicates a lot of the code from `register_circuit`: https://github.com/openbraininstitute/obi-one/blob/ab678dc0a7e02798e3943c0ae5d8fb0b3887a289/obi_one/utils/circuit_registration/register.py#L90

I think we should refactor the existing unified `register_circuit` function to follow the new approach: Register as "draft" -> Validate -> Generate additional assets

And use it consistently for all tasks that register circuits. Instead of having different versions of the same functionality.

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+        # 6b. Upload original archive as compressed_sonata_circuit (skips compression stage)
+        db_client.upload_file(
+            entity_id=registered.id,  # ty:ignore[invalid-argument-type]
+            entity_type=models.Circuit,
+            file_path=archive_path,
+            file_content_type=ContentType.application_gzip,
+            asset_label=AssetLabel.compressed_sonata_circuit,
+        )
+
+        # 6c. Generate and register stats + visualization assets
```
chr-pok: All this should come after validation, i.e., handled by the additional asset generation callback.

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+        "status": "draft",
+        "number_neurons": int(number_neurons),
+        "number_synapses": int(number_synapses),
+        "number_connections": int(number_connections) if number_connections is not None else None,
+        "scale": str(scale),
+    }
+
+
+@router.post("/circuit/{circuit_id}/generate-assets")
+def generate_assets_endpoint(
```
chr-pok, line 191: Would be useful to have more fine-grained endpoints that will be triggered on success (since they are partly independent of each other):
- Validation -> Compression
- Validation -> Matrix extraction -> Stats plots

(But this could come later.)

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+    ls_client: LaunchSystemClientDep,
+    db_client: Annotated[entitysdk.client.Client, Depends(get_client)],
+    name: Annotated[str, Form()],
+    description: Annotated[str, Form()],
+    brain_region_id: Annotated[UUID, Form()],
+    subject_id: Annotated[UUID, Form()],
+    build_category: Annotated[str, Form()],
+    target_simulator: Annotated[str, Form()],
+    circuit_archive: Annotated[UploadFile, File()],
+    scale_override: Annotated[str | None, Form()] = None,  # noqa: ARG001
```
chr-pok: Currently not used anywhere.

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+            number_synapses=number_synapses,
+            number_connections=number_connections,
+            has_morphologies=has_morphologies,
+            has_point_neurons=has_point_neurons,
+            has_electrical_cell_models=has_electrical_cell_models,
+            has_spines=has_spines,
+            scale=scale,
+            build_category=build_category,  # ty:ignore[invalid-argument-type]
+            target_simulator=target_simulator,  # ty:ignore[invalid-argument-type]
+            root_circuit_id=parent_circuit_id,
```
chr-pok: Root and parent circuits are different things. The parent circuit needs to be set by creating a derivation link.

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+    name: Annotated[str, Form()],
+    description: Annotated[str, Form()],
+    brain_region_id: Annotated[UUID, Form()],
+    subject_id: Annotated[UUID, Form()],
+    build_category: Annotated[str, Form()],
+    target_simulator: Annotated[str, Form()],
+    circuit_archive: Annotated[UploadFile, File()],
+    scale_override: Annotated[str | None, Form()] = None,  # noqa: ARG001
+    parent_circuit_id: Annotated[UUID | None, Form()] = None,
+    derivation_type: Annotated[str | None, Form()] = None,
```
chr-pok, line 39: Currently not used. Should create an actual derivation link.

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+    atlas_id: Annotated[UUID | None, Form()] = None,
+    license_id: Annotated[UUID | None, Form()] = None,
```
chr-pok, line 41: Currently not used anywhere.

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+    contact_email: Annotated[str | None, Form()] = None,  # noqa: ARG001
```
chr-pok: Currently not used anywhere.

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+    authorized_public: Annotated[bool, Form()] = False,  # noqa: FBT002
```
chr-pok, line 43: Currently not used anywhere.

app/endpoints/circuit_registration.py
chr-pok: Also, derivation, contributions and publication links should be added if this is supposed to be a generally usable registration endpoint (as in the existing `register_circuit`). I think these links should be created before validation in "draft" status already.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+        number_synapses=number_synapses,
+        number_connections=number_connections,
+        has_morphologies=has_morphologies,
+        has_point_neurons=has_point_neurons,
+        has_electrical_cell_models=has_electrical_cell_models,
+        has_spines=has_spines,
+        scale=scale,
+        build_category=parent.build_category,
+        target_simulator=parent.target_simulator,
+        root_circuit_id=parent.id,
```
chr-pok: Root and parent are not necessarily the same! A derivation link should be created for the parent.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+    merged_files = {p.relative_to(staged_dir): p for p in staged_dir.rglob("*") if p.is_file()}
+    db_client.upload_directory(
+        entity_id=registered.id,  # ty:ignore[invalid-argument-type]
+        entity_type=models.Circuit,
+        name="sonata_circuit",
+        paths=merged_files,  # ty:ignore[invalid-argument-type]
+        label=AssetLabel.sonata_circuit,
+    )
+
+    # Generate and register stats + visualization assets
```
chr-pok: Additional assets should not be generated here but after validation, i.e., by the asset generation callback.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+        scale, number_neurons, number_synapses, number_connections = get_circuit_size(c)
+        has_morphologies, has_point_neurons, has_electrical_cell_models, has_spines = (
+            get_circuit_properties(c)
+        )
+    except (OSError, ValueError, RuntimeError) as e:
+        raise HTTPException(
+            status_code=422, detail=f"Failed to compute circuit metadata: {e}"
+        ) from e
+
+    circuit_model = models.Circuit(
```
chr-pok, line 511: This duplicates a lot of the code from `register_circuit`: https://github.com/openbraininstitute/obi-one/blob/ab678dc0a7e02798e3943c0ae5d8fb0b3887a289/obi_one/utils/circuit_registration/register.py#L90

I think we should refactor the existing unified `register_circuit` function to follow the new approach: Register as "draft" -> Validate -> Generate additional assets

And use it consistently for all tasks that register circuits. Instead of having different versions of the same functionality.

For customized circuits, only the uncompressed circuit folder asset will exist, in which case the compression task will be forced to run.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+
+    try:
+        from obi_one.scientific.library.circuit import Circuit as OBICircuit  # noqa: PLC0415
+        from obi_one.utils.circuit import (  # noqa: PLC0415
+            get_circuit_properties,
+            get_circuit_size,
+        )
+
+        c = OBICircuit(name=name, path=str(merged_config))
+        scale, number_neurons, number_synapses, number_connections = get_circuit_size(c)
```
chr-pok: Duplication of unified circuit registration.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,799 @@
...
+            hoc_paths=hoc_paths,
+            mod_paths=mod_paths,
+            node_paths=node_paths,
+            node_sets_path=node_sets_path,
+            cfg_path=cfg_path,
+            pop_map=pop_map,
+        )
+
+    # 5. Trigger async validation task via launch-system
+    _trigger_validation_task(
```
chr-pok: Should be triggered as part of the unified circuit registration functionality whenever a new circuit is registered.

app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,350 @@
...
+
+            generate_sim_designer_image_asset(
+                plot_dir=plot_dir,
+                output_dir=viz_dir,
+                client=db_client,
+                circuit_entity=registered,
+            )
+
+    # 7. Trigger validation task
+    _trigger_validation_task(
```
chr-pok: Should be triggered as part of the unified circuit registration functionality whenever a new circuit is registered.

chr-pok (commented):
My first batch of comments. Will look more into it...

Most importantly, seems like the circuit registration code needs to be refactored and unified more, as there are a lot of code duplications, inconsistencies, and parts missing (like creating derivation links etc.). And it does not follow the agreed pattern: Registration as "draft" -> Validation -> Additional assets

Also, there is an important distinction between root and parent circuit to be resolved.

PR #851, opened 2026-06-23
Brian2 stimulus limit

james-isbister:


chr-pok (approved):
Looks ok for a quick fix

PR #853, opened 2026-06-23
Fixed a potential indexing error with EM Synapse Mapping.

MWolfR:
It occured because I was increasing the size of a KDTree query until a hit was achieved or an arbitrary maximum size was reached. Usually, a hit is achieved long before the maximum size is reached. However, in some circumstances this allowed the size of the query to exceed the number of elements in the KDTree. In that case, invalid indices were returned.

Now, the maximum size is set to the size of the KDTree.


jplanasc (approved):
LGTM

PR #856, opened 2026-06-24
Adding point population types

chr-pok:
Adding support to all new point neuron types


james-isbister (approved):
LGTM. Maybe we're ok to merge it now before we start testing things? It would be one less thing to think about

PR #857, opened 2026-06-24
initial flywire spike replay

mgeplf:


obi_one/scientific/library/simulation/brian2/simulate_brian2.py
```diff
@@ -31,9 +31,12 @@
 REQUIRED_PATH = click.Path(exists=True, readable=True, dir_okay=False, resolve_path=True)
 L = logging.getLogger(__name__)
 KNOWN_UNITS = {u for u in dir(brian2.units) if not u.startswith("_")}
+POPULATION_NAME = "drosophila"
```
james-isbister, line 34: Could we remove the dependence of the simulation script on this constant?

james-isbister (commented):
LGTM. Just one comment

obi_one/scientific/library/simulation/brian2/simulate_brian2.py
mgeplf: sure, but perhaps in a different PR; at the moment it's tied to the population in a number of places and wasn't be handled right; I'm just making it a single point so it's easier to fix after.

james-isbister (approved):

PR #860, opened 2026-06-25
By default, brain2 uses `seconds` for spike times, but `ms` was expected

mgeplf:


james-isbister (commented):
I find the names of the index variables a bit intransparent (i.e. v, vs, k)

And the goal of dividing through by ms is to convert back to seconds?

james-isbister (approved):

PR #861, opened 2026-06-26
Combined neuron sets

james-isbister:


chr-pok (approved):
LGTM! Some remaining issues to be addressed separately, as discussed.

PR #815, opened 2026-06-10
Skip morphology metrics containing NaN values

ilkilic:
Fix: https://github.com/openbraininstitute/prod-explore-functionality/issues/500

PR #870, opened 2026-06-30
Drop incomplete morphology aggregate metrics

ilkilic:
This follows up on https://github.com/openbraininstitute/obi-one/pull/815.

That change handled NaN values coming from the raw NeuroM results. This PR handles an additional output-side case where an aggregate metric can still be serialized with only part of the expected statistics, for example minimum and median present, but maximum, mean, and standard_deviation missing. For aggregate morphometrics, partial results are misleading, so this change drops the whole aggregate metric unless all expected statistic values are valid.

### DIFF
--- a/obi_one/scientific/library/morphology_measurement_annotation.py
+++ b/obi_one/scientific/library/morphology_measurement_annotation.py
@@ -335,31 +335,40 @@ def _is_valid_measurement_value(value: Any) -> bool:
     """Return True when a measurement item value should be kept."""
     if value is None:
         return False
-    return not (isinstance(value, (float, np.floating)) and np.isnan(value))
+    return not (isinstance(value, (float, np.floating)) and not np.isfinite(value))
 
 
 def _filter_valid_measurement_kinds(
     measurement_kinds: list[dict[str, Any]],
 ) -> list[dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


obi_one/scientific/library/morphology_measurement_annotation.py
```diff
@@ -330,17 +331,51 @@ def fill_json(
...
     neuron = nm.load_morphology(str(morphology_path))
     results_dict = build_results_dict(get_morphology_analysis_dict(), neuron)
     filled = fill_json(copy.deepcopy(get_morphology_template()), results_dict, entity_id="temp_id")
     measurement_kinds = filled["data"][0]["measurement_kinds"]
-    return [
-        mk
-        for mk in measurement_kinds
-        if any(mi.get("value") is not None for mi in mk.get("measurement_items", []))
-    ]
+    return cast("list[MeasurementKind]", _filter_valid_measurement_kinds(measurement_kinds))
```
GianlucaFicarelli, line 378: The cast to `list[MeasurementKind]` doesn't look right, since _filter_valid_measurement_kinds returns a list of dicts.
However, it's not changed in this PR, so it could be fixed separately.

GianlucaFicarelli (approved):

PR #872, opened 2026-06-30
Retry transient CAVEClient materialization errors in EM synapse mapping

pgetta:
## Context

Addresses openbraininstitute/prod-build-circuit#45 — the CAVEClient materialization engine intermittently returns **503** errors during the EM synapse mapping task (notably via `materialize.get_tables()` / `materialize.get_versions()`), failing the whole run with a raw `requests.HTTPError`.

caveclient already retries 502/503/504 at the HTTP-session layer, but with weak defaults (`max_retries=3`, `backoff_factor=0.1` → ~0.6 s of total backoff), which isn't enough to outlast a transient outage.

## Changes

- **Tune caveclient's built-in retry** (`obi_one/config.py`): new env-overridable knobs on `CaveClientConfig` — `max_retries=8`, `retry_backoff_factor=0.5`, `retry_backoff_max=120.0`, `retry_status_forcelist=(429, 502, 503, 504)`. With these, the backoff schedule is 0/1/2/4/8/16/32/64 s (~127 s total) before giving up.
- **Apply the config once** via `_configure_caveclient_retries()` at module import in `em_dataset_from_id.py`. Because every materialize call shares the client's HTTP session, this covers all of them (`synapse_query`, `query_table`, `get_table_metadata`, `get_tables`, `get_versions`, `viewer_resolution`).
- **Graceful failure**: a `_graceful_materialize_errors` decorator on the public materialize-backed methods converts an exhausted `requests` failure into a clean `OBIONEError` with an actionable message instead of a raw traceback. Non-request errors propagate unchanged.
- Tests for retry configuration and graceful-failure handling.

Client *initialization* is intentionally **not** wrapped — per discussion on the ticket, the materialization-engine calls are the issue.

### DIFF
--- a/obi_one/scientific/from_id/em_dataset_from_id.py
+++ b/obi_one/scientific/from_id/em_dataset_from_id.py
@@ -1,17 +1,52 @@
+import functools
+from collections.abc import Callable
 from typing import ClassVar
 
 import numpy  # NOQA: ICN001
 import pandas  # NOQA: ICN001
-from caveclient import CAVEclient
+import requests
+from caveclient import CAVEclient, set_session_defaults
 from entitysdk import Client
 from entitysdk.models import EMDenseReconstructionDataset
 from entitysdk.models.entity import Entity
 from pydantic import PrivateAttr
 
+from obi_one.config import settings
 from obi_one.core.entity_from_id import EntityFromID
+from obi_one.core.exception import OBIONEError
 
 _C_P_LOCS = ["synapse_x", "synapse_y", "synapse_z"]
 _NM_to_UM = 1e-3
+
+
+def _graceful_materialize_errors[T](func: Callable[..., T]) -> Callable[..., T]:
+    """Convert exhausted CAVEClient request failures into a clean OBIONEError.
+
+    The CAVEClient materialization engine intermittently returns transient errors
+    (e.g. 503). caveclient retries these at the HTTP-session layer (see
+    ``_make_cave_client``); once retries are exhausted it raises a ``requests``
+    exception. This decorator translates that into an actionable ``OBIONEError``
+    so the task fails gracefully rather than with a raw traceback.
+
+    Args:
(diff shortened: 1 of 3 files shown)


james-isbister (commented):
LGTM. But I'll let @GianlucaFicarelli take a look and approve

obi_one/scientific/from_id/em_dataset_from_id.py
```diff
@@ -1,25 +1,82 @@
...
+    cfg = settings.cave_client_config
+    set_session_defaults(
+        max_retries=cfg.max_retries,
+        backoff_factor=cfg.retry_backoff_factor,
+        backoff_max=cfg.retry_backoff_max,
+        status_forcelist=cfg.retry_status_forcelist,
+    )
+
+
+_configure_caveclient_retries()
```
GianlucaFicarelli, line 70: Changing the settings used by CAVEclient globally can be problematic if obi-one is imported in other code that uses CAVEclient on its own, but it seems that `backoff_factor`, `backoff_max` and `status_forcelist` can be set only in this way, while `max_retries` could be set also when the client is instantiated.
An alternative could be to use an external library like [tenacity](https://github.com/jd/tenacity), or implement the retry logic as in [entitysdk](https://github.com/openbraininstitute/entitysdk/blob/8d8e38d0e07a50e4d1bfbe93cc1ca7706a460430/src/entitysdk/utils/execution.py), but it's probably not worth the additional complexity.

GianlucaFicarelli (approved):
I added a comment about global settings but LGTM.

obi_one/scientific/from_id/em_dataset_from_id.py
pgetta: Yeah, thank you Gianluca. I was considering tenacity as well, but probably the current implementation is the simplest and is enough.

If the future usage confirms the opposite - we can apply more complex solution.

PR #874, opened 2026-06-30
Fix morphometrics measurement kind typing

ilkilic:
This removes the incorrect cast from morphometrics output (https://github.com/openbraininstitute/obi-one/pull/870#discussion_r3497420921). The computation and preview paths now return the dict payload they actually produce and registration converts those payloads into MeasurementKind models before creating the annotation.


GianlucaFicarelli (approved):
LGTM

PR #875, opened 2026-06-30
brian2: create empty spikes file when no spikes

mgeplf:
* update LE to a commit that also creates an empty spike file


james-isbister (approved):
LGTM but we shouldn't merge to main until after the push to production

PR #878, opened 2026-07-01
Add prototype Build Synaptome form

ilkilic:
Includes:
- BuildSynaptomeScanConfig / BuildSynaptomeSingleConfig
- ME-model context
- afferent synapse groups
- per-group morphology-location placement strategy
- reused synaptic physiology model blocks
- reused distribution blocks

This does not implement SONATA generation, backend build execution or registration logic.


obi_one/scientific/tasks/build_synaptome.py
```diff
@@ -0,0 +1,154 @@
...
+class BlockGroup(StrEnum):
+    """Block groups for the Build Synaptome form."""
+
+    INFO = "Info"
+    ME_MODEL = "ME-model"
+    SYNAPTIC_PHYSIOLOGY = "Synaptic physiology"
+    SYNAPSE_GROUPS = "Synapse groups"
+
+
+class SynapseGroup(Block):
```
james-isbister: This is definitely different in function to the [SynapticModelAssigner blocks](https://github.com/openbraininstitute/obi-one/tree/main/obi_one/scientific/blocks/synaptic_model_assigners) which assign synaptic models to existing synapses in a circuit

I think we could be more precise in the name of this block type though, as we will probably have other blocks for selecting subsets of synapses on existing circuits which this could be confused with.

Maybe SynapticModelPlacer?

obi_one/scientific/tasks/build_synaptome.py
```diff
@@ -0,0 +1,154 @@
...
+    """Incoming synapse group for a single-cell synaptome build."""
+
+    group_name: str = Field(
```
james-isbister: I think we could use the name of this new block type? Also where would this name be used?

obi_one/scientific/tasks/build_synaptome.py
```diff
@@ -0,0 +1,154 @@
...
+            "Existing morphology-location block used to place this group's incoming synapses. "
+            "The number of locations corresponds to the number of synapses."
+        ),
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.BLOCK_UNION,
+        },
+    )
+
+
+class BuildSynaptomeScanConfig(ScanConfig):
```
james-isbister: Personally I would avoid using the term synaptome in obi-one and entitycore, but its personal preference!

MEModelSynapticModelPlacementScanConfig, SingleNeuronSynapticModelPlacementScanConfig or SynaptomeSynapticModelPlacementScanConfig might be more precise names that allow us to expand to other scales in future?

james-isbister (commented):
LGTM, just a few comments mostly on naming

obi_one/scientific/tasks/build_synaptome.py
ilkilic: Good point, I renamed to `SynapticModelPlacer`.

obi_one/scientific/tasks/build_synaptome.py
ilkilic: It wasn't used anywhere downstream, removed.

obi_one/scientific/tasks/build_synaptome.py
ilkilic: Agreed, renamed to `MEModelSynapticModelPlacementScanConfig`.

obi_one/scientific/tasks/build_synaptome.py
```diff
@@ -0,0 +1,148 @@
...
+
+    synaptic_model: SynapticModelReference = Field(
+        title="Synaptic model",
+        description="Synaptic physiology model assigned to this incoming synapse group.",
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.REFERENCE,
+            SchemaKey.REFERENCE_TYPES: [SynapticModelReference.__name__],
+        },
+    )
+    placement_strategy: MorphologyLocationUnion = Field(
```
james-isbister: At the momemnt block unions are supported at the root level of the config, not within blocks.

So this might be better as MorphologyLocationsBlockReference with a block_dictionary at the root level?

ilkilic: Yes, that was also my intention. I have updated it to use a root-level morphology locations block dictionary and reference it from each synapse group.

obi_one/scientific/tasks/build_synaptome.py
```diff
@@ -0,0 +1,168 @@
...
+    INFO = "Info"
+    ME_MODEL = "ME-model"
+    SYNAPTIC_PHYSIOLOGY = "Synaptic physiology"
+    SYNAPSE_GROUPS = "Synapse groups"
+
+
+class SynapticModelPlacer(Block):
+    """Places synapses with a given synaptic model on a single-cell morphology."""
+
+    synaptic_model: SynapticModelReference = Field(
```
james-isbister: For both these reference parameters there is no None default in the union or and no strings specified to use for the default behaviour

Probably that is better but we should check/ask for the UI to support that if it doesn't already

james-isbister (commented):
Just one comment to check first

obi_one/scientific/tasks/build_synaptome.py
ilkilic: Agreed, both should be required. I have added this to the frontend ticket to confirm support: https://github.com/openbraininstitute/prod-singlecell-simulation/issues/299

james-isbister (approved):
Nice work! Looking forward to trying it in preview!

PR #882, opened 2026-07-03
Add user endpoint for subject registration

jplanasc:
Fixes #794

### DIFF
--- a/app/endpoints/subject.py
+++ b/app/endpoints/subject.py
@@ -1,0 +1,96 @@
+"""Endpoints for user-driven subject registration."""
+
+from http import HTTPStatus
+
+from entitysdk import models
+from fastapi import APIRouter, Depends
+from starlette.requests import Request
+
+from app.dependencies.auth import UserContextDep, user_verified
+from app.dependencies.entitysdk import DatabaseClientDep
+from app.errors import ApiError, ApiErrorCode
+from app.schemas.subject import SubjectRegisterRequest
+
+router = APIRouter(
+    prefix="/declared/subject",
+    tags=["declared"],
+    dependencies=[Depends(user_verified)],
+)
+
+
+@router.get(
+    "",
+    summary="Search for an existing subject by name.",
+    description="Returns the subject if it exists, or 404 if not found.",
+)
+def get_subject(
+    name: str,
+    db_client: DatabaseClientDep,
+    user_context: UserContextDep,  # noqa: ARG001
+) -> dict:
+    """Look up a subject by name."""
+    existing = db_client.search_entity(
+        entity_type=models.Subject, query={"name": name}
+    ).one_or_none()
+
+    if not existing:
+        raise ApiError(
+            message=f"Subject with name '{name}' not found",
+            error_code=ApiErrorCode.NOT_FOUND,
+            http_status_code=HTTPStatus.NOT_FOUND,
+        )
+
+    return existing.model_dump(mode="json")
+
+
+@router.post(
+    "",
+    summary="Register a new subject.",
+    description=(
+        "Registers a new subject with the provided metadata. "
+        "Returns 409 if a subject with the same name already exists."
+    ),
+    status_code=HTTPStatus.CREATED,
+)
+def register_subject(
+    json_model: SubjectRegisterRequest,
+    db_client: DatabaseClientDep,
+    user_context: UserContextDep,  # noqa: ARG001
+    request: Request,  # noqa: ARG001
+) -> dict:
+    """Register a new subject in entitycore."""
+    # Check for duplicate name
+    existing = db_client.search_entity(
+        entity_type=models.Subject, query={"name": json_model.name}
+    ).one_or_none()
+
+    if existing:
+        raise ApiError(
+            message=f"Subject with name '{json_model.name}' already exists (id={existing.id})",
+            error_code=ApiErrorCode.INVALID_REQUEST,
+            http_status_code=HTTPStatus.CONFLICT,
+        )
+
+    # Resolve species (and optionally strain) by ID
+    species = db_client.get_entity(entity_type=models.Species, entity_id=json_model.species_id)
+    strain = (
+        db_client.get_entity(entity_type=models.Strain, entity_id=json_model.strain_id)
+        if json_model.strain_id
+        else None
+    )
+
+    subject = models.Subject(
+        name=json_model.name,
+        description=json_model.description,
+        sex=json_model.sex,
+        weight=json_model.weight,
+        age_value=json_model.age_value,
+        age_min=json_model.age_min,
+        age_max=json_model.age_max,
+        age_period=json_model.age_period,
+        species=species,
+        strain=strain,
+    )
+
+    registered = db_client.register_entity(entity=subject)
+    return registered.model_dump(mode="json")
--- a/app/endpoints/subject.py
+++ b/app/endpoints/subject.py
@@ -9,7 +9,9 @@ from starlette.requests import Request
 from app.dependencies.auth import UserContextDep, user_verified
 from app.dependencies.entitysdk import DatabaseClientDep
 from app.errors import ApiError, ApiErrorCode
-from app.schemas.subject import SubjectRegisterRequest
+from app.schemas.subject import SubjectRegisterRequest, normalize_name_for_comparison
+
+_DUPLICATE_SEARCH_PREFIX_LENGTH = 4
 
 router = APIRouter(
     prefix="/declared/subject",
@@ -43,6 +45,67 @@ def get_subject(
     return existing.model_dump(mode="json")
 
 
+def _find_duplicate_subject_name(db_client: DatabaseClientDep, name: str) -> object | None:
+    """Find a duplicate subject name using normalized comparison.
+
+    Normalizes the input name by lowercasing and stripping all non-alphanumeric
+    characters, then searches for candidates via case-insensitive ILIKE and
+    compares their normalized forms. This ensures that e.g. "Average Rat",
+    "average rat", "AverageRat", "Average-rat", "Average_rat" are all
+    considered duplicate names.
+    """
+    normalized_input = normalize_name_for_comparison(name)
+    if not normalized_input:
+        return None
+
+    # Strategy: build ILIKE patterns that are broad enough to catch all variants.
+    # We insert '%' between each alphanumeric character of the normalized name
+    # so that "averagerat" becomes "%a%v%e%r%a%g%e%r%a%t%" — this matches any
+    # string containing those characters in order regardless of separators.
+    # However, this could be too broad for long names and too slow.
+    #
+    # Practical approach: search using each word of the original name as an ilike
+    # fragment with wildcards, PLUS a second search using just the first word.
+    # Then compare normalized forms client-side.
+
+    seen_ids: set = set()
+    candidates = []
+
+    # Pattern 1: fragments from whitespace-split words
+    words = name.lower().split()
+    fragments = ["".join(c for c in w if c.isalnum()) for w in words]
+    fragments = [f for f in fragments if f]
+
+    if fragments:
+        ilike_pattern = "%" + "%".join(fragments) + "%"
+        results = db_client.search_entity(
+            entity_type=models.Subject, query={"name__ilike": ilike_pattern}
+        ).all()
+        for r in results:
+            if r.id not in seen_ids:
+                seen_ids.add(r.id)
+                candidates.append(r)
+
+    # Pattern 2: if the name has no spaces (e.g. "AverageRat"), also search with
+    # a shorter fragment to catch existing entries stored with separators.
+    # Use the first few alphanumeric characters as a broad match.
+    if len(normalized_input) >= _DUPLICATE_SEARCH_PREFIX_LENGTH:
+        short_pattern = f"%{normalized_input[:_DUPLICATE_SEARCH_PREFIX_LENGTH]}%"
+        results = db_client.search_entity(
+            entity_type=models.Subject, query={"name__ilike": short_pattern}
+        ).all()
+        for r in results:
+            if r.id not in seen_ids:
+                seen_ids.add(r.id)
+                candidates.append(r)
+
+    for candidate in candidates:
+        if normalize_name_for_comparison(candidate.name) == normalized_input:
+            return candidate
+
+    return None
+
+
 @router.post(
     "",
     summary="Register a new subject.",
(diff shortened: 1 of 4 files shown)


app/endpoints/subject.py
```diff
@@ -0,0 +1,96 @@
...
+def register_subject(
+    json_model: SubjectRegisterRequest,
+    db_client: DatabaseClientDep,
+    user_context: UserContextDep,  # noqa: ARG001
+    request: Request,  # noqa: ARG001
+) -> dict:
+    """Register a new subject in entitycore."""
+    # Check for duplicate name
+    existing = db_client.search_entity(
+        entity_type=models.Subject, query={"name": json_model.name}
```
chr-pok: Just wondering if we should make this check for existing subjects more broad, like searching for different spellings/capitalized versions? Otherwise we may end up with different versions like "Average rat", "Average Rat", "AverageRat" etc. which is difficult to maintain and confusing to use.

We could perhaps enforce some naming conventions even, since these subjects will show up as global entities to everyone. For example, we should probably prevent users to create "Test" subjects (i.e., subject names or descriptions containing "test" therein).

jplanasc: Good point.

If you give me the list of what's allowed / forbidden, I can implement it.

chr-pok: Ok, thanks, I will think about it. And discuss with the scientific team on Thursday what would be most useful.

jplanasc: In case it helps, some AI suggestions:

Rules:
**1. Case-insensitive duplicate detection**
When checking if a subject already exists, compare names case-insensitively. "Average Rat" and "average rat" should be considered the same subject.
Implementation: normalize to lowercase before searching, e.g. query={"name__iexact": name} or search with .lower() and compare.

**2. Canonical casing: Title Case**
Enforce that stored names use Title Case (first letter of each word capitalized). This gives consistent display across the platform without burdening users with remembering the exact casing used previously.
Input: "average rat" → stored as "Average Rat"
Input: "AVERAGE RAT" → stored as "Average Rat"

**3. Whitespace normalization**
Strip leading/trailing whitespace
Collapse multiple internal spaces into one
Reject names that are empty after stripping

**4. Forbidden content (blocklist)**
Reject names or descriptions containing any of these patterns (case-insensitive):
"test", "tmp", "todo", "placeholder", "example", "foo", "bar", "asdf", "xxx", "delete me"
This prevents obviously junk/throwaway entries from polluting the global namespace.

**5. Minimum meaningful length**
Name must be at least 3 characters (after stripping)
Description must be at least 10 characters

**6. Character restrictions**
Allow: letters (Unicode), digits, spaces, hyphens, parentheses, commas, periods and slashes
Reject: special characters like @, #, $, !, _, etc.

chr-pok: Thanks, this helps a lot actually! I think we should enforce everything except for 2., since we don't even use title case for our existing subjects, and it won't work for names containing acronyms, e.g., "IARPA MICrONS mouse" (which is valid).

jplanasc: Oh, I forgot to copy the last rule:

**7. No purely numeric names**
Reject names that are just numbers (e.g. "12345").

Shall I enforce this one, too?

chr-pok: Yes, please, this makes sense too.

eleftherioszisis: > 1. Case-insensitive duplicate detection
When checking if a subject already exists, compare names case-insensitively. "Average Rat" and "average rat" should be considered the same subject.
Implementation: normalize to lowercase before searching, e.g. query={"name__iexact": name} or search with .lower() and compare.

Keep in mind that Subject has many columns, and some of them are age, sex, etc. Therefore, I am not sure if it's a duplicate having the same name but a different characteristic. See https://github.com/openbraininstitute/entitycore/blob/8f339d444d6e00f32708531dd2aeae296baab410/app/db/model.py#L684

chr-pok: True, but in practice I think it still does not make sense to allow different subjects with the same name (or just slightly different spelling). This would lead to great confusion, e.g., if we had "Average rat" (male) and "Average Rat" (female) subjects.

eleftherioszisis: IMO uniqueness constraints should be added in the database, which means to decide on the combination of columns that should be considered unique. I agree in normalizing the name, however the name shouldn't replace the queries using the dedicated columns by encoding all information in it.

chr-pok: Hmm, I think uniqueness can only be determined by a combination of all columns, excluding description but including name. There could always be two different subjects with same age, sex, species but coming from different data sets and representing different individuals.

jplanasc: I'll implement whatever you guys tell me, but I think here what we only want is to restrict the subject name based on what's already registered in the platform + some rules. Strictly speaking, the definition of uniqueness will be determined by the subject ID in entitycore (once registered)?

chr-pok (commented):
Thanks @jplanasc, the rules look good!

Just wondering, is it really necessary to do two ilike searchs for patterns 1 & 2 cases? Wouldn't it be easier/faster to retrieve all subjects at once (usually there should not be too many subjects in the database) and do all checks client side?

app/endpoints/subject.py
```diff
@@ -0,0 +1,146 @@
...
+    # We insert '%' between each alphanumeric character of the normalized name
+    # so that "averagerat" becomes "%a%v%e%r%a%g%e%r%a%t%" — this matches any
+    # string containing those characters in order regardless of separators.
+    # However, this could be too broad for long names and too slow.
+    #
+    # Practical approach: search using each word of the original name as an ilike
+    # fragment with wildcards, PLUS a second search using just the first word.
+    # Then compare normalized forms client-side.
+
+    # Pattern 1: fragments from whitespace-split words
```
chr-pok: Just wondering, do we really need ilike searchs for two patterns? Wouldn't it be easier/faster to retrieve all subjects at once (usually there should not be too many subjects in the database) and do all checks client side?

app/schemas/subject.py
```diff
@@ -0,0 +1,132 @@
...
+class SubjectRegisterRequest(BaseModel):
+    """Request body for registering a new subject."""
+
+    name: str = Field(description="Unique name for the subject")
+    description: str = Field(description="Description of the subject")
+    species_id: UUID = Field(description="ID of the species entity")
+    strain_id: UUID | None = Field(default=None, description="ID of the strain entity (optional)")
+    sex: Sex = Field(description="Sex of the subject (male or female)")
+    weight: float | None = Field(default=None, description="Weight in grams", gt=0.0)
+    age_value: timedelta | None = Field(
```
chr-pok: I think the direct use of `timedelta` is not so useful here, since it is not clear how the user should enter age values (as a string, as an int?). Would it be possible to preserve the unit in the user-facing part of the endpoint? Like `age_unit: StrEnum` ("year", "month", "day", "hour", ...) together with `age: int`

The underlying timedelta can be easily constructed like: `age_value = timedelta(**{age_unit: age})`

app/schemas/subject.py
```diff
@@ -0,0 +1,132 @@
...
+        default=None, description="Age value interval", gt=timedelta(0)
+    )
+    age_min: timedelta | None = Field(
```
chr-pok: Same for `age_min` and `age_max`. They could rely on the same `age_unit` mentioned before (i.e., same unit has to be used for age/age_min/age_max).

app/endpoints/subject.py
```diff
@@ -0,0 +1,146 @@
...
+    ).one_or_none()
+
+    if not existing:
+        raise ApiError(
+            message=f"Subject with name '{name}' not found",
+            error_code=ApiErrorCode.NOT_FOUND,
+            http_status_code=HTTPStatus.NOT_FOUND,
+        )
+
+    return existing.model_dump(mode="json")
```
chr-pok, line 42: When an existing subject is returned, the `age_value` seems to be returned as total seconds: e.g., `"age_value": 1209600,` (for a P14 rat).

This is not obvious to a user, so is there a better representation to return to the user which would somehow preserve the value and unit (like `{"age": 14, "age_unit": "days"}`). (Similar concern about registration below.)

app/schemas/subject.py
```diff
@@ -0,0 +1,132 @@
...
+    """Request body for registering a new subject."""
+
+    name: str = Field(description="Unique name for the subject")
+    description: str = Field(description="Description of the subject")
+    species_id: UUID = Field(description="ID of the species entity")
+    strain_id: UUID | None = Field(default=None, description="ID of the strain entity (optional)")
+    sex: Sex = Field(description="Sex of the subject (male or female)")
+    weight: float | None = Field(default=None, description="Weight in grams", gt=0.0)
+    age_value: timedelta | None = Field(
+        default=None, description="Age value interval", gt=timedelta(0)
```
chr-pok:
```suggestion
        default=None, description="Age value", gt=timedelta(0)
```

app/endpoints/subject.py
eleftherioszisis: No, retrieving all subjects per request is not scalable. There are more than 11k subjects in staging and fetching them all locally took for me more than 2 minutes. This is an unnecessary burden for both cpu and memory.

Concerning the ilike searches, I just cleaned up what was already there. IMO it would be sufficient to just ilike search with a lower case stripped from whitespace string e.g. "ratsubject" instead of adding wildcards. However, if you believe a `*rat*subject*` has some value to consider, that's fine with me.

app/schemas/subject.py
eleftherioszisis: timedelta is more descriptive because it can incorporate more than one age units in it. For example:
```
datetime.timedelta(days=2, seconds=3)
```
Therefore, the unit cannot be preserved because there isn't a single unit that is supported at a time.

A user should enter the age as a timedelta, same way we do it in entitysdk (https://github.com/openbraininstitute/entitysdk/blob/fbb31dd7973e0c962c4ca74073841b5ee9858ccd/src/entitysdk/models/base.py#L17)

app/schemas/subject.py
eleftherioszisis: May I suggest to remove age_min and age_max range when registering subjects? IMO having Subject representing a group of subjects is a conceptual mistake and I would prefer keeping it contained instead of encouraging registering new ones.

app/endpoints/subject.py
eleftherioszisis: The user shouldn't consume a service's payload directly. This is what SDKs are for. In entitysdk, the entitycore value is deserialized into a timedelta that has human readable entries like days, hours, seconds, etc.

app/endpoints/subject.py
chr-pok: > There are more than 11k subjects in staging and fetching them all locally took for me more than 2 minutes.

Yes I know, but these subject are here by mistake and should be cleaned up imho! There are thousands of "unkown" subjects instead of a single one. I created a cleanup ticket already a while ago: https://github.com/openbraininstitute/prod-contribute-fix-data/issues/82 (not being worked on yet)

I think normally we should not expect the total number of subjects being larger than ~100 at most.

chr-pok: > However, if you believe a `*rat*subject*` has some value to consider, that's fine with me.

Sorry, I am not exactly sure how ilike search exactly works and what the difference would be. So let's leave it as is I would say.

app/schemas/subject.py
chr-pok: Ok, I see. Still not sure how to use it. Can you give an example how exactly I would enter an age of 14 days? As a float in seconds, i.e., `1209600.0`?

Because the pre-filled example body shows as string `"age_value": "P3D"`, so this is confusing.

app/schemas/subject.py
chr-pok: Agreed, let's remove them.

app/endpoints/subject.py
eleftherioszisis: > > There are more than 11k subjects in staging and fetching them all locally took for me more than 2 minutes.
> 
> Yes I know, but these subject are here by mistake and should be cleaned up imho! There are thousands of "unkown" subjects instead of a single one. I created a cleanup ticket already a while ago: [openbraininstitute/prod-contribute-fix-data#82](https://github.com/openbraininstitute/prod-contribute-fix-data/issues/82) (not being worked on yet)
> 
> I think normally we should not expect the total number of subjects being larger than ~100 at most.

Agreed, however it is not a good design to create an algorithm based on a best case scenario. It is always better to delegate queries to the database, which is implemented for fast queries where indexes can further optimize searches.

> orry, I am not exactly sure how ilike search exactly works and what the difference would be. So let's leave it as is I would say.

In my examples, `ratsubject` would search for the word ratsubject without taking case into account, whereas `*rat*subject*` would match anything in between the two words as well e.g. ratatouillesubject

app/schemas/subject.py
eleftherioszisis: This is how it works in entitysdk:
```
from datetime import timedelta
from pydantic import BaseModel, ConfigDict

class Model(BaseModel):
    model_config = ConfigDict(
        ser_json_timedelta="float",
    )

    age_value: timedelta

m = Model(age_value=timedelta(days=14))

print(m.model_dump_json())
# {"age_value":1209600.0}
```

app/endpoints/subject.py
chr-pok: Hmm, ok, I think this is not needed indeed. I cannot think of any use case atm.

app/schemas/subject.py
chr-pok: Ok, I see. But can we change the example body in the obi-one Swagger UI which shows `"age_value": "P3D"` prefilled?

app/schemas/subject.py
eleftherioszisis: Changed it.

app/schemas/subject.py
eleftherioszisis: done

app/endpoints/subject.py
eleftherioszisis: I simplified the search

app/endpoints/subject.py
```diff
@@ -0,0 +1,111 @@
...
+    considered duplicate names.
+    """
+    normalized_input = normalize_name(name)
+    if not normalized_input:
+        return None
+
+    ilike_pattern = "*" + "?".join(split_name(name)) + "*"
+
+    for result in db_client.search_entity(
+        entity_type=models.Subject, query={"name__ilike": ilike_pattern}
```
chr-pok, line 61: Looks like this ilike search does not work. While testing, I was able to register multiple subjects with the same name "Average rat P14" within the "Shared results + tests" project (in addition to the public one that existed already before).

The corresponding `ilike_pattern` is `"*average?rat?p14*"` which returns empty results for some reason.

chr-pok (commented):
Thanks for the changes, looks good now in general! But the ilike search and therefore the duplicate detection does not seem to work (see my comment below).

app/endpoints/subject.py
eleftherioszisis: Fixed. Thanks for catching that.

chr-pok (approved):
Thanks for the fix! I think this is good enough for the time being, since we cannot possibly capture all different name variations. May think of extending the schema in the future.

PR #883, opened 2026-07-03
Flywire2 current stim

mgeplf:


obi_one/scientific/library/simulation/brian2/simulate_brian2.py
```diff
@@ -207,6 +326,61 @@ def _get_spike_replay(
...
+                )
+            )
+            idx = injection.get_selection(population=population, node_sets=node_sets).flatten()
+            mask = np.zeros((neuron_count,), dtype=bool)
+            mask[idx] = True
+            indicators[f"is_stimulated{i}"] = mask
+            objs[f"stim{i}"] = injection.get_currents(self.simulation.dt, self.simulation.run.tstop)
+
+        model += "\n" + "\n".join(lines) + injection_sum + ": amp\n"
+
```
james-isbister, line 390: Seems like Brian2 has a bit of a suboptimal way of representing stimuli in terms of memory. I see this is the way to do it though

My only concern is for stimulus repetitions. If the user specifies 100 repetitions of the same stimulus in obi-one, this would correspond with 100 sonata stimulus blocks, and 100 x timesteps x number neurons memory (If I've understood correctly).

Could one approach be to sum currents over all sonata stimulus blocks targeting the same node_set, to create a per node_set stimuli in Brian2?

james-isbister (commented):
Looks good. Just one suggestion on the stimulus summing

obi_one/scientific/library/simulation/brian2/simulate_brian2.py
mgeplf: > Could one approach be to sum currents over all sonata stimulus blocks targeting the same node_set, to create a per node_set stimuli in Brian2?

Yes, that would be a nice optimization.  I can do it if you want, but on the other hand, I generally do optimizations like this once they have been demonstrated to be a problem.

I guess it comes down to how much we want to do these optimizations in the first place; the indicator array only takes `127400` bytes for each current stimulus, which isn't too bad, and the TimedArray is (sim_duration/ dt * 8) bytes.

However, are users going to do 100 stimuli?

james-isbister: Hm, yes that's not bad indeed. 

The repetitions are something I'd like to demonstrate to people. Maybe 20 repetitions to several (3-10) large node_sets is a realistic use case for a keen user. I'd probably go for addressing this once now we're focussed on it. But indeed its not totally essential at this stage so I'll let you decide.

mgeplf: > But indeed its not totally essential at this stage so I'll let you decide.

Should be implemented

james-isbister (approved):
LGTM. Thanks for implementing. 

Just needs the latest version of entitysdk in the lock file with "make compile-deps"

PR #884, opened 2026-07-06
improved async implementation

dkeller9:
### Summary

Previously, `POST /declared/{entity_id}/register-mesh` registered a GLB mesh and dispatched
LOD (Level of Detail) generation as a `TaskConfig` submitted to the launch-system — a
separate remote job with its own container, dependencies, and a 1-hour timelimit. This PR
removes that indirection: LOD generation now runs directly from the endpoint, in-process,
as a backgrounded step after the primary mesh is registered.

### What changed

- **Endpoint accepts `.obj` or `.glb`.** If `.obj` is uploaded, it's converted to `.glb`
  (via `pylmesh.load_mesh`/`pylmesh.save_mesh`) for the registered primary asset — but LOD
  generation always runs against the *original* `.obj`, never the converted GLB.
- **Re-registration replaces, not duplicates.** Before uploading, the endpoint checks for
  and deletes any existing asset with the same file name (and, separately, any existing
  `lod-mesh-directory` asset) so re-running registration on the same entity doesn't
  accumulate stale duplicates.
- **LOD generation runs in an isolated subprocess**, not a thread. `ultraliser`'s LOD
  generator crashed the whole API worker outright when invoked via `run_in_threadpool`
  (a background thread) — it appears not to be safe to call off the main thread. Running it
  in a `ProcessPoolExecutor` (spawn context) means a crash there can only kill that
  subprocess, never the API process serving other requests.
- **LOD generation is a genuine background task** (`BackgroundTasks`), not awaited in the
  request. The response returns as soon as the primary GLB is registered; LOD
  generation (subprocess call + upload) continues after.
- **Background task is `async`, not a blocking call in a thread.** `_generate_lods_background_task`
  and `_generate_lods`/`try_generate_and_upload_lods` are now `async def`. Rather than
  blocking a thread on `future.result(timeout=...)` for the `ProcessPoolExecutor` future
  (which occupies one of anyio's limited background-task thread-pool slots — 40 by default
  — for the entire subprocess run), the future is wrapped with `asyncio.wrap_future` and
  awaited via `asyncio.wait_for`, so waiting costs an event-loop callback rather than a
  thread. The executor is shut down via `loop.run_in_executor(None, executor.shutdown)`
  rather than as a context manager, since `ProcessPoolExecutor.__exit__` blocks on
  `shutdown(wait=True)`. The subsequent `entitysdk.upload_directory` call runs via
  `asyncio.to_thread` rather than inline. A few remaining blocking `pathlib.Path` calls
  (`mkdir`, `rglob`, `unlink`) inside these async functions are similarly dispatched to a
  thread (`asyncio.to_thread` / `run_in_threadpool`) to satisfy `ASYNC240`.
- Removed the launch-system task entirely: `TaskConfig`/`TaskDefinition` wiring,
  `TaskType.mesh_lod_generation`, the `obi_one.scientific.tasks.mesh_lod_generation`
  package, `TASK_MAP`/`TASK_DEFINITIONS` entries, the associated accounting branch, and the
  launch-system dependency manifest. No `TaskActivity`/status tracking replaces it — see
  Known limitations below.
- Added `ENTITYSDK_CONNECT_TIMEOUT=30.0` alongside the existing `ENTITYSDK_READ_TIMEOUT=300.0`
  override in `app/application.py`'s lifespan. The shared `httpx.Client` was previously
  relying on entitysdk's default `connect_timeout` of 5s — there was no connect override in
  place despite the read timeout already being bumped. This 5s window was the direct cause
  of the `httpx.ConnectTimeout` failures we were consistently hitting on large directory
  uploads (see Known limitations).

### Why

The launch-system dispatch added a layer of indirection (job submission, remote container,
polling) for what's fundamentally a synchronous transformation on data the endpoint already
has locally. Running it in-process removes that overhead for the common case.

### Known limitations / follow-ups

- **No completion/failure signal to the caller.** Since LOD generation is now a background
  task, the HTTP response can't report whether it ultimately succeeded — the caller finds
  out only by later checking the entity's assets. We deliberately dropped
  `TaskActivity`-based status tracking as part of this change; worth reconsidering if
  callers need to know outcome without polling.
- **`ConnectTimeout` on large directory uploads — resolved client-side.** Uploads were
  consistently failing with `httpx.ConnectTimeout` inside `entitysdk`'s
  `_complete_upload_directory` → `make_db_api_request`. Root cause: entitysdk's default
  `connect_timeout` is 5s, and this app never overrode it (only `read_timeout` had been
  bumped). Fixed by setting `ENTITYSDK_CONNECT_TIMEOUT=30.0` (see above).
- **Large mesh directories (1000+ LOD files) can still fail against remote entitycore.**
  This is a separate, still-open issue from the `ConnectTimeout` above: `entitysdk.upload_directory`
  sends the entire file manifest in a single request/response (`_initiate_directory_upload`);
  for meshes producing ~1500 LOD tiles, this consistently times out around ~150s regardless
  of client-side timeout settings, and the resulting directory asset is created but empty
  (no file content was ever PUT). This appears to be either a gateway timeout between here
  and remote entitycore, or entitycore's own per-file-synchronous handling of the initiate
  request not scaling to this file count. There's no batching/append primitive in
  entitysdk's directory-upload API to work around this from the client side, and switching
  to N individual `upload_file` calls isn't an option here since downstream already depends
  on LODs being one grouped directory asset. **This needs to be resolved on the entitycore
  side** — a bug report with reproduction details is being filed separately.
- `mesh.save`/native library calls in `_convert_obj_to_glb` and `_generate_lods_worker`
  aren't wrapped in explicit per-call timeouts beyond the subprocess's own
  `LOD_GENERATION_TIMEOUT_S` (600s) — fine for now, worth revisiting if mesh sizes grow.

### Testing

- Verified end-to-end against staging entitycore with a simple test cube (`.obj`): convert →
  register → LOD generation → upload, all succeeding synchronously before the background-task
  change.
- Verified the crash-on-thread issue reproduces reliably via `run_in_threadpool` and is
  resolved by subprocess isolation.
- Verified re-registration against the same entity replaces existing assets rather than
  erroring with `ASSET_DUPLICATED`.
- Verified the `ConnectTimeout` failures on large (but <1000-file) directory uploads no
  longer occur after setting `ENTITYSDK_CONNECT_TIMEOUT=30.0`.


app/endpoints/mesh_registration.py
```diff
@@ -1,153 +1,149 @@
...
-            callbacks=[],
-        )
+        if glb_path != temp_mesh_path:
+            await run_in_threadpool(glb_path.unlink, missing_ok=True)
 
-        activity_id = (
-            getattr(task_info, "activity_id", None)
-            if not isinstance(task_info, dict)
-            else task_info.get("activity_id")
+        background_tasks.add_task(
```
jdcourcol, line 124: @dkeller9 we don't want this to be launched on the web server.

We want to have it launch through launch-system, where you can get a dedicated container for that.

What I mentioned is to get rid of the accounting/taskconfig things, however, we still want to launch the computation on the launch system.

Does that make sense ?

jdcourcol (approved):

app/endpoints/mesh_registration.py
dkeller9: @jdcourcol would you be able to specify which endpoint currently does things in the desired way, please? I can then adhere to that approach.

jdcourcol: look here: https://github.com/openbraininstitute/obi-one/pull/844/changes#diff-df8ced94f9cee5c7ac41b2ba72d543d9615daed83c057a234093d4e20f028e87R182

This is a circuit registration endpoint, launching an async job with launch system.

and [here](https://github.com/openbraininstitute/obi-one/pull/844/changes#diff-33941992f8ab357d9705dce23e18cd208c63636fac2aec87673a6464c9930bf7R496) you have the actual call to the launch system


There is a bit more complexity in that use case with the callback, but you should ignore that part.

PR #885, opened 2026-07-06
Skip overview image generation from template

chr-pok:
After https://github.com/openbraininstitute/prod-explore-functionality/issues/516, there is no need any more to register the circuit overview image from a placeholder template if not provided by the user. This PR skips the image registration from the template.

Note: The sim designer image is still generated from a template.


james-isbister (approved):
LGTM

PR #888, opened 2026-07-07
Remove pre-registered `TaskConfig` from mesh LOD generation, run directly via launch-system

dkeller9:
### Summary

Refactors the mesh registration → LOD generation flow to no longer require registering a `TaskConfig` entity in entitycore before the launch-system runs the job, mirroring the pattern already used by `circuit_registration.py`'s `_trigger_validation_task`/`_trigger_asset_generation_task`. The launch-system job now receives `entity_id`, `mesh_asset_id`, `mesh_format`, `project_id`, and `virtual_lab_id` as plain CLI arguments, and a dedicated launch script builds the config in-process instead of fetching it back from a registered `TaskConfig`/`task_config.json` asset.

### Changes

**`app/endpoints/mesh_registration.py`**
- Removed `_register_task_config` and the `task_service.submit_task_job`/`TaskConfig` registration path.
- Added `_trigger_mesh_lod_generation_task`, which builds the launch-system job payload inline (code location/ref, resources, CLI `inputs`, callbacks) and posts it directly via `ls_client`, matching `circuit_registration.py`'s trigger functions.
- `compute_cell` is now resolved per-request via `ComputeCellDep` rather than hardcoded, since the launch-system rejects unregistered executor names (`"local"` only exists on local/dev instances).
- Added `_delete_existing_assets`: before uploading a new source mesh, any existing asset labeled `cell_surface_mesh` (glb or obj) on the entity is deleted first, so re-registering a mesh replaces rather than duplicates.
- `MeshRegistrationResponse` no longer includes `activity_id` — no `TaskActivity` is created for this flow, matching the circuit trigger functions it's modeled on.

**`obi_one/scientific/tasks/mesh_lod_generation/task.py`**
- `_upload_lod_directory` now deletes any existing `lod_mesh_block` asset on the entity before calling `upload_directory`, so re-running LOD generation replaces the directory instead of erroring/duplicating.

**`obi_one/launch_scripts/launch_mesh_lod_generation/main.py`** (new)
- Dedicated launch script, structured like `launch_scripts/launch_circuit_asset_generation/main.py`: parses `--entity_id`, `--mesh_asset_id`, `--mesh_format`, `--virtual_lab_id`, `--project_id`, builds an `entitysdk.Client` and `MeshLodGenerationSingleConfig` directly, and calls `MeshLODGenerationTask.execute()`. Best-effort — logs and returns a non-zero exit code on failure.
- Reuses the existing dependencies file at `launch_task_for_single_config_asset/dependencies/mesh_lod_generation.txt` rather than duplicating it under the new launch path.
- Overrides `entitysdk.config.settings` timeouts (`connect_timeout=30`, `read_timeout=900`, `write_timeout=900`, `pool_timeout=30`) before constructing the client. `entitysdk`'s `make_db_api_request` passes an explicit `timeout=` on every request derived from this settings singleton, which overrides any timeout set on the `httpx.Client` passed in via `http_client=`, so that route doesn't work — the settings singleton has to be mutated directly.

### Why

- Registering a `TaskConfig` entity purely to hand the launch-system a config to fetch back was unnecessary indirection for this task; the circuit endpoints already establish the simpler pattern of passing parameters as CLI args and building the config where it's used.
- Re-running mesh registration or LOD generation for the same entity previously left orphaned/duplicate assets (`cell_surface_mesh`, `lod_mesh_block`) instead of replacing them.
- Default `entitysdk` timeouts (30s read / 5s connect) are far too short for directory operations on the LOD output (~1548 files), causing `ReadTimeout`/SSL handshake timeouts on both the launch-system side (`upload_directory`, `delete_asset`) and the API-server side (`register_mesh`'s synchronous entitysdk calls).

### Known limitations / follow-ups

- `_complete_upload_directory` in entitysdk sends all file paths for a directory asset in a single finalization request; for large LOD directories this is slow enough server-side that even a 900s timeout is a band-aid, not a fix. Worth a follow-up issue against `entitysdk`/`entitycore` to batch this or make it async.
- No `TaskActivity` is created for mesh LOD generation jobs anymore, so there's no DB-tracked provenance/status polling for this task type the way `submit_task_job` provided. If that's needed later, it can be added back independently of the `TaskConfig` removal (create the activity with `used=[]`, pass its id as a CLI arg, update it from the launch script).


obi_one/scientific/tasks/mesh_lod_generation/task.py
jplanasc, line 52: Talking from experience with the em synapse task... consider moving all the checks you can before task launch: if there's an error in the input files (e.g.: unsupported format) or config error, we want to fail as fast as possible, preferably before launching the task (which implies requesting resources to AWS and all the task launch process, which may take time and resources).

The only thing I doubt about is the HAS_MESHING check, since you need to make sure that this is installed where the task needs to run.

jplanasc (approved):
LGTM, just a minor comment to consider.

obi_one/scientific/tasks/mesh_lod_generation/task.py
dkeller9: Thanks, I'll make sure the initial checks are robust. The HAS_MESHING flag has worked previously with the neuromorphomesh code, and the tasks complete okay when submitted to the launch system.

PR #889, opened 2026-07-07
Extracellular locations improvements + endpoint

james-isbister:
Reworks the extracellular electrode-location blocks so probe geometry and world-space placement are correct, specifies orientation as bounded rotation angles, adds a flat configurable grid and a fixed Utah array, exposes summary endpoints, and lets a recording array combine multiple probes and register its plot + locations as assets — with UI-friendly type names, an expanded test suite and an example notebook.

## Block hierarchy

- `PatternedExtracellularLocations` — placement via `origin` (x/y/z) and the aiming rotations `rotation_x`/`rotation_z` (degrees), shared by all patterns.
  - `LinearExtracellularLocations` — a 1D line of electrodes along the local `+Y` axis (no roll).
  - **`TwoDPatternedExtracellularLocations`** (new) — planar patterns (layout in the local X-Y plane); adds `rotation_y`, the roll about the array's own axis.
    - `Neuropixels1ExtracellularLocations`
    - `GridExtracellularLocations`
    - **`UTAHArrayExtracellularLocations`** (new)

## Orientation & placement

- Orientation is specified as **rotation angles in degrees**, each bounded to `[0, 360)`, rather than a direction vector. `rotation_x` and `rotation_z` aim the array's local `+Y` axis; planar arrays additionally expose `rotation_y` to roll about that axis (`R = Rz·Rx·Ry`). `get_global_electrode_xyz_locations` builds the rotation matrix, applies it to the local pattern and translates by `origin` — always a rigid transform, with all-zero rotations the identity (array along `+Y`).
- A **1-D line has no roll** about its own axis, so `LinearExtracellularLocations` exposes only `rotation_x`/`rotation_z`; the roll (`rotation_y`) lives on the shared 2D intermediate `TwoDPatternedExtracellularLocations`.
- Because the caller supplies the orientation there's no reference-axis choice and no ±Z-pole discontinuity. A small notebook helper, `rotation_to_align_y`, converts a target direction (e.g. a volume's principal axis) into these angles, so probes can still "follow the volume".

## Geometry fixes

- Correct the Neuropixels 1.0 layout to the imec spec (staggered, 2 sites/row 32 µm apart, 20 µm rows, 16 µm stagger → 4 columns); previously the second half of the sites overlapped the first and the within-row spacing was wrong.
- Fix the linear array lying on the (1,1,1) diagonal (spacing inflated by √3) — electrodes now sit on the array axis at the configured spacing.
- Separate local vs global coordinates: `get_local_electrode_xyz_locations` returns the origin-relative pattern; `get_global_electrode_xyz_locations` applies the rotation and `origin` once as a proper rigid transform.
- `origin` marks the probe's centre-top (was the side edge).
- Fixed a latent bug where numeric constraints (`n_electrodes`, `spacing`, …) crashed on parameter-sweep lists.

## Blocks

- **Grid** (`GridExtracellularLocations`): a flat rectangular grid in the local X-Y plane with independent `x_offset` (columns) / `y_offset` (rows) spacing.
- **Utah array** (`UTAHArrayExtracellularLocations`, new): a **fixed** 10×10 grid at 400 µm spacing — the dimensions/spacing are locked (not constructor parameters); only the placement (`origin` + rotations) is configurable.
- Each concrete type carries a UI `title`/`description` (e.g. "Linear Probe", "Neuropixels 1.0", "Rectangular Grid", "Utah Array").

## API & summaries

Declared endpoints returning a block's world-coordinate `locations` **plus its properties** (`type`, `origin_*`, `rotation_*`, and the pattern-specific parameters):

- `POST /declared/extracellular-locations/block_summary` — a single block → its summary.
- `POST /declared/extracellular-locations/block_dictionary_summary` — a `{name: block}` dictionary → `{name: summary}`.

Backed by library helpers `extracellular_locations_block_summary` / `extracellular_locations_block_dictionary_summary`.

## Recording-array task

- `electrode_locations` on the config is a dict of blocks, so one recording array combines multiple probes (electrode names are prefixed with the block name).
- Renders the array-vs-somas figure via the shared `plot_extracellular_arrays` helper (in `obi_one/scientific/library/extracellular_locations.py`) and registers, on the `SimulatableExtracellularRecordingArray` entity:
  - the plot as the **`electrode_array_image`** asset, and
  - the locations summary JSON as the **`electrode_locations`** asset.

## Tests & docs

- Expanded test suite: block geometry / rotation / roll / rigid-transform / bounds / parameter-sweeps (Linear, Neuropixels, Grid, Utah, 2D base), the library summary helpers + `plot_extracellular_arrays`, and the two endpoints.
- Example notebook: places Linear + Neuropixels along the soma cloud's principal axis (orientation computed as rotation angles) alongside the Grid and Utah arrays, with a rotation demo and a combined figure; updated the `create_recording_array` notebook.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


obi_one/scientific/blocks/extracellular_locations/extracellular_locations.py
```diff
@@ -1,9 +1,27 @@
...
 from obi_one.core.block import Block
 from obi_one.core.schema import SchemaKey, UIElement
+from obi_one.core.units import Units
+
+# If the array direction is (anti-)parallel to the reference axis used to construct the
+# electrode plane, the cross product degenerates; fall back to a different reference axis.
+_DIRECTION_PARALLEL_TOLERANCE = 0.999
+
+
+def _cross(
```
cattabiani: I understand that you want to keep this vanilla for pydantic but numpy is already a code dep and using it here (instead of reinventing the cross product) would make this more readable and less error-prone

obi_one/scientific/blocks/extracellular_locations/extracellular_locations.py
```diff
@@ -72,104 +94,276 @@ class PatternedExtracellularLocations(ExtracellularLocations, ABC):
...
+        direction_z = float(self.direction_z)  # ty:ignore[invalid-argument-type]
+
+        direction_norm = (direction_x**2 + direction_y**2 + direction_z**2) ** 0.5
+        if direction_norm <= 0.0:  # a norm is non-negative, so this catches the zero vector
+            msg = "Direction vector must be non-zero."
+            raise ValueError(msg)
+        forward = (
+            direction_x / direction_norm,
+            direction_y / direction_norm,
+            direction_z / direction_norm,
```
cattabiani: since you compute norm and normalized array at least twice consider creating a small geometry function like cross to do this. Writing geometry formulae over and over is not DRY and error-prone

obi_one/scientific/blocks/extracellular_locations/extracellular_locations.py
```diff
@@ -72,104 +94,276 @@ class PatternedExtracellularLocations(ExtracellularLocations, ABC):
...
-            new_x = self.origin_x + (x * unit_direction_x)
-            new_y = self.origin_y + (y * unit_direction_y)
-            new_z = self.origin_z + (z * unit_direction_z)
-            xyz_locations.append((new_x, new_y, new_z))
-        return xyz_locations
+        # Build an orthonormal basis (right, forward, out) mapping the local (X, +Y, Z) axes into
+        # world space. Pick a reference axis that is not (anti-)parallel to `forward`.
+        reference = (0.0, 0.0, 1.0)
+        if abs(forward[2]) > _DIRECTION_PARALLEL_TOLERANCE:
+            reference = (1.0, 0.0, 0.0)
```
cattabiani: when forward crosses the 0.999 threshold the reference axis snaps from Z to X, which causes the lateral axes (right/out) to jump discontinuously. For 2D arrays (Neuropixels, Grid, Utah) this means the face orientation flips if someone parameter-sweeps direction through the ±Z poles. Probably fine for normal usage but it could be surprising

cattabiani (commented):
request:

>1) the logic of the coordinate transforms in this [file](https://github.com/openbraininstitute/obi-one/blob/376ba72c00fe7ca46e2b02ec6ffa0a18292dabb1/obi_one/scientific/library/extracellular_locations.py), 
> 2) That it runs with the local version of BlueRecording - I haven't had a chance to check as the PR didn't touch too heavily the BlueRecording part

I tested locally with:

```python
import numpy as np

from bluerecording.weights import Electrode, ElectrodeType

import obi_one as obi



probe = obi.LinearExtracellularLocations(n_electrodes=4, spacing=20.0, origin_x=100.0, origin_y=200.0, origin_z=300.0)

electrodes = [

    Electrode(name=f'e_{i}', position=np.array(loc, dtype=float), type=ElectrodeType.POINT_SOURCE)

    for i, loc in enumerate(probe.get_global_electrode_xyz_locations())

]

for e in electrodes:

    print(f'  {e.name}: {e.position}')

print('Integration OK')
```

result:

```
numprocs=1

  e_0: [100. 200. 300.]

  e_1: [100. 220. 300.]

  e_2: [100. 240. 300.]

  e_3: [100. 260. 300.]

Integration OK
```

so the coordinates pass though as expected.

Second request:

> The [notebook](https://github.com/openbraininstitute/obi-one/blob/376ba72c00fe7ca46e2b02ec6ffa0a18292dabb1/examples/K_create_recording_array/create_recording_array.ipynb) for running the task, and the [notebook](https://github.com/openbraininstitute/obi-one/blob/376ba72c00fe7ca46e2b02ec6ffa0a18292dabb1/examples/obi_one/scientific/blocks/extracellular_locations/extracellular_locations.ipynb) for playing with the locations/transforms

[Notebook 2 (transforms)](github.com/openbraininstitute/obi-one/blob/376ba72c00fe7ca46e2b02ec6ffa0a18292dabb1/examples/obi_one/scientific/blocks/extracellular_locations/extracellular_locations.ipynb): All array types render correctly against a circuit. Transforms, axial rotation, and plotting work.

[Notebook 1 (full pipeline)](https://github.com/openbraininstitute/obi-one/blob/376ba72c00fe7ca46e2b02ec6ffa0a18292dabb1/examples/K_create_recording_array/create_recording_array.ipynb): Ran with a 10-neuron circuit. obi-one positions flow into BlueRecording, compute_weights succeeds, weights file has correct electrode positions.

So, for me even this part worked. 

Do you want a full review? I see that everything is still in draft

obi_one/scientific/blocks/extracellular_locations/extracellular_locations.py
james-isbister: Yes, this was a bit awkward indeed. I replaced specification of the orientation by a direction, with specification by 3 angles in degrees

obi_one/scientific/blocks/extracellular_locations/extracellular_locations.py
```diff
@@ -1,16 +1,36 @@
...
+def _rotation_matrix(x_deg: float, y_deg: float, z_deg: float) -> np.ndarray:
+    """Return the rotation matrix ``Rz(z) @ Rx(x) @ Ry(y)`` (degrees).
+
+    ``Ry`` (roll about the local ``+Y`` axis) is applied first, then ``Rx`` and ``Rz`` aim ``+Y``.
+    """
+    x, y, z = np.radians([x_deg, y_deg, z_deg])
+    rx = np.array([[1.0, 0.0, 0.0], [0.0, np.cos(x), -np.sin(x)], [0.0, np.sin(x), np.cos(x)]])
+    ry = np.array([[np.cos(y), 0.0, np.sin(y)], [0.0, 1.0, 0.0], [-np.sin(y), 0.0, np.cos(y)]])
+    rz = np.array([[np.cos(z), -np.sin(z), 0.0], [np.sin(z), np.cos(z), 0.0], [0.0, 0.0, 1.0]])
+    return rz @ rx @ ry
```
cattabiani: scipy has exactly this (so we do not need to hardcode formulae). However, scipy is not a core dependency. I let you decide if you think it is worth adding scipy as code dependency

nitpick:

you are computing multiple times sines and cosines. You can consider precomputing them using the values later (so you compute the values only once).

Anyway, not a big deal. We are talking about minuscule speed-up (maybe) in a non-hot path. As you prefer

obi_one/scientific/blocks/extracellular_locations/extracellular_locations.py
```diff
@@ -47,129 +71,273 @@ class PatternedExtracellularLocations(ExtracellularLocations, ABC):
         },
     )
 
-    direction_x: float | list[float] = Field(
+    rotation_x: (
```
cattabiani, line 72: Now we have gimbal locks instead of orientation snapping:

https://en.wikipedia.org/wiki/Euler_angles

I do not think it is a problem. You can use Euler angles or direction, np. We should just know that there are trade offs. 

The way to avoid everything is with quaternions but this requires to reinvent hot water or import scipy

cattabiani (commented):
The rotation via euler angles looks fine. 

**Nitpick:** a few more tests could help. Atm there is:

- No direct test that rotation_y=90 on a 2D array moves the X extent into Z (the old test_rotation_90_moves_width_into_z)
- No rotation_y = 360 ≈ 0 equivalence test

obi_one/scientific/blocks/extracellular_locations/extracellular_locations.py
james-isbister: There is a Rotation in scipy which I'm now using. scipy was already installed implicitly through quite a few other packages, so I've made it an explicitly dependency now

cattabiani (commented):
LGTM for the rotation stuff. I did not check the rest since you asked only for this part

cattabiani (approved):
LGTM. I did not check the frontend but they tell me it is fine

PR #891, opened 2026-07-08
Select standard morphology H5 asset for morphometrics

ilkilic:
This PR updates morphometrics asset selection so the preview/register metric endpoints use the standard morphology asset instead of accidentally selecting a morphology-with-spines H5. Previously, compute_measurement_kinds() selected the first asset matching the requested content type. For h5, that could pick an asset labeled morphology_with_spines depending on asset ordering.


dkeller9 (approved):
LGTM

PR #892, opened 2026-07-08
Default edge popul fix

chr-pok:
Small fix for resolving the default edge population based on the population name. Required for circuits like nbS1 which has local and midrange connectivity.


james-isbister (approved):
LGTM!

PR #893, opened 2026-07-10
Circuit extraction resource allocation update

chr-pok:
Updates of resource allocation for circuit extraction:
- Estimation of disk space based in the number of output synapses
- Estimation of number of output synapses as fraction of extracted neurons
- Reusing number of extracted neurons from accounting parameters, if available
- Otherwise, fall back to maximum count, as before
- Updated count estimate in accounting parameters based on new neuron sets
- Requesting extra ephemeral storage space, if needed (21-200GB)
- Raising errors in case
  - 0 neurons extracted
  - estimated storage exceeds 200GB
- Added & updated tests

Related issues:
- https://github.com/openbraininstitute/prod-build-circuit/issues/28
- https://github.com/openbraininstitute/prod-build-circuit/issues/34

### DIFF
--- a/app/services/resource_estimation/circuit_extraction.py
+++ b/app/services/resource_estimation/circuit_extraction.py
@@ -131,7 +131,7 @@ def estimate_task_resources(  # noqa: PLR0914
     )
     input_size_synapses = (sbio + svirt) if single_config.initialize.do_virtual else sbio  # ty:ignore[unresolved-attribute]
     output_fraction = (accounting_parameters.count / nbio) if accounting_parameters else 1.0
-    output_size_synapses = input_size_synapses * output_fraction # Rough scaling based on fraction of neurons, if available
+    output_size_synapses = input_size_synapses * output_fraction
     output_size_gb = 1 + output_size_synapses * 1.85e-7
     storage_gb = _get_required_extra_storage_space(output_size_gb)
 
--- a/app/services/resource_estimation/circuit_extraction.py
+++ b/app/services/resource_estimation/circuit_extraction.py
@@ -6,6 +6,7 @@ from entitysdk import models
 
 from app.schemas.accounting import AccountingParameters
 from app.schemas.task import Resources, TaskDefinition, TaskLaunchSubmit
+from app.types import MachineExecutorImageType
 from obi_one import deserialize_obi_object_from_json_data
 from obi_one.core.registry import task_registry
 from obi_one.scientific.library.circuit_metrics import (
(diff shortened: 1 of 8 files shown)


jdcourcol (approved):
cool, thank you

app/services/resource_estimation/circuit_extraction.py
```diff
@@ -132,5 +143,7 @@ def estimate_task_resources(  # noqa: PLR0914
             "memory": mem_gb,
             "timelimit": f"{time_h:02d}:00",
             "compute_cell": compute_cell,
+            "image_type": MachineExecutorImageType.python_3_12_compiler,
```
eleftherioszisis, line 146: Why override the image_type here? It shouldn't change based on this estimation.

app/services/resource_estimation/circuit_extraction.py
```diff
@@ -121,9 +131,10 @@ def estimate_task_resources(  # noqa: PLR0914
         ]
     )
     input_size_synapses = (sbio + svirt) if single_config.initialize.do_virtual else sbio  # ty:ignore[unresolved-attribute]
-    output_size_synapses = input_size_synapses  # Using maximum output count
+    output_fraction = (accounting_parameters.count / nbio) if accounting_parameters else 1.0
```
eleftherioszisis, line 134: I am not sure why accounting is involved in the calculation of how much storage the task requires.

app/services/resource_estimation/circuit_extraction.py
chr-pok: But it is a requirement for using "ephemeral_storage" if I understood the docs correctly. So it has to be python_3_12_compiler I thought.

eleftherioszisis: This is defined in the task declaration in app/mappings.py. By default it uses the python_3_12_compiler. If you want to be explicit about it you can set it in there, but not really needed because it is the default.

However, given that for a task the image type does not change (unless it is intentional for some reason), it shouldn't be overridden dynamically.

app/services/resource_estimation/circuit_extraction.py
chr-pok: This was already planned in the initial accounting PR. The idea is to reuse the extracted target neuron count from `accounting_parameters.count` instead of re-running the same code for resolving the target neuron set twice (since it can be costly).

eleftherioszisis: I see. IMO it would be better to pass the number of neurons and not the accounting parameters themselves.

chr-pok: Yes, I thought about it, but then it would no longer be general for all tasks. I mean, we would need to distinguish already in `task_launch_endpoint` if the task is circuit extraction (in which case the count is target neuron count) or some other task. By passing the accounting parameters, the estimation function of each task can use that information if needed, or just ignore it.

app/services/resource_estimation/circuit_extraction.py
chr-pok: Ok I see, I will remove that override in a follow-up PR (since I merged that already).

PR #894, opened 2026-07-10
Add `.glb` support to mesh validation endpoint

dkeller9:
### Summary
Extends the `/declared/test-mesh-file` endpoint to accept and validate `.glb` files in addition to `.obj`, with matching test coverage for both formats.

### Changes

**`app/endpoints/mesh_validation.py`**
- Replaced the hardcoded `.obj`-only extension check with a `VALID_MESH_EXTENSIONS = {".obj", ".glb"}` set; the 400 error now lists all accepted extensions.
- `_save_upload_to_tempfile` is now called with the uploaded file's actual extension (`suffix=file_extension`) instead of a hardcoded `.obj`, so temp files preserve the correct suffix for either format.
- Generalized the load-failure message from `"Failed to load OBJ file with PyVista"` to `"Failed to load MESH file with PyVista"` since the reader is no longer OBJ-specific.
- Updated docstrings, the `File(description=...)` annotation, and the endpoint `description` to reflect `.obj, .glb` support.

**`app/routers/declared/test_mesh_validation.py`**
- Parametrized all extension-relevant tests (success, empty file, too-large, reader-fails, os-error, background-cleanup, too-large-via-size-header) over `.obj` and `.glb`, driven by the endpoint's exported `VALID_MESH_EXTENSIONS`.
- `test_validate_mesh_file_invalid_extension` now asserts the error message references both supported extensions.
- Updated the two `pytest.raises(..., match=...)` assertions to match the generalized `"Failed to load MESH file"` message.
- Added a `.glb`-specific case to `validate_mesh_reader` success/failure unit tests.

### Testing
- `python -m py_compile` on both files passes.
- Existing test suite structure preserved; new parametrization multiplies extension-dependent cases by 2 (one run per supported extension) without changing assertions' intent.

### DIFF
--- a/app/endpoints/mesh_validation.py
+++ b/app/endpoints/mesh_validation.py
@@ -19,6 +19,9 @@ router = APIRouter(prefix="/declared", tags=["declared"], dependencies=[Depends(
 # Max file size: 5 GB
 MAX_FILE_SIZE = 5 * 1024 * 1024 * 1024
 
+# Supported MESH file extensions
+VALID_MESH_EXTENSIONS = {".obj", ".glb"}
+
 
 class ValidationStatus(StrEnum):
     """Enumeration of possible validation outcomes."""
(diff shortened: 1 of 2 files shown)


jplanasc (approved):
LGTM.

PR #896, opened 2026-07-10
Circuit extraction new neuron sets update

chr-pok:
Required updates to make the circuit extraction form compatible with the new neuron sets (incl. combined neuron sets):
- Moved target neuron set to initialize and turned into reference
- Added neuron set dict
- Added logic for handling default neuron sets

(All following the same logic as for circuit simulations.)

Related issues:
- https://github.com/openbraininstitute/prod-circuit-simulation/issues/145
- https://github.com/openbraininstitute/prod-build-circuit/issues/17


james-isbister (approved):
LGTM!

PR #897, opened 2026-07-13
Add CircuitSimplification tasks

darshanmandge:


obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,468 @@
...
+        )
+        mod_dir: str | None = Field(
+            default=None,
+            title="MOD Directory",
+            description="Custom MOD files directory (optional).",
+            json_schema_extra={
+                SchemaKey.UI_ELEMENT: UIElement.STRING_INPUT,
+            },
+        )
+        export_format: ExportFormat = Field(
```
james-isbister: Rather than creating a new ExportFormat enum, we can use the existing libsonata types:

```
from libsonata import SimulatorType
SimulatorType.NEURON
SimulatorType.Brian2
SimulatorType.LearningEngine
"NEST" # Does not exist yet in the sonata standard
```
Also see the Literal notation for string_selection in the docs, and other places which its used

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,468 @@
...
+
+
+class BlockGroup(StrEnum):
+    """Block groups for the simplification form."""
+
+    SETUP = "Setup"
+    SIMPLIFICATION = "Simplification"
+
+
+class SimplificationAlgorithm(StrEnum):
```
james-isbister: Will all these simplification algorithms be available for all simulator types? And for all future simulator types we might end up supporting. Possibly not? Probably we need a different structure for such specificity. Can discuss one if you agree

Also SimplificationAlgorithm seems like the wrong term. It's really a target model

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,468 @@
...
+        )
+        biophysical_population: str | None = Field(
+            default=None,
+            title="Biophysical Population",
+            description="Name of the biophysical population to simplify (default: auto-detect).",
+            json_schema_extra={
+                SchemaKey.UI_ELEMENT: UIElement.STRING_INPUT,
+            },
+        )
+        simplified_params_path: str | None = Field(
```
james-isbister: Do you think simplified_params_path, morphology_dir, hoc_dir, mod_dir really need to be exposed to the user? Maybe we can just choose them for the user to reduce complexity?

Or put them into an Advanced block like we do for parameters in the EMSynapseMapping task, if they're really useful

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,468 @@
...
+            description="Select one or more simplification algorithms. Each selected algorithm"
+            " produces a separate simplified output circuit.",
+            json_schema_extra={
+                SchemaKey.UI_ELEMENT: UIElement.STRING_SELECTION,
+            },
+        )
+        compute_filters: bool = Field(
+            default=True,
+            title="Compute Filters",
+            description="Whether to compute tau_corr and w_corr filters for soma correction.",
```
james-isbister: Not clear what these would be for a new user. Could we have a better title and description?

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,468 @@
...
+            json_schema_extra={
+                SchemaKey.UI_ELEMENT: UIElement.BOOLEAN_INPUT,
+            },
+        )
+        biophysical_population: str | None = Field(
```
james-isbister: Are we just going to simplify a single biophysical population?

james-isbister (commented):
Initial comments/questions on the config

pyproject.toml
```diff
@@ -164,6 +165,7 @@ explicit = true
 repo_name = { git = "https://github.com/openbraininstitute/ion-channel-builder.git", branch="main" }
 neuromorphomesh = { index = "obi-codeartifact" }
 ultraliser = { index = "obi-codeartifact" }
+sonata-simplify = { path = "../sonata_simplify", editable = true }
```
danifr: this line will need to be changed to `sonata-simplify = { index = "obi-codeartifact" }` once this is merged https://github.com/openbraininstitute/sonata-simplify/pull/1

darshanmandge: Thanks for your help with the package, @danifr! I have changed the above line.

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,606 @@
...
+        SchemaKey.GROUP_ORDER: [
+            BlockGroup.SETUP,
+            BlockGroup.SIMPLIFICATION,
+        ],
+        SchemaKey.PROPERTY_ENDPOINTS: {
+            "circuit": "/mapped-circuit-properties/{circuit_id}",
+        },
+    }
+
+    _campaign_task_config_type: ClassVar[TaskConfigType] = (
```
james-isbister, line 141: single_coord_class_name, _campaign_task_config_type, _campaign_generation_task_activity_type, _single_task_config_type, _single_task_activity_type are no longer needed in this file now this information is in config_task_map.py

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,606 @@
...
+    class Initialize(Block):
+        circuit: CircuitDiscriminator | list[CircuitDiscriminator] = Field(
+            title="Circuit",
+            description="Input SONATA circuit to be simplified.",
+            json_schema_extra={
+                SchemaKey.UI_ELEMENT: UIElement.MODEL_IDENTIFIER,
+                SchemaKey.PARAMETER_ORDER_PRIORITY: 100,
+            },
+        )
+        node_set: BiophysicalNeuronSetReference | None = Field(
```
james-isbister, line 167: This can be called target_neuron_set, it's called node_set in simulation generation for legacy reasons

james-isbister: There is no neuron_set block_dictionary so no where to get a neuron set reference

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,606 @@
...
+            json_schema_extra={
+                SchemaKey.UI_ELEMENT: UIElement.REFERENCE,
+                SchemaKey.REFERENCE_TYPES: ATOMIC_BIOPHYSICAL_NEURON_SETS_REFERENCE_TYPES,
+                SchemaKey.PARAMETER_ORDER_PRIORITY: 99,
+                SchemaKey.UI_HIDDEN: True,
+            },
+        )
+
+    class Simplification(Block):
+        algorithms: list[SimplificationModelType] = Field(
```
james-isbister, line 183: I think we said that we would have a block_dictionary of simplifications which in future could be applied to different neuron sets? Or was it something else? Could you remind me of the outcome from the discussion with @MWolfR ?

james-isbister (commented):
Initial comments and questions about the config

app/mappings.py
```diff
@@ -276,6 +276,24 @@
...
     ),
+    TaskType.circuit_simplification: TaskDefinition(
+        task_type=TaskType.circuit_simplification,
+        config_type=TaskConfigType.circuit_simplification__config,
+        activity_type=TaskActivityType.circuit_simplification__execution,
+        code=PythonRepositoryCode(
+            location=settings.OBI_ONE_REPO,
+            ref=APP_TAG,
+            path=OBI_ONE_CODE_PATH,
+            dependencies=str(OBI_ONE_DEPS_DIR / "circuit_simplification.txt"),
```
mgeplf, line 287: need to add `capabilities=Capabilities(private_packages=True),`

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,606 @@
...
+    3. Registers each output (SONATA + NEST export) as a Circuit entity
+       with derivation links to the parent, using the correct target_simulator.
+    """
+
+    config: CircuitSimplificationSingleConfig
+    _circuit: Circuit | None = PrivateAttr(default=None)
+    _circuit_entity: models.Circuit | None = PrivateAttr(default=None)
+    _temp_dir: tempfile.TemporaryDirectory | None = PrivateAttr(default=None)
+
+    def __del__(self) -> None:
```
mgeplf, line 257: this seems very complicated when a `with tempfile.TemporaryDirectory` could be called when the temp directory is needed in `execute`

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,606 @@
...
+
+        Args:
+            circuit_config_path: Path to the simplified circuit_config.json.
+            algorithm_name: Algorithm name (for error messaging).
+
+        Raises:
+            RuntimeError: If the circuit cannot be loaded.
+        """
+        try:
+            import bluepysnap  # ruff: ignore[import-outside-top-level]
```
mgeplf, line 420: bluepysnap should always be installed, so this isn't needed

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,606 @@
...
+            "manifest": {"$BASE_DIR": str(Path(input_circuit_path).parent)},
+            "network": "$BASE_DIR/" + str(Path(input_circuit_path).name),
+            "output": {
+                "output_dir": str(output_dir / "output"),
+                "spikes_file": "spikes.h5",
+            },
+            "run": {
+                "dt": 0.025,
+                "random_seed": 1,
+                "tstop": 3000.0,
```
mgeplf, line 344: where do these numbers come from? is the tstop used by simplification?

obi_one/scientific/tasks/circuit_simplification/task.py
```diff
@@ -0,0 +1,606 @@
...
+            # Create output directory for this algorithm
+            output_dir = self.config.coordinate_output_root / algorithm_name
+            output_dir.mkdir(parents=True, exist_ok=True)
+
+            # Build simulation config for the pipeline.
+            # If a node set was resolved, include node_set (name) and
+            # node_sets_file (relative path to node_sets.json in the output root).
+            # The sim config is in coordinate_output_root/algorithm_name/, so
+            # node_sets.json in coordinate_output_root/ is at "../node_sets.json".
+            node_sets_file = "../node_sets.json" if node_set_name is not None else None
```
mgeplf, line 525: this seems strange to me, shouldn't the file path be encoded in the simulation_config? if so, one should be able to get it properly without having to hardcode a relative path

PR #899, opened 2026-07-13
Move block modules into their own subpackages

james-isbister:
* Gives every module under `scientific/blocks/` its own subpackage, so the directory is uniformly package-based — matching `neuron_sets/`, `stimuli/` and `extracellular_locations/`.
* Splits `recording.py` into `recordings/{base,soma,ion_channel}.py`; moves `ion_channel_model.py`, `ion_channel_equations.py` and `afferent_synapses.py` into same-named subpackages (git tracks those three as pure renames).
* Pure code move, no behaviour change: the `type` discriminator comes from the class name, not the module path, so serialized configs still deserialize unchanged.
* Importers repointed — `obi_one/__init__.py`, the `unions/` modules, `tasks/ion_channel_modeling.py` (which imports `ion_channel_equations` as a module namespace, so it needed the inner module), one test, three notebooks.
* Full suite 1317 passed; the 4 mesh-LOD failures are pre-existing (`No module named 'ultraliser'`). Schema tests and ruff clean.


AurelienJaquier (approved):
looks good !

PR #900, opened 2026-07-13
Emodel Building Workflows: Efeature Extraction + Optimisation + Export & Validation 

darshanmandge:
- Creating a form for the Intracellular EFeature Extraction workflow of the EModel Building Pipeline


obi_one/scientific/from_id/task_result_from_id.py
```diff
@@ -0,0 +1,76 @@
...
+        asset_label: AssetLabel,
+        dest_dir: Path = Path(),
+        db_client: entitysdk.client.Client = None,  # ty:ignore[invalid-parameter-default]
+    ) -> Path:
+        """Download a single asset matching the given label to dest_dir.
+
+        Returns the path to the downloaded file.
+        """
+        output_dir = create_dir(dest_dir)
+        asset = db_client.download_assets(
```
eleftherioszisis: Use fetch_assets. It is better not to download if the local data store has the data available.

obi_one/scientific/from_id/task_result_from_id.py
```diff
@@ -0,0 +1,76 @@
...
+from obi_one.core.entity_from_id import EntityFromID
+
+
+class TaskResultFromID(EntityFromID):
+    """Wrapper for downloading assets from a TaskResult entity."""
+
+    entitysdk_class: ClassVar[type[Entity]] = TaskResult
+    _entity: TaskResult | None = PrivateAttr(default=None)
+
+    def download_asset_by_label(
```
eleftherioszisis, line 20: This is a utility function. If not already available add it in obi_one/utils/db_sdk.py

obi_one/scientific/from_id/task_result_from_id.py
```diff
@@ -0,0 +1,76 @@
...
+        output_dir = create_dir(dest_dir)
+        asset = db_client.download_assets(
+            self.entity(db_client=db_client),
+            selection={"label": asset_label},
+            output_path=output_dir,
+        ).one()
+
+        return asset.path
+
+    def download_json_asset_by_label(
```
eleftherioszisis: Sane. This already exists in obi_one/utils/db_sdk.py

obi_one/scientific/from_id/task_result_from_id.py
```diff
@@ -0,0 +1,76 @@
...
+            asset_or_id=db_client.select_assets(
+                entity=self.entity(db_client=db_client),
+                selection={"label": asset_label, "content_type": ContentType.application_json},
+            ).one(),
+        )
+        import json  # noqa: PLC0415
+
+        return json.loads(bytes_content)
+
+    def download_directory_asset_by_label(
```
eleftherioszisis, line 37: Same. If not already available add it in obi_one/utils/db_sdk.py.

obi_one/scientific/library/electrical_cell_recording_properties.py
```diff
@@ -0,0 +1,213 @@
...
+def get_recording_protocols(
+    recording_ids: list[str],
+    db_client: entitysdk.client.Client,
+) -> dict[str, list[str]]:
+    """Return ``{recording_id: [protocol_name, ...]}`` for each recording.
+
+    Reads protocol names from the ``stimuli`` field of each
+    ``ElectricalCellRecording`` entity — no NWB download required.
+    """
+    from entitysdk.models import ElectricalCellRecording  # noqa: PLC0415
```
eleftherioszisis: Why internal?

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/blocks.py
```diff
@@ -0,0 +1,251 @@
...
+        default="SSCX",
+        title="Brain region",
+        description="Brain region tag passed to ``EModel_pipeline``.",
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.STRING_INPUT},
+    )
+    etype: str = Field(
+        title="E-type",
+        description="Electrical type tag (selected by the user in the UI).",
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.STRING_INPUT},
+    )
```
eleftherioszisis: Is it common instead of existing database entities to let the user pass any string?

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/blocks.py
```diff
@@ -0,0 +1,251 @@
...
+        description=(
+            "Ion channel model entities whose ``.mod`` assets are staged into"
+            " ``./mechanisms/``. The params file is built dynamically from these"
+            " models."
+        ),
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.MODEL_IDENTIFIER_MULTIPLE},
+    )
+
+
+class ParamsFileSelection(Block):
```
eleftherioszisis, line 97: I am not sure I understand this from the point of view of the UI. Why the user has to provide a pre-built json dict by bluepyemodel. Where does that dict live? How users generate that? Is it a database entity?

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/blocks.py
```diff
@@ -0,0 +1,251 @@
...
+        description=(
+            "Embedded BluePyEModel params JSON. Must contain top-level"
+            " keys: 'mechanisms', 'distributions', 'parameters'."
+            " Leave empty to use the dynamic builder."
+        ),
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.BLOCK_DICTIONARY},
+    )
+
+
+def validate_params_file(params: dict) -> None:  # noqa: C901
```
eleftherioszisis: Schema validation isn't what pydantic Blocks are for? Why the user is allowed to pass an arbitrary dictionary only to be validated later on. Who has the knowledge of what to add in that dictionary? Shouldn't it be a pydantic schema instead?

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/blocks.py
```diff
@@ -0,0 +1,251 @@
...
+                    f" which is not defined in 'distributions'"
+                    f" (available: {sorted(dist_names)})."
+                )
+                raise OBIONEError(msg)
+
+
+class OptimizationParams(Block):
+    """``optimisation_params`` block (passed verbatim to BluePyEModel)."""
+
+    offspring_size: PositiveInt | list[PositiveInt] = Field(
```
eleftherioszisis, line 187: We should have an allowed range here.

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/blocks.py
```diff
@@ -0,0 +1,251 @@
...
+    )
+    seed: NonNegativeInt | list[NonNegativeInt] = Field(
+        default=1,
+        title="Random seed",
+        description="Seed forwarded to ``pipeline.optimise(seed=...)``.",
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.INT_PARAMETER_SWEEP},
+    )
+
+    # Analysis settings (used by pipeline.plot() after optimisation)
+    plot_currentscape: bool = Field(
```
eleftherioszisis: If we plan to view the results in a homogeneous way, would it be better perhaps to have a standard behavior instead of providing options of what to plot?

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/blocks.py
```diff
@@ -0,0 +1,251 @@
...
+        default=True,
+        title="Plot currentscape",
+        description="Generate currentscape plots during analysis.",
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.BOOLEAN_INPUT},
+    )
+    currentscape_title: str = Field(
```
eleftherioszisis: Same here. Is it really necessary for the user to change the title?

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/config.py
```diff
@@ -0,0 +1,248 @@
...
+            self.initialize.extraction_task_result.entity(db_client=db_client),
+            self.morphology_selection.morphology.entity(db_client=db_client),
+        ]
+        entities.extend(
+            icm.entity(db_client=db_client) for icm in self.parameters_selection.ion_channel_models
+        )
+        return entities
+
+    @property
+    def use_params_file(self) -> bool:
```
eleftherioszisis, line 90: If parameters are passed to the task, they shouldn't be optional.

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        # --- 3. Download ion channel models (.mod files) ---
+        self._stage_mechanisms(coord_root, db_client)
+
+        # --- 4. Fetch traces via derivation chain ---
+        self._stage_traces(extraction_tr, coord_root, db_client)
+
+        # --- 5. Merge recipe ---
+        # Rename extraction key ("emodel") to the user-specified emodel name
+        if "emodel" in base_recipes and emodel != "emodel":
+            base_recipes[emodel] = base_recipes.pop("emodel")
```
eleftherioszisis: Why is there an "emodel" that has different name in base_recipes in the first place? That shouldn't be allowed. The task parameterization should construct the recipe. There should be no other custom values that are not provided by the user.

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        recipes = base_recipes
+        recipes.setdefault(emodel, {})
```
eleftherioszisis: Minor: Set `recipes[emodel] = {...}`  instead of initializing empty and updating.

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        recipes[emodel]["morph_path"] = "./morphologies/"
+        recipes[emodel]["morphology"] = [[mtype, morph_filename]]
+        recipes[emodel]["features"] = f"config/features/{emodel}.json"
```
eleftherioszisis: It is error prone to reconstruct the paths that have been already generated/defined before. set this to a variable either passing it where it's needed or making the function return it and use it here.

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        params_filename = self._stage_params(coord_root, db_client)
```
eleftherioszisis: The params staging should be done before the creation of the recipe and not during its creation.

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        recipes[emodel]["params"] = f"config/params/{params_filename}"
```
eleftherioszisis: Same here. make your _stage_params return a full Path of the generated file. If you need the filename use path.name

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        coord_root: Path,
+        db_client: entitysdk.client.Client,
+    ) -> None:
+        """Download targets config JSON from extraction TaskResult."""
+        from entitysdk.types import AssetLabel  # noqa: PLC0415
+
+        targets_dir = coord_root / "config" / "extract_config"
+        targets_dir.mkdir(parents=True, exist_ok=True)
+        try:
+            extraction_tr.download_asset_by_label(
```
eleftherioszisis: Does this asset exist?

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        self,
+        extraction_tr: TaskResultFromID,
+        coord_root: Path,  # noqa: ARG002
+        db_client: entitysdk.client.Client,
+    ) -> dict:
+        """Download recipes JSON from extraction TaskResult and return parsed dict."""
+        from entitysdk.types import AssetLabel  # noqa: PLC0415
+
+        try:
+            recipes_dict = extraction_tr.download_json_asset_by_label(
```
eleftherioszisis: Does this asset exist?

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+            validate_params_file,
+        )
+
+        params_dir = coord_root / "config" / "params"
+        params_dir.mkdir(parents=True, exist_ok=True)
+        params_filename = "params.json"
+        params_path = params_dir / params_filename
+
+        # Params-file mode: validate and write the embedded content
+        if self.config.use_params_file:
```
eleftherioszisis: As commented above. If the task receives parameters, they shouldn't be optional.

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        # Dynamic builder mode: TODO — build from ion channel models
+        if not params_path.exists():
+            params_path.write_text(
+                json.dumps({"mechanisms": [], "distributions": {}, "parameters": []}, indent=4),
+                encoding="utf-8",
+            )
+            L.warning("Wrote placeholder params file. Dynamic builder not yet implemented.")
+        return params_filename
+
+    def _stage_traces(  # noqa: PLR6301
```
eleftherioszisis: Do we need to download traces for optimization?

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+
+    def _derive_mtype(self, db_client: entitysdk.client.Client) -> str:
+        """Derive mtype from the selected morphology entity."""
+        morph_entity = self.config.morphology_selection.morphology
+        entity = morph_entity.entity(db_client=db_client)
+        # Try to get mtype from the morphology entity's classification
+        if hasattr(entity, "mtype") and entity.mtype:
+            return str(entity.mtype)
+        # Fallback: use a placeholder
+        return "unknown"
```
eleftherioszisis: If the morphology has no mtypes, the task should fail imo

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+                if fp.is_file():
+                    rel = str(fp.relative_to(sonata_dir))
+                    paths[rel] = str(fp)
+            if paths:
+                db_client.upload_directory(
+                    entity_id=task_result.id,  # ty:ignore[invalid-argument-type]
+                    entity_type=TaskResult,
+                    name="sonata",
+                    paths={Path(k): Path(v) for k, v in paths.items()},
+                    label=AssetLabel.emodel_optimization_output,
```
eleftherioszisis: Is this correct?

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        for icm in self.config.parameters_selection.ion_channel_models:
+            icm_entity = icm.entity(db_client=db_client)
+            db_client.register_entity(
+                Derivation(
+                    used=icm_entity,
+                    generated=task_result,
+                    derivation_type=DerivationType.unspecified,
+                )
+            )
+
```
eleftherioszisis: Do we really need to add all these derivations when we can get the provenace by traversing the activity?

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+            # final.json is a dict of emodel_name -> list of model dicts
+            models = final_data.get(emodel_name, []) if isinstance(final_data, dict) else []
+            if models:
+                best = models[0]
+                score = float(best.get("fitness", best.get("score", 0.0)))
+
+        # Get species and brain region entities
+        species_entity = morph_entity.species if hasattr(morph_entity, "species") else None
+        brain_region_entity = (
+            morph_entity.brain_region if hasattr(morph_entity, "brain_region") else None
```
eleftherioszisis: brain_region is mandatory for morphologies

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+            with final_path.open(encoding="utf-8") as f:
+                final_data = json.load(f)
+            # final.json is a dict of emodel_name -> list of model dicts
+            models = final_data.get(emodel_name, []) if isinstance(final_data, dict) else []
+            if models:
+                best = models[0]
+                score = float(best.get("fitness", best.get("score", 0.0)))
+
+        # Get species and brain region entities
+        species_entity = morph_entity.species if hasattr(morph_entity, "species") else None
```
eleftherioszisis: species is accessible via subject

obi_one/scientific/tasks/emodel_optimization/task2_emodel_optimization/task.py
```diff
@@ -0,0 +1,608 @@
...
+        brain_region_entity = (
+            morph_entity.brain_region if hasattr(morph_entity, "brain_region") else None
+        )
+
+        emodel_entity = db_client.register_entity(
+            EModel(
+                species=species_entity,  # ty:ignore[invalid-argument-type]
+                brain_region=brain_region_entity,  # ty:ignore[invalid-argument-type]
+                iteration="0",
```
eleftherioszisis: Do we hardcode that to "0"?

obi_one/scientific/tasks/emodel_optimization/task3_export_and_validation/blocks.py
```diff
@@ -0,0 +1,159 @@
...
+        default="SSCX",
+        title="Brain region",
+        description="Brain region for the model (e.g. ``SSCX``).",
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.STRING_INPUT},
+    )
+    etype: str = Field(
+        title="E-type",
+        description="Electrical type tag (must match optimisation stage).",
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.STRING_INPUT},
+    )
```
eleftherioszisis: Aren't these all extracted from memodel?

obi_one/scientific/tasks/emodel_optimization/task1_efeature_extraction/blocks/protocol_and_feature_selection.py
```diff
@@ -0,0 +1,485 @@
...
+        title="Protocols",
+        description=(
+            "Protocols to extract features from. Defaults mirror the L5PC"
+            " example; the frontend can repopulate this from the catalogue and"
+            " the protocols discovered from the recordings' NWB files."
+            " Ignored when autoselect is enabled."
+        ),
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.SELECT_EFEATURES_BY_PROTOCOL,
+            "available_efeatures_by_protocol": available_features_by_protocol_name(),
```
james-isbister: What is the motivation for "available_features_by_protocol_name" and using it here? This is the same information that is available in the Protocol Union

obi_one/scientific/tasks/emodel_optimization/task1_efeature_extraction/blocks/protocol_and_feature_selection.py
```diff
@@ -0,0 +1,485 @@
...
+            type_name = "int"
+        elif f.type is bool or f.type == "bool":
+            type_name = "bool"
+        else:
+            type_name = "str"
+        result[f.name] = {"default": f.default, "type": type_name}
+    return result
+
+
+def available_efeatures_by_category() -> dict[str, list[str]]:
```
james-isbister: How do you foresee these being added to the request body by the frontend? At the moment there is no way to input them into the schema

I would actually leave this out for now, and get the basics working as @jdcourcol suggested

Using strings like this in the json extra schema is definitely not the way anyway, we should use the Pydantic pattern we're already using in obi_one/scientific/tasks/emodel_optimization/task1_efeature_extraction/protocols_and_features/protocols.py

In theory the Protocols could inherit from a base class with all the efeatures defined, then the subclasses could list which ones to show immediately in the json extra schema

To get things working I would just leave the protocols as they are and remove available_efeatures_by_category from the json_extra_schema

We could add categories, and which to show by default and which to hide in a next iteration, rather than trying to do everything at once

Let me know if its useful to discuss further

obi_one/scientific/tasks/emodel_optimization/task1_efeature_extraction/blocks/protocol_and_feature_selection.py
```diff
@@ -0,0 +1,485 @@
...
+    APWaveform,
+    IDrest,
+    IDthresh,
+    Protocol,
+    ProtocolUnion,
+    available_features_by_protocol_name,
+)
+
+
+def _efel_settings_defaults() -> dict[str, dict]:
```
james-isbister: This is not how types are represented or checked in OpenAPI or Pydantic

obi_one/scientific/tasks/emodel_optimization/task1_efeature_extraction/blocks/protocol_and_feature_selection.py
```diff
@@ -0,0 +1,485 @@
...
+        title="Threshold-based amplitudes",
+        description=(
+            "When enabled, extraction uses relative amplitudes (% of rheobase)"
+            " from per-protocol ``extraction_amplitudes``. When disabled (default),"
+            " amplitudes are auto-discovered in absolute nA from the NWB."
+        ),
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.BOOLEAN_INPUT},
+    )
+
+    autoselect: bool = Field(
```
james-isbister: Where does the frontend get the information to fill out features and protocols?

Also this would be part of the SELECT_EFEATURES_BY_PROTOCOL not a bool at this level 

I think this would also be better to take out and leave to a next iteration

obi_one/scientific/tasks/emodel_optimization/task1_efeature_extraction/blocks/protocol_and_feature_selection.py
```diff
@@ -0,0 +1,485 @@
...
+        title="Automatically fill the features and protocols",
+        description=(
+            "When enabled, protocols and features are selected automatically"
+            " using BluePyEModel's auto_targets presets. Manual protocol/feature"
+            " selection below is ignored."
+        ),
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.BOOLEAN_INPUT},
+    )
+
+    auto_targets_presets: str = Field(
```
james-isbister: I don't think we want the user to fill out a comma separated list

obi_one/scientific/tasks/emodel_optimization/task1_efeature_extraction/blocks/settings.py
```diff
@@ -0,0 +1,170 @@
...
+    threshold_nvalue_save: int | list[int] = Field(
+        default=1,
+        title="Min values to save feature",
+        description=(
+            "bluepyefe ``threshold_nvalue_save``: minimum number of recordings a feature"
+            " must be measured on to be kept in the output."
+        ),
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.INT_PARAMETER_SWEEP},
+    )
+    pickle_cells: bool = Field(
```
james-isbister: I thought we weren't using pickle files?

obi_one/scientific/tasks/emodel_optimization/task1_efeature_extraction/blocks/settings.py
```diff
@@ -0,0 +1,170 @@
...
+            " threshold_based is enabled and rmp_protocol_name is set."
+            " Set to 0 to skip."
+        ),
+        json_schema_extra={SchemaKey.UI_ELEMENT: UIElement.FLOAT_PARAMETER_SWEEP},
+    )
+
+    # ------------------------------------------------------------------
+    # Validation hold-out (carried forward to Workflow A via recipe)
+    # ------------------------------------------------------------------
+    validation_protocols: str = Field(
```
james-isbister: Again, do we really want the user to fill a comma separated list?

james-isbister (commented):
I have focused on ProtocolAndFeatureSelection which is the main UI element which will be implemented for the extraction task.

There are some major questions outstanding

If we want it to be assigned for UI development tomorrow, I suggest we simplify and focus on an initial implementation that allows users to select the standard features by protocol.

Then we can iterate with allowing users to autofill, select extra features etc. This will allow us to get an MVP working initially and give us time to make the correct choices for the expanded functionality

PR #902, opened 2026-07-14
Implement Build Synaptome SONATA backend

ilkilic:
Implements https://github.com/openbraininstitute/prod-singlecell-simulation/issues/298

This is a stacked PR based on `build-synaptome` (https://github.com/openbraininstitute/obi-one/pull/878)


james-isbister (approved):
Approving — none of the below is blocking. Recording it here as follow-up so it isn't lost.

Solid piece of work. The error handling is genuinely good — every failure is attributed to a named synapse group, and validating the artifact by loading it back through BluePySnap is the right acceptance test rather than checking that files exist. Test coverage is thorough. The `afferent_segment_offset`-as-physical-µm fix and `model_type=virtual` on generated nodes are both correct per the SONATA spec.

Two seeding issues below are the ones I'd want resolved before anyone actually simulates one of these circuits.

## Worth fixing before this is used in anger

**1. Two synapse groups on default settings get identical synapse positions.** `MorphologyLocationsBlock.random_seed` defaults to `0`, and `generate_neurite_locations_on` does a global `np.random.seed(seed)` (`morphology_locations.py:381`). So two placement blocks left at their defaults — same morphology, same `section_types`, same count — produce byte-identical location sets, and two groups pointing at the *same* placement block certainly do. The natural first thing a user builds (excitatory basal + inhibitory basal) silently stacks every inhibitory synapse on top of an excitatory one. Both existing multi-group tests dodge this by using distinct seeds (1/9) *and* distinct section types, so nothing catches it. Mixing the group index into the seed inside `build_synaptome_artifact`, or refusing duplicate `(placement block, seed)` across groups, would close it.

**2. The default synaptic parameters are near-perfectly correlated across synapses.** This is base-branch code (`tsodyks_markram.py::sample`), but this PR is what writes the numbers into a simulatable circuit, so flagging it here. Every distribution defaults to `random_seed=1` and each builds its own `np.random.default_rng(seed)`, so they all draw the same underlying stream. Measured over n=1000 with the built-in defaults:

```
corr(u_syn, delay)              = 0.96
corr(decay_time, u_syn)         = 0.77
corr(facilitation, depression)  = 0.22
decay_time pinned at its min:    51%   (Normal(mean=1.7) with min=1.7, clipped not resampled)
u_syn pinned at a bound:         32%
```

A synapse with a long delay reliably has a high `u_syn`. That's not a plausible synapse population. Two fixes, both cheap: derive per-parameter seeds from one campaign seed, or pass a single shared `rng` through `sample()` (`sample_with_constraints` already accepts one). Separately, `_DEFAULT_DECAY_TIME = Normal(min=1.7, max=1.9, mean=1.7, sd=0.1)` puts the mean *on* the lower bound, so half the values clip to exactly 1.7 — possibly meant to be `mean=1.8`.

**3. `build_synaptome.py:120` — `morphology.sections[section_id - 1]` is unguarded against `section_id == 0`.** Soma is 0 in the SONATA convention, and `sections[-1]` silently returns the *last* section rather than raising, so the `except (IndexError, KeyError)` right below never fires. Today `SectionType = Literal[3, 4]` keeps soma out, but that's a constraint one file away from this one. A `if section_id < 1: raise` costs one line.

## Worth discussing

**4. The placement morphology and the simulated morphology are downloaded by two independent paths.** `build_synaptome.py:268` loads via `CellMorphologyFromID.morphio_morphology` (asc → swc → h5), while `stage_circuit` gets its copy from entitysdk's `download_memodel` (asc → swc). Every `afferent_section_id` is computed against the first and consumed against the second. They agree only because both happen to prefer asc-then-swc — nothing enforces it, and the new h5 tier exists in one list and not the other. Reading the morphology back out of the staged circuit after `stage_circuit` would make the guarantee structural instead of coincidental, and would also stop downloading the morphology twice.

**5. `cell_morphology_from_id.py:27` — the h5 tier doesn't filter on `asset.label`.** `AssetLabel.morphology_with_spines` assets are also `application/x-hdf5`, so on an entity with no asc/swc, `next(...)` can hand back the spiny morphology (different section count, different indices) or pick arbitrarily between two h5 assets depending on API ordering. Worth an `asset.label == AssetLabel.morphology` filter.

**6. Virtual node populations can contain sources with no edges.** `source_count = int(source_ids.max()) + 1` (`build_synaptome.py:317`), but `map_presynaptic_ids` assigns source ids by `randint`, so ids can be missing from the middle; `iloc[:count]` truncation can drop more. `validate_synaptome_artifact` only checks `min >= 0` / `max < expected_sources`, so this passes. Harmless for spike replay, but "number of presynaptic cells" won't read correctly off the population size. Densifying with `pd.factorize` would fix it.

**7. `clustered.py:69,108,170` — `ceil` then `iloc[:count]` always truncates the *last* cluster.** With `count=10, n_clusters=3` you get 4/4/2 rather than a balanced or randomly-thinned split; for the grouped variants it can remove whole source groups. A random subsample of `count` rows would be unbiased for the same effort. Also `ClusteredPathDistanceMorphologyLocations` still doesn't `.drop(columns=[_CEN_IDX])` while its two siblings do — and `test_clustered_placement_rounds_up_then_truncates` now freezes that asymmetry in with `drops_center=False`. If it's deliberate it deserves a comment; if not, the test is locking in a bug.

**8. `morphology_locations.py:305` — the `.to_numpy()` removal looks like a no-op.** I checked it against non-monotonic, duplicated and non-integer indices: identical results either way, because `df[col] = series` takes the fast path when the index already matches. If it was fixing something real, the reproducing case should be in a test; otherwise it's churn in a shared function that several placers depend on.

## Minor

- `build_synaptome.py:247` — `edges.get(edges.ids(), properties=sorted(required))` discards its result. I assume it's "prove every column is readable", which is reasonable, but it reads as dead code; a comment (or `_ = ...`) would save the next reader.
- `build_synaptome.py:115-120` — `morphio.Morphology.sections` rebuilds its list on every access, and it's accessed once per location inside the loop. At the 20k-location ceiling that's real time for a one-line hoist.
- `build_synaptome.py:264` raises `FileExistsError` while everything else in the module raises `BuildSynaptomeError`. Callers now need to catch two types.
- `build_synaptome_artifact` carries a four-rule complexity suppression (`14 > 10` complexity, 66 statements, 26 locals) plus a five-rule file-level one. The per-group body is a clean `_build_group()` extraction that would drop most of it.
- `write_edges` opens with `h5py.File(..., "a")` and needs the directory to exist; it works only because `write_virtual_nodes` runs first and does the `mkdir`. Fragile ordering.
- `config_task_map.py:330` still reads `task_cls=GenerateSimulationTask,  # placeholder until BuildSynaptomeTask is added`, so this isn't reachable through `run_tasks` yet — notebook-only. Assuming that's the next PR, but worth confirming it's tracked.

The example notebook is clean (no outputs, no credentials) and its API usage all resolves.

ilkilic:
Thanks for the detailed review. Agreed on the findings.

### Addressed in this PR

1. **Placement seeding:** Per-group seeds now use `SeedSequence([random_seed, group_index])`, so groups using the same placement settings and default seed no longer produce identical locations.

3. **Soma section IDs:** `section_id < 1` is now rejected explicitly instead of allowing `section_id == 0` to resolve to `sections[-1]`.

4. **Morphology consistency:** Build Synaptome now stages the ME-model first and loads morphology from the staged `MEModelCircuit`. Placement and the generated SONATA circuit therefore use the same morphology representation and section ordering.

5. **H5 asset selection:** The standard MorphIO loader now requires `AssetLabel.morphology` for H5 assets.

6. **Sparse source IDs:** Source IDs are densified with `pd.factorize` before writing virtual nodes and edges, so the population size matches the referenced sources.

8. **No-op morphology-location change:** The original `.to_numpy()` expression was restored.

### Follow-ups

2. **Synaptic-parameter correlations:** I agree that the parameter correlations and decay-time default should be addressed. Since this affects the shared synaptic distribution layer, I will handle it in a separate PR with dedicated tests rather than expand this Build Synaptome change.

7. **Clustered placement:** Cluster balancing and the `drop_center` asymmetry require a separate shared-placement change.

I will also track the minor maintainability items as follow-up work.

james-isbister:
Re-read after 21e82a8. The section-id and per-group-seed fixes look right — checked `nrn_order` against plain `morphio.Morphology` on a morphology with interleaved neurite types and the section order genuinely differs (`[4, 3, 4]` vs `[3, 4, 4]`), so that one was a real placement bug, not a cleanup. Nothing below is blocking; recording the outstanding questions so they aren't lost.

### Group placement now depends on dict position

`_derive_group_seed` keys on the `enumerate(config.synapse_groups.items())` index, so inserting or reordering a synapse group silently changes the placements of every group after it, and `random_seed=5` on the second group no longer produces what `random_seed=5` produces on the first — while the field description still reads "Seed used when randomly selecting morphology locations".

For an artifact that is meant to be reproducible I'd key the derivation on the group *name* rather than its position, and say in the field description that the effective seed is derived per group.

### `MorphologyLocationsTask` has the bug this PR just fixed, plus one more

`obi_one/scientific/tasks/morphology_locations.py:98`:

```python
if isinstance(self.config.initialize.morphology, Path):
    m = load_morphology_nrn_order(self.config.initialize.morphology)
else:
    m = self.config.initialize.morphology.morphio_morphology  # never called
```

The entity branch binds the method instead of calling it, so `points_on(m)` receives a bound method — and even called, `morphio_morphology` is plain `morphio.Morphology(path)`, i.e. not `nrn_order`, contradicting the branch immediately above it. Out of scope here, but this PR is what un-stubbed `morphio_morphology` (it raised `NotImplementedError` before) and what established `nrn_order` as the convention for this data, so it seems the right moment to raise it.

### Every distribution still defaults to `random_seed = 1`

Base-branch rather than this PR, but it lands on the synaptome that this PR generates. All eight distribution blocks default to seed 1 and each builds its own `np.random.default_rng(self.random_seed)`, so the default Tsodyks–Markram parameters are drawn from the same stream and come out correlated. Sampling 500 synapses from the defaults in `tsodyks_markram.py`:

```
corr(u_syn, delay)      = 0.953
corr(decay_time, u_syn) = 0.757
decay_time == min       = 50%
```

The `decay_time` figure is separate and worth a look on its own: `NormalDistribution(min=1.7, max=1.9, mean=1.7, standard_deviation=0.1)` puts the mean *on* the lower bound, so half of every draw is pinned at 1.7.

The new `_derive_group_seed` decorrelates placement across synapse groups; nothing yet decorrelates the physiology distributions from each other.

### `clustered.py`: truncation bias and the `_CEN_IDX` asymmetry

`math.ceil` + `iloc[:count]` never under-produces, which was the point, but it always truncates the tail of the *last* cluster, so cluster sizes are systematically uneven — with `n_clusters=3, number_of_locations=10` the clusters come out 4/4/2 rather than 4/3/3. Distributing the remainder across clusters would avoid that.

Separately, `ClusteredPathDistanceMorphologyLocations._make_points` still returns without `.drop(columns=[_CEN_IDX])` while its two siblings drop it, and the new test encodes that asymmetry in a `drops_center` parameter rather than resolving it. Is the extra column intentional there?

### Minor

- `validate_synaptome_artifact` (`build_synaptome.py:266`) discards the result of `edges.get(edges.ids(), properties=sorted(required))`. If the intent is "assert these are all readable", an explicit check would say so.
- `_location_edge_properties` re-indexes `morphology.sections` once per row; hoisting it out of the loop is free.
- `build_synaptome_artifact` raises a bare `FileExistsError` for the output directory but `BuildSynaptomeError` for everything else, so callers need to catch both.
- The four complexity suppressions on `build_synaptome_artifact` (`complex-structure`, `too-many-branches`, `too-many-locals`, `too-many-statements`) are load-bearing — removing them surfaces genuine complexity. Worth splitting the per-group body into a helper rather than keeping the suppressions.
- `config_task_map.py:330` still carries `task_cls=GenerateSimulationTask,  # placeholder until BuildSynaptomeTask is added`.
- `ty` reports an unused `ty: ignore[invalid-argument-type]` at `cell_morphology_from_id.py:211`.

james-isbister:
After considering the agent's comments above I think this is good to merge. There isn't a way around the seeding of morphology locations based on position, and the other comments will be looked at in future tickets/PRs

ilkilic:
Thanks for the re-read and approval. I fixed the remaining MorphologyLocationsTask entity loading issue. The distribution, clustered-placement and minor maintainability items will remain follow-ups.

PR #903, opened 2026-07-14
Include PredefinedNeuronSet in deprecated neuron sets so legacy configs open

james-isbister:
To support opening of old configs

- Adds `PredefinedNeuronSet` to `neuron_sets.deprecated` as a `DeprecatedSampleNeuronSet` + `node_set` — exactly the fields legacy configs store. It was the one old neuron set the population-typed rework never gave a backwards-compat shim, so configs using it can't be opened from the workflow table (no definition in the schema the frontend binds against; `union_tag_invalid` server-side).
- Renames the multi-population block that had taken over the name to `MultiPopulationPredefinedNeuronSet`, matching its `{Biophysical,Virtual,Point}PopulationPredefinedNeuronSet` siblings. It's in no union, so no stored config can name it and nothing needs migrating.
- Collapses the deprecated sets into one `_DEPRECATED_NEURON_SETS` group and accepts it in the **virtual** union as well as the biophysical one. They predate the taxonomy — a neuron set had no population type then — so pre-rework configs use them in either slot: `AllNeurons` (nominally biophysical) selects the virtual input population of an ME-model-with-synapses simulation.
- Keeps them **out of the point union** (and `NeuronSetReference` out of the point reference union): point-neuron support arrived with the same rework that deprecated them, so no pre-rework config can target a point population.
- Behaviour is unchanged for running: every deprecated neuron set still raises when resolved against a circuit. This only affects which stored configs load — they can now be opened and edited, then migrated.
- New `test_deprecated.py`, parametrized over all 8 deprecated neuron sets: each deserializes, resolves through the biophysical and virtual unions, has a schema definition, is rejected for a point population, and still refuses to resolve against a circuit.

### DIFF
--- a/obi_one/__init__.py
+++ b/obi_one/__init__.py
@@ -148,6 +148,7 @@ __all__ = [
     "MorphologyMetricsScanConfig",
     "MorphologyMetricsSingleConfig",
     "MorphologyMetricsTask",
+    "MultiPopulationPredefinedNeuronSet",
     "MultiPulseCurrentClampSomaticStimulus",
     "NamedPath",
     "NamedTuple",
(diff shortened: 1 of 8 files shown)


ilkilic (approved):
LGTM

PR #905, opened 2026-07-15
Separate Brian2 stimulus and simulation target default neuron sets

james-isbister:
- One default served both the simulation target and an untargeted stimulus — it named the simulation target "Default: Sugar gustatory receptor neurons" and drove the stimulus into every point neuron, tripping its 100-neuron limit.
- Split into two: the simulation defaults to all point neurons; an untargeted Brian2 Poisson stimulus defaults to the `sugar` node set. The UI still shows "Sugar…" on the stimulus; the block-name special-case is removed.
- Also caps the Poisson stimulus frequency at 150 Hz.
- Adds a generation regression test.

### DIFF
--- a/obi_one/scientific/tasks/generate_simulations/task/task.py
+++ b/obi_one/scientific/tasks/generate_simulations/task/task.py
@@ -208,7 +208,13 @@ class GenerateSimulationTask(Task):
                 if is_optional_neuronsetreference(attr_type):
                     attr_value = getattr(block, attr_name, None)
                     if attr_value is None:
-                        setattr(block, attr_name, self._default_neuron_set_ref())
+                        # A Brian2 Poisson stimulus with no target drives the `sugar` node set,
+                        # not the simulation-wide default (every point neuron); see
+                        # Brian2SimulationScanConfig.
+                        if isinstance(block, Brian2DirectPoissonStimulus):
+                            setattr(block, attr_name, self._default_stimulus_neuron_set_ref())
+                        else:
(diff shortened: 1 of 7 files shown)


obi_one/scientific/tasks/generate_simulations/task/task.py
```diff
@@ -317,6 +323,17 @@ def _default_neuron_set_ref(self) -> ALL_NEURON_SETS_REFERENCE_UNION:
...
         return default_neuron_set_ref
 
+    def _default_stimulus_neuron_set_ref(self) -> ALL_NEURON_SETS_REFERENCE_UNION:
+        """Returns the reference for the default stimulus neuron set (Brian2: the `sugar` set).
+
+        The circuit is already resolved: ``execute`` calls ``_resolve_circuit`` before it fills
+        in the missing neuron set references.
+        """
+        ref = self.config.default_stimulus_neuron_set_reference(self._circuit)  # ty:ignore[unresolved-attribute,invalid-argument-type]
+        if ref.block_name not in self.config.neuron_sets:  # ty:ignore[unresolved-attribute]
```
ilkilic, line 333: Could there be an edge case when a neuron set with this name already exists? It looks like we could check the generated sugar set but use the existing set in the simulation. Could that bypass the 100-neuron limit? Should we handle this name collision explicitly?

james-isbister: Good point. This is just a temporary stop gap whilst we fix the memory usage with the simulator, so I think its ok as is for now

ilkilic (approved):

PR #910, opened 2026-07-15
Optimize morphometrics path-length bottleneck metrics

ilkilic:
## Description

This PR addresses the slow morphometrics endpoint outliers identified in the timing investigation (https://github.com/openbraininstitute/prod-synthesize-morphologies/issues/14).

Profiling showed that H5 loading and response serialization are fast, but three NeuroM path-length metrics dominate runtime for highly branched morphologies:

| Metric | Outlier 1 | Outlier 2 |
|---|---:|---:|
| section_path_distances | ~5.8 s | ~5.9 s |
| partition_asymmetry_length | ~5.6 s | ~5.8 s |
| terminal_path_lengths | ~3.0 s | ~3.0 s |

This PR adds a narrow per-morphology cache for only those three metrics. All other metrics continue to use nm.get(...).

Result on the two outlier H5 files:

| Variant | Runtime |
|---|---:|
| Original | ~18–19 s |
| Optimized | ~3.7 s |

The optimization is intentionally limited to the profiled bottlenecks to keep the implementation small and maintainable.

### DIFF
--- a/obi_one/scientific/library/morphology_measurement_annotation.py
+++ b/obi_one/scientific/library/morphology_measurement_annotation.py
@@ -4,6 +4,7 @@ import copy
 import json
 import logging
 from collections import defaultdict
+from dataclasses import dataclass
 from functools import cache
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


dkeller9 (approved):
Looks good, the previous redundant calling of the mesurement function is fixed.

PR #912, opened 2026-07-16
Compile NMODL mechanisms on the fly before BlueRecording

### LINKED ISSUE #911: Compile NMODL mechanisms on the fly before calling BlueRecording
## Problem

`CreateExtracellularRecordingArrayTask` calls `bluerecording.compute_weights`, which internally initializes neurodamus. Neurodamus requires a pre-compiled mechanism library (`libnrnmech.dylib`/`.so`) available via `NRNMECH_LIB_PATH`.

Currently this only works if the environment already has the compiled library set up (e.g. via `.envrc` or `dev_setup.sh`). There is no guarantee the compiled library matches the circuit being processed, and in service/cluster deployments the env var may not be configured.

## Expected behavior

obi-one should compile the required `.mod` files automatically before invoking BlueRecording, so that tasks work without requiring a pre-configured environment.

## Notes

- neurodamus already ships a compilation utility with caching: `neurodamus.utils.compile_mods`
- The circuit's SONATA config declares its `mechanisms_dir`, so the required mod files are discoverable at runtime
- The compiled output should be cached within `scan_output_root/entity_cache/compiled_mods/`, consistent with the existing entity cache pattern

cattabiani:
Fix: https://github.com/openbraininstitute/obi-one/issues/911

- Add `ensure_mechanisms_compiled` to `obi_one/utils/circuit.py`
- Call it in `CreateExtracellularRecordingArrayTask` before invoking BlueRecording
- Unit tests (mocked) + integration test (real nrnivmodl compilation)
- Only compiles the circuit's own `mechanisms_dir`, not neurodamus internals


obi_one/utils/circuit.py
```diff
@@ -303,6 +305,64 @@ def copy_hoc_files(
...
+    )
+
+    input_dirs = _extract_mechanisms_dir(circuit_config_path)
+    L.info(
+        "Compiling mechanisms from %d directories into %s",
+        len(input_dirs),
+        cache_dir,
+    )
+
+    cache_dir.mkdir(parents=True, exist_ok=True)
```
eleftherioszisis: cache_dir? Where does that live?

james-isbister (commented):
LGTM. There is just a repetition of the `circuit_config_path = ...` in the executre() function

I imagine you've tested the notebook runs, so otherwise good to merge if so

obi_one/scientific/tasks/create_recording_array/create_recording_array.py
```diff
@@ -202,6 +202,19 @@ def execute(
         # Use BlueRecording to generate a weights file for the circuit and test locations
         # Using the value of self.config.initialize.calculation_method
         import numpy as np  # noqa: PLC0415
+        from neurodamus.utils.compile_mods import (  # noqa: PLC0415 # ty:ignore[unresolved-import]
```
eleftherioszisis: Why aren't you using the public CLI that is available in neurodamus? See here for an example: https://github.com/openbraininstitute/obi-one/blob/075136822abfa759874e3e8a71755eb90af687a0/obi_one/scientific/library/simulation/process.py#L165

In addition, it is not nice having the main process crashing because of nrnivmodl segfaulting.

obi_one/utils/circuit.py
```diff
@@ -303,6 +305,64 @@ def copy_hoc_files(
...
+    options = Options(simulator=simulator, incflags=incflags, loadflags=loadflags)
+    env = _build_mod_files(input_dirs, cache_dir, nrnivmodl_path=nrnivmodl_path, options=options)
+
+    nrnmech_lib_path = env["NRNMECH_LIB_PATH"]
+    os.environ["NRNMECH_LIB_PATH"] = nrnmech_lib_path
+    L.info("NRNMECH_LIB_PATH set to %s", nrnmech_lib_path)
+
+    if "CORENEURONLIB" in env:
+        os.environ["CORENEURONLIB"] = env["CORENEURONLIB"]
```
eleftherioszisis: Do not change the process-wide global state. Use a subprocess instead with the desired env.

obi_one/utils/circuit.py
```diff
@@ -303,6 +305,64 @@ def copy_hoc_files(
...
+    Returns:
+        The absolute path to the compiled ``libnrnmech`` shared library.
+    """
+    from neurodamus.utils.compile_mods import (  # noqa: PLC0415 # ty:ignore[unresolved-import]
+        Options,
+        _build_mod_files,  # noqa: PLC2701
+        _extract_mechanisms_dir,  # noqa: PLC2701
+    )
+
+    input_dirs = _extract_mechanisms_dir(circuit_config_path)
```
eleftherioszisis: The neurodamus cli already handles that, doesn't it?

eleftherioszisis (changes requested):

obi_one/utils/circuit.py
cattabiani: it is written in the first file:

`self.config.scan_output_root / "entity_cache" / "compiled_mods",`

do you prefer another place? Let me know

obi_one/scientific/tasks/create_recording_array/create_recording_array.py
cattabiani: because using subprocess for something that I can do directly feels awkward. Anyway, done

obi_one/utils/circuit.py
cattabiani: ok, I see that you like subprocesses. Got it

cattabiani: also, this means that I need to run bluerecording in a subprocess. This is a bigger change than what was required. BlueRecording expects this variable to be set. I will run that with a subprocess

cattabiani: actually this is going to be a bigger rework than initially expected. Are you sure you want to proceed in this way? 

I need to: 
1. `ensure_mechanisms_compiled` returns the env dict (`NRNMECH_LIB_PATH`, etc.) without setting os.environ (ok)
2. Write electrode positions to a temporary CSV (format: index=name, columns x/y/z/type) (bad)
3. Call bluerecording write_weights <circuit_config> <electrode.csv> <output.h5> as a subprocess, passing the compiled env via subprocess.run(env={**os.environ, **compiled_env}) (ugly, to me)
4. Remove all in-process bluerecording/neurodamus imports from the task (good)

I personally do not like to write temporary files just to serve them to a cli.

obi_one/utils/circuit.py
cattabiani: removed already

obi_one/utils/circuit.py
eleftherioszisis: If a subprocess is difficult to use here, you can also change the env vars temporarily with a context manager.

cattabiani: i am using subprocess now. I let you check

eleftherioszisis (dismissed):

eleftherioszisis (approved):

PR #915, opened 2026-07-17
brian2: support `connection_override`

mgeplf:
* add support to Brian2 for SONATA `connection_overrides` block (see https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#connection-overrides)
* Only `weight` and `synapse_delay_override` supported
* use an `Event` style heapq with the next stopping point; run the simulation from one stopping point until the next one


obi_one/scientific/library/simulation/brian2/simulate_brian2.py
```diff
@@ -412,7 +424,7 @@ def _get_inputs(
 
 def _get_reports(
     simulation: bluepysnap.Simulation, neurons: brian2.NeuronGroup
```
james-isbister, line 443: Maybe could just add a docstring to describe the purpose is of the newly returned `id_mapping`

james-isbister (commented):
LGTM, just one small suggestion

obi_one/scientific/library/simulation/brian2/simulate_brian2.py
mgeplf: done; I also added some other docstrings, and a raise an error when a stimulus isn't supported.

james-isbister (approved):
LGTM!

PR #916, opened 2026-07-20
Emodel 01 efeature extraction

james-isbister:
Adds **task 1 of the e-model building pipeline — e-feature extraction**: extracting electrophysiology features from electrical cell recordings (NWB) via BluePyEModel / BluePyEfe.

### What's included
- **`task1_efeature_extraction` ScanConfig + Task** — protocols, features and eFEL settings as composable blocks; runs BluePyEModel's `extract_save_features_protocols`.
- **eFEL settings cascade** — resolved *feature > protocol > global*.
- **`/mapped-electrical-cell-recording-properties` endpoint** — per-recording protocols and amplitudes-by-protocol, used to drive feature/amplitude selection.
- **Framework wiring** — registered in the scan-config/task unions, config→task map, `types` and public exports, plus app mappings and the optional `emodel` extra (`bluepyemodel`).
- **UI schema** — new `FLOAT_OPTIONAL` and `SELECT_EFEATURES_BY_PROTOCOL` elements (with docs + tests), and an example notebook under `examples/obi_one/…`.


obi_one/scientific/from_id/electrical_cell_recording_from_id.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def download_nwb_asset(
+        self,
+        dest_dir: Path = Path(),
+        db_client: entitysdk.client.Client = None,  # ty:ignore[invalid-parameter-default]
+    ) -> Path:
+        output_dir = create_dir(dest_dir)
+        asset = db_client.fetch_assets(
+            self.entity(db_client=db_client),
+            selection={"content_type": ContentType.application_nwb},
```
eleftherioszisis: You could also include the label here.

app/endpoints/electrical_cell_recording_properties.py
```diff
@@ -0,0 +1,66 @@
...
+    "/mapped-electrical-cell-recording-properties",
+    summary="Mapped electrical cell recording properties",
+    description=(
+        "Return a dictionary of mapped ``ElectricalCellRecording`` properties for the"
+        " requested recordings, keyed by ``ElectricalCellRecordingMappedProperties``: the"
+        " protocol (ecode) names per recording and their union, plus the per-protocol step"
+        " amplitudes (nA) discovered from the recordings' NWBs."
+    ),
+)
+def mapped_electrical_cell_recording_properties_endpoint(
```
eleftherioszisis, line 29: I am concerned that this endpoint potentially downloads and reads large nwb files on the service side. Is there any estimate of how much memory is required per request? In addition, can we add these properties in the metadata instead of fetching them each time with expensive requests?

darshanmandge: The files can be huge, sometimes 100s of MB and more. I have not done the memory assessment of the requests.  However, the extraction task downloads it again when the extraction is launched. The main purpose of the endpoint is to display available amplitudes for the user to select, rather than requiring the user to manually enter the values.

Just FYI, each electrical cell recording contains 100s of traces from different protocols with varying amplitudes. There are 2247 recordings on the platform, and we plan to add 1000's more. Moreover, when registering a new recording, we will need to find these and store them. If you prefer, we can store them and any additional metadata that may be useful later.

obi_one/scientific/from_id/electrical_cell_recording_from_id.py
darshanmandge: Done. thanks.

app/endpoints/electrical_cell_recording_properties.py
eleftherioszisis: Given that the information is static with respect to the file in the asset, I believe it would be better indeed to store that information as metadata so that it is calculated once on registration and then fetched for free. Otherwise, we induce costs and potential performance issues for something that doesn't really need to be calculated on the fly.

obi_one/scientific/tasks/emodel_building/task1_efeature_extraction/config.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+        L.info("Campaign entity registered: %s", campaign.id)
+        return
+
+    def create_campaign_generation_entity(  # noqa: PLR6301
+        self,
+        generated: list,  # noqa: ARG002
+        db_client: Client,  # noqa: ARG002
+    ) -> None:
+        return None
```
bilalesi: this is preventing to create the campaign generation, the base class already calling 
`self.form.create_campaign_generation_entity`,  so this will override it and nothing will be generated

darshanmandge: I removed this override in the latest commit. Is it fine?

app/endpoints/electrical_cell_recording_properties.py
GianlucaFicarelli: Questions and points to help deciding:
- What would be the size of the additional metadata?
- How can we be sure that the metadata are kept  in sync with the assets?
- It should be considered that these metadata would become part of the model, so they are returned with the json model even if not needed.
- The current code in the PR uses `client.fetch_assets` so it should be able to access the assets in the s3 mount, but only if they are public. They are downloaded by s3-mount when accessed, and cached until there is space in the cache.

darshanmandge: Thanks for the analysis, @GianlucaFicarelli.
- These will probably be small dictionaries of the form `protocol_name: [amplitudes]`. Each protocol can have 100s of traces in multiple repetitions and sweeps
-  The nwb data will ideally not change. So, I assume the metadata sync issue isn't a concern.
- Yes, it is fine. The model already returns stimuli entries in the metadata.
- Private data can't be mounted by `fetch_assets` if it's part of a user's project?

GianlucaFicarelli: > Private data can't be mounted by fetch_assets if it's part of a user's project?

It's technically doable, although all the private data should be mounted at the same time, and the obi-one service would be the only safeguard against unauthorized access to other projects.
It makes the risk higher in case of bugs in obi-one, although it can be acceptable as long as only entitysdk is used to access those data.

darshanmandge: Thanks for the info. We will use public data initially.

Since the [plan](https://github.com/openbraininstitute/prod-build-emodel/issues/207) is to add amplitudes to the metadata, should I go ahead and open a PR in entitycore?

GianlucaFicarelli: > These will probably be small dictionaries of the form protocol_name: [amplitudes]. Each protocol can have 100s of traces in multiple repetitions and sweeps

@darshanmandge could you make an example to understand better the amount of data expected for a single cell recording?

obi_one/scientific/tasks/emodel_building/task1_efeature_extraction/config.py
james-isbister: Indeed. create_campaign_entity_with_config and create_single_entity_with_config should also be removed from this file

app/endpoints/electrical_cell_recording_properties.py
darshanmandge: - Something similar to this: 
`stimuli_amplitudes: {APThreshold: [0.1], APWaveform: [0.0, 0.1, 0.103, 0.122, 0.137, 0.141, 0.159, 0.162, 0.181, 0.184, 0.2], Delta: [0.0, 0.003], GenericStep: [0.05, 0.053], IDRest: [-0.003, 0.053, 0.081, 0.103, 0.131, 0.153], IDThresh: [0.05, 0.053, 0.056, 0.059, 0.062, 0.066, 0.069, 0.072, 0.075, 0.078, 0.081, 0.084, 0.087, 0.094, 0.097, 0.103, 0.106], IDDepol: [0.078, 0.1, 0.128], IDHyperpol: [0.084, 0.106, 0.134], IRDepol: [-0.003, 0.0], IRHyperpol: [-0.009, -0.006], IV: [-0.003, -0.002, 0.0, 0.003, 0.006], SpikeRec: [0.0, 0.002, 0.003], Spontaneous: [0.0], SAHP: [0.0]},`
- Another metadata entry would be for `rheobase`, e.g.  `rheobase_current: {value: 0.05, unit: nA}`, which would help us to implement amplitude selection of relative protocol amplitude (relative_amplitude =  amp_nA/rheobase).

eleftherioszisis: What would be the worst case scenario with respect to the size of that dictionary?

darshanmandge: The total amplitude counts across all protocols combined can be in 100s. A single NWB file can contain experiments with several protocols, each with its own repetitions and sweeps.

eleftherioszisis: Hmm, @GianlucaFicarelli was concerned that the read_many payload will grow very large. With such a big number of protocols I believe he was right to be concerned.

Would storing them as a json file asset and open that instead on the service-side do the trick?

darshanmandge: > Would storing them as a json file asset and open that instead on the service-side do the trick?

Yes, a JSON asset can work. So, not as a separate table as morphometrics?

bilalesi (approved):

PR #917, opened 2026-07-20
brian2: Delay and duration for Poisson stimulus

mgeplf:
* Add `delay` and `duration` support for Poisson stimulation
* this required a move to event based start and end Poisson stimulation
* add option for internal profiling


obi_one/scientific/library/simulation/brian2/simulate_brian2.py
```diff
@@ -570,41 +562,189 @@ def _create_synapses(
...
+            for start, end in selections.ranges:
+                p = brian2.PoissonInput(
+                    target=net.neurons[start:end],
+                    target_var="v",
+                    N=1,
+                    rate=self.config.rate * brian2.units.Hz,
+                    weight=self.config.weight * brian2.units.mV,
+                )
+                # no refractory period for Poisson targets
+                net.neurons[start:end].rfc = 0 * brian2.units.ms
```
james-isbister: What's the reason for setting the refractory period to 0? Did they do that in the original code?

obi_one/scientific/library/simulation/brian2/simulate_brian2.py
```diff
@@ -570,41 +562,189 @@ def _create_synapses(
...
+
+        assert self.config.delay + self.config.duration <= t
+        L.debug("Removing InputPoisson: %s: (delay: %s, duration: %s): rate: %f weight: %f",
+                self.config.node_set,
+                self.config.delay,
+                self.config.duration,
+                self.config.rate,
+                self.config.weight,
+                )
+        ret = NetworkOperation(add=[], remove=self._inputs)
```
james-isbister, line 736: Does the refractory period get reset outside the stimulus window with this removal operation?

james-isbister: I imagine that could lead to a tonne of spikes

james-isbister (commented):
Looks good. Just a question about the refractory period

obi_one/scientific/library/simulation/brian2/simulate_brian2.py
mgeplf: Good question; it's done in the original repo:
https://github.com/philshiu/Drosophila_brain_model/blob/main/model.py#L92

obi_one/scientific/library/simulation/brian2/simulate_brian2.py
mgeplf: it does not at the moment, is that what you want it to do?  I don't have a mechanism for handling resets like that at the moment, so I'd need to add it

obi_one/scientific/library/simulation/brian2/simulate_brian2.py
mgeplf: As discussed, I have removed setting the refractory period to zero.

james-isbister (approved):

PR #920, opened 2026-07-22
feat(circuit-viz): allow 'single' scale circuits

pgetta:
## What

Adds `CircuitScale.single` to the scale allowlist in `circuit_asset_id`, so `/circuit/viz/{id}/nodes` and `/circuit/viz/{id}/morphologies/{path}` accept single-neuron circuits.

## Why

This unblocks reusing the shared 3D circuit viewer on synaptome (beta) pages, which are circuits with `scale='single'`. Today those pages run a separate in-browser SONATA/h5wasm loader that duplicates this endpoint; the frontend side of that consolidation is tracked in [openbraininstitute/prod-singlecell-simulation#284](https://github.com/openbraininstitute/prod-singlecell-simulation/issues/284).

### DIFF
--- a/app/services/circuit_visualization.py
+++ b/app/services/circuit_visualization.py
@@ -35,12 +35,12 @@ def circuit_asset_id(client: Client, circuit_id: UUID) -> UUID:
             },
         ) from e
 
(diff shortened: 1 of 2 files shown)


ilkilic (approved):
Thanks! LGTM

PR #932, opened 2026-07-29
Polish morphology location schema text

ilkilic:
Address https://github.com/openbraininstitute/obi-one/pull/498#pullrequestreview-4805501353

## Changes

- Rename the default morphology-location reference label to `Default: No Locations`
- Group NEURON neuron sets and morphology locations under `Targets`
- Add readable titles for morphology-location block types
- Improve descriptions for morphology-location parameters and block dictionaries
- Add micrometer units to distance-based morphology-location parameters


obi_one/scientific/blocks/morphology_locations/clustered.py
```diff
@@ -18,19 +21,27 @@
...
     )
     cluster_max_distance: float | list[float] = Field(
-        title="Cluster maximum distance",
-        description="Maximum distance in um of generated locations from the center of their \
-            cluster",
+        title="Cluster Maximum Distance",
+        description=(
+            "Maximum soma path distance between each generated location and the center of "
+            "its cluster."
+        ),
```
james-isbister, line 46: IIRC the synapses are distributed uniformly - that could be mentioned here perhaps.

Also the title is not explicit. Maybe it should be "Maximum distance from cluster center"?

Also could we just double check that that is correct?

obi_one/scientific/blocks/morphology_locations/clustered.py
```diff
@@ -96,26 +109,31 @@ class ClusteredPathDistanceMorphologyLocations(ClusteredMorphologyLocations):
     groups within each cluster. This exposes the full possible complexity.
     """
 
+    title: ClassVar[str] = "Clustered Path Distance Morphology Locations"
+
     path_dist_mean: float | list[float] = Field(
-        title="Path distance mean",
-        description="Mean of a Gaussian, defined on soma path distance in um. Used to determine \
-            locations.",
+        title="Path Distance Mean",
```
james-isbister: Maybe the title could be "Cluster Path Distance Mean" and below "Cluster Path Distance Standard Deviation"

obi_one/scientific/blocks/morphology_locations/clustered.py
```diff
@@ -69,6 +80,8 @@ class ClusteredGroupedMorphologyLocations(
 ):
     """Clustered random locations, grouped in to conceptual groups."""
```
james-isbister, line 87: I think the grouping for ClusteredGroupedMorphologyLocations and RandomGroupedMorphologyLocations doesn't add anything for our current use case? As the grouping isn't used for the injection stimuli. It is more useful for creating synaptomes / circuits or in the future if we add more complex stimuli for morphology location stimuli

I think these two can be removed from the union for simulations for now?

obi_one/scientific/blocks/morphology_locations/clustered.py
```diff
@@ -96,26 +109,31 @@ class ClusteredPathDistanceMorphologyLocations(ClusteredMorphologyLocations):
...
         json_schema_extra={
             SchemaKey.UI_ELEMENT: UIElement.FLOAT_PARAMETER_SWEEP,
+            SchemaKey.UNITS: Units.MICROMETERS,
         },
     )
     n_groups_per_cluster: int | list[int] = Field(
         default=1,
-        title="Number of groups per cluster",
-        description="Number of conceptual groups per location cluster to generate",
+        title="Number of Groups per Cluster",
```
james-isbister: "Groups per Cluster" is maybe enough here as the title?

obi_one/scientific/blocks/morphology_locations/path_distance.py
```diff
@@ -13,20 +16,27 @@
...
     )
     path_dist_tolerance: float | list[float] = Field(
-        title="Path distance tolerance",
-        description="Amount of deviation in um from mean path distance that is tolerated. Must be \
-            > 1.0",
+        title="Path Distance Tolerance",
+        description=(
+            "Allowed deviation from the target soma path distance. Must be greater than 1.0."
+        ),
         json_schema_extra={
```
james-isbister: Uniformly distributed?

obi_one/scientific/blocks/morphology_locations/clustered.py
```diff
@@ -96,26 +109,31 @@ class ClusteredPathDistanceMorphologyLocations(ClusteredMorphologyLocations):
     groups within each cluster. This exposes the full possible complexity.
     """
```
james-isbister, line 116: I think it should describe something like: "Locations uniformly distributed within clusters, with cluster centers normally distributed around a given path distance." - my version is not well written but I think it should express this ideally 

I think we can drop: "This exposes the full possible complexity"

obi_one/scientific/blocks/morphology_locations/path_distance.py
```diff
@@ -13,20 +16,27 @@
 class PathDistanceMorphologyLocations(MorphologyLocationsBlock):
     """Locations around a specified path distance."""
```
james-isbister: "Locations uniformly distributed around a specified path distance"?

obi_one/scientific/blocks/morphology_locations/clustered.py
```diff
@@ -18,19 +21,27 @@
 class ClusteredMorphologyLocations(MorphologyLocationsBlock):
     """Clustered random locations."""
```
james-isbister, line 26: Something like "Clustered locations with cluster centers randomly distributed across the selected neurite types"?

james-isbister (commented):
Looks good. Just some minor suggestions. 

I think we can go for Targets for now, until we ever come up with something else!

obi_one/scientific/blocks/morphology_locations/clustered.py
ilkilic: Good point. I updated the title to `Maximum Distance from Cluster Center` and clarified that locations are sampled uniformly from valid morphology intervals within that distance. I double-checked it, in `select_places_from_candidate_list`: candidate intervals are weighted by interval length then offsets are sampled with np.random.rand(...) within [min, max].

obi_one/scientific/blocks/morphology_locations/clustered.py
ilkilic: Correct, it is not used for compartment-set. I removed them.

obi_one/scientific/blocks/morphology_locations/clustered.py
ilkilic: Updated to: `Locations are uniformly sampled within clusters whose centers are normally distributed around the specified soma path distance.`

PR #933, opened 2026-07-30
Move task config types to task mapping

james-isbister:
* Move ClassVars `campaign_task_config_type`, `campaign_generation_task_activity_type`, `single_task_config_type`, `single_task_activity_type` to `config_task_map.TASK_MAP`
* Change  TASK_MAP entries from 3-tuple to a TaskRegistration dataclass
* Update scan generation logic to use this dataclass rather than original ClassVars
* Make it's simpler to add new task, all the additional types are in one place

Extra:
* Remove contribute scan configs and task, which did nothing and weren't used
* Add synapse parameterization to TASK_MAP and app/mappings.py


obi_one/core/registry.py
```diff
@@ -5,63 +5,95 @@
...
+        for klass in config_cls.__mro__:
+            registration = self._by_config_cls.get(klass)
+            if registration is not None:
+                return registration
+        return None
 
     def get_configs_task_type(self, config: object) -> type:
         """Return the Task class for a given config instance."""
-        return self._config_task_map[config.__class__]
+        registration = self._by_config_cls.get(config.__class__)
```
ilkilic: Should this use `get_registration_for_config(type(config))`? That method also handles config subclasses I think.

james-isbister: I split _by_config_cls into `_by_single_config_cls` and `_by_scan_config_cls` as `_by_config_cls` was previously holding both scan configs and single configs, and only a single config should ever map to a Task.      `get_configs_task_type(SomeScanConfig())` now raises

The associated `get_registration_for_single_config` and `get_registration_for_scan_config` now use the MRO via a shared helper, meaning that both work for subclasses as you suggested

ilkilic (approved):
Thank you! LGTM

PR #934, opened 2026-07-30
Start dbsdk directory with entity specific registration subdirectories

james-isbister:
**Start consolidating re-usable entitysdk operations into obi_one/db_sdk/, starting with morphologies and circuits, and moving db_sdk.py**

This is a starting point, not the finished job. It establishes the package and moves the two cleanest entities across; several more are still scattered around the codebase and will follow in separate PRs. Reviewing the shape here matters more than the contents — everything after this lands in the structure this PR sets up.

Entitycore write paths are currently spread across utils/, library/, tasks/ and app/endpoints/. The target is one package, with a subfolder per major entity under registration/.

What's in this PR

- Add obi_one/db_sdk/ containing db_sdk.py plus a registration/ subpackage, one subfolder per entity
- Move obi_one/utils/db_sdk.py → obi_one/db_sdk/db_sdk.py
- Move obi_one/utils/circuit_registration/ → obi_one/db_sdk/registration/circuit/     - Move obi_one/scientific/libra.py →obi_one/db_sdk/registration/morphology/register.py, with an __init__.py re-exporting its public functions so callers never reach into the inner module                         - Fold the circuit-specific add/ add_connectivity_matrix_asset / add_image_assets helpers, and the OVERVIEW_IMAGE_NAME / SIM_DESIGNER_IMAGE_NAME constants, out of db_sdk.py into registration/circuit/assets.py                       - Delete add_circuit_folder_assregister_asset(asset_label="sonata_circuit") back in d14460f0 (the same commit that added CIRCUIT_ASSET_MAPPING) and left with no production caller
- Drop five deferred in-function db_sdk imports from registration/circuit/generate.py; the import cycle they worked around disappears now that the helpers live in assets.py     - Update all import sites acroscripts/, tests/ and threenotebooks, and move the test modules to mirror the new layout                         
No behaviour change. Every move is verbatim, and the only deletion is dead code.      
registration/README.md
                                                                                      Added to keep the follow-up worng in one PR thread. It documents the convention, what belongs in the package and what deliberately doesn't (task-side _register_output methods, endpoints, compute helpers, staging helpers like resolve_circuit), and the ranked list of candidates not yet moved:                    
1. library/simulation/neuron/registration.py → registration/simulation_result/ — also collapses a duplicate implementrian2.py:957
2. tasks/skeletonization/registration.py — but hoisting Contribution/Derivation out ofcircuit/links.py into a shared ikely the better first step
3. tasks/ion_channel_modeling.py — ~180 of 620 lines, interleaved with fitting computeextraction, not a move
4. tasks/create_recording_array/create_recording_array.py — ~48 lines inline in execute()
5. The duplicated GLB mesh upload between library/morphology_mesh.py and              app/endpoints/convert_morpholog

It also records two gotchas worth knowing before touching this code, both below.      
Notes for reviewers

- db_sdk.py drops 494 → 350 lines. It is not fully generic yet: resolve_circuit stays, since it stages a circuit for reading rather than registering one, and it keeps the   Circuit / CircuitFromID imports
- generate.py now binds the asset helpers at module level, so tests must patch        ...registration.circuit.generate assets module where they'redefined                                                                               - app/services/validator.py harite methods and asserts on callcounts — future refactors here can break config validation silently
                                                                                      Testing — make format and make 64 passed, 4 fewer than on main,exactly the four deleted add_circuit_folder_asset tests. The 5 remaining failures are pre-existing locally and come from the uninstalled ultraliser and connalysis optional extras.
                                                                                      ---
The docs-check caveat from before still stands: nothing in docs/ references the moved paths, so if that check fires, l rather than a filler edit — the new README lives in obi_one/, so it won't satisfy a docs/-scoped check.


chr-pok (dismissed):
LGTM! Seems to be purely re-organization of files, no functional changes, as intended. Nice readme file added.

chr-pok (approved):

PR #935, opened 2026-07-31
Move simulation result registration into db_sdk/registration

james-isbister:
Continues the `db_sdk/registration/` consolidation started in #934. Adds `simulation_result/` alongside the existing `circuit/` and `morphology/` packages.

* Move `scientific/library/simulation/neuron/registration.py` to `db_sdk/registration/simulation_result/register.py`
* Change `register_simulation_results` to take a `simulation_id` and plain report paths instead of the neuron-specific `SimulationMetadata` / `SimulationResults` schemas, so it is not tied to one backend
* Replace the duplicate inline registration in `simulate_brian2.sonata_main` with a call to the shared function
* Move the test module to mirror the new layout

Both backends now register under the same name and description. brian2 previously used `description=""`; it takes the shared default now. `name` and `description` stay parameterized for future backends, though nothing overrides them today.

Dropping brian2's hardcoded `application_x_hdf5` for voltage reports is safe: libsonata always appends `.h5` to report filenames when the configured name lacks it (`weird.dat` comes back as `weird.dat.h5`), so the shared extension lookup always resolves to the same content type.

Also updates `registration/README.md` to mark this candidate done and renumber the rest.

**Testing** — `make format` and `make lint` clean. Full suite 1523 passed; the 5 failures are pre-existing locally and come from the uninstalled `ultraliser` and `connalysis` optional extras.


obi_one/db_sdk/registration/simulation_result/register.py
```diff
@@ -0,0 +1,96 @@
...
+) -> None:
+    """Upload a single report file as an asset on the simulation result."""
+    asset = client.upload_file(
+        entity_id=simulation_result.id,
+        entity_type=type(simulation_result),
+        file_path=file_path,
+        file_content_type=file_content_type,
+        asset_label=asset_label,
+    )
+    L.info(
```
mgeplf: I'd put the logging before the `client.upload_file` so that we know what is being attempted, in case it fails.

obi_one/db_sdk/registration/simulation_result/register.py
```diff
@@ -0,0 +1,96 @@
...
+) -> models.SimulationResult:
+    """Register a SimulationResult and attach its spike and voltage report assets.
+
+    Args:
+        client: The entitycore SDK client.
+        simulation_id: ID of the simulation these results belong to.
+        spike_report_file: Path to the spike report (HDF5).
+        voltage_report_files: Paths to the voltage reports. The content type of each is
+            derived from its extension via EXTENSION_TO_CONTENT_TYPE.
+        name: Name for the registered entity. Both current backends take the default.
```
mgeplf: `Both current backends take the default.` is the sort of comment that will get stale very fast.

If nothing is using `name` or `description`, let's remove them for now; or, alternative, have the `backends` set the default values they want, rather than rely on `register_simulation_results` having a default.

obi_one/db_sdk/registration/simulation_result/register.py
```diff
@@ -0,0 +1,96 @@
...
+        description: Description for the registered entity. Both current backends take
+            the default.
+
+    Returns:
+        The registered SimulationResult entity.
+    """
+    # TODO: Add proper name, consider adding a description
```
mgeplf: stale?

obi_one/db_sdk/registration/simulation_result/register.py
james-isbister: Good point. I went for the backends writing it exclusively, so there is a clear place for us to write better/custom strings in future

mgeplf (approved):

PR #936, opened 2026-07-31
Additional simulation generation tests

james-isbister:
Tests for `GenerateSimulationTask`, plus two bugs they turned up.

**Tests** — 365 tests in `tests/obi_one/scientific/tasks/simulation_campaign_generation/`

- Drive the task directly (no scan, no DB): build a SingleConfig, execute, assert on the files written
- Cover every block type: neuron sets, stimuli, recordings, timestamps, distributions, morphology locations, synaptic + neuronal manipulations
- Sweep every block with its reference fields left unset (`None`), which is the most common config and where the default-resolution logic is spread across several places
- Assert no generated config names a node set that is missing from `node_sets.json`
- Parametrised sweeps build their cases from the block unions, so a new block type is covered without editing the tests
- Also: circuit resolution/staging, compartment sets, entitycore asset uploads, whole-config snapshots per config family (Circuit, MEModel, MEModel+synapses, Brian2, LearningEngine)

**Fixes**

- Point/virtual neuron set defaults moved from `NeuronSimulationScanConfig` up to `BaseSimulationScanConfig`. Brian2 and LearningEngine configs accept `PointCombinedNeuronSet` but had no `default_point_neuron_set_reference`, so leaving its base unset raised `AttributeError` instead of picking up the point default.
- Registered `NeuronalManipulationReference` in the block reference registry. It was declared in `reference_types` on `MEModelSimulationScanConfig.neuronal_manipulations` but never registered, so `ScanConfig.add()` could not add a neuronal manipulation at all.

### DIFF
--- a/obi_one/scientific/mappings_and_registry/block_reference_registry.py
+++ b/obi_one/scientific/mappings_and_registry/block_reference_registry.py
@@ -20,6 +20,9 @@ from obi_one.scientific.unions_and_references.neuron_sets import (
     PointNeuronSetReference,
     VirtualNeuronSetReference,
 )
+from obi_one.scientific.unions_and_references.neuronal_manipulations import (
+    NeuronalManipulationReference,
+)
 from obi_one.scientific.unions_and_references.recordings import RecordingReference
 from obi_one.scientific.unions_and_references.stimuli import StimulusReference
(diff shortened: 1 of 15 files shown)


chr-pok (approved):
Great addition of tests! Thanks for the bug fixes along with it.

PR #937, opened 2026-08-03
Turning benchmark file creation error into warning

chr-pok:
Small improvement for the rare case that no disk space is left after task execution on the launch-system, to avoid an unhandled exception when writing the benchmark file. This PR turns such an error into a warning. The actual benchmark results will be accessible through the launch-system logs anyway.


james-isbister (approved):
LGTM!

PR #938, opened 2026-08-03
enable debug logging in circuit extraction

mgeplf:


obi_one/scientific/tasks/circuit_extraction/task.py
```diff
@@ -358,7 +358,7 @@ def execute(  # ruff: ignore[complex-structure, too-many-locals, too-many-statem
         entity_cache: bool = False,
         execution_activity_id: str | None = None,
     ) -> str | None:  # Returns the ID of the extracted circuit
-        # Start benchmark tracking
+        logging.basicConfig(level=logging.DEBUG)
```
chr-pok, line 361: Just wondering, wouldn't this change the logging behavior globally, and never sets it back to the level it was before? Could we use `L.setLevel(logging.DEBUG)` instead which I think would change it at the module level only?

chr-pok (commented):
I think we could use the already existing settings pattern for this, by adding the log level as an option to `CircuitExtractionSettings`. To keep it clearly scoped within this task/module.

obi_one/scientific/tasks/circuit_extraction/task.py
mgeplf: `L.setlevel` only sets it for within this module, iirc, I want to have logs from `brainbuilder`, as well.

Since this is a task, and only runs for a short while, I didn't think it would be a problem to turn it on globally.

chr-pok: > `L.setlevel` only sets it for within this module, iirc, I want to have logs from `brainbuilder`, as well.

Ok, I see, `L.setlevel` won't probably work in this case then.

> Since this is a task, and only runs for a short while, I didn't think it would be a problem to turn it on globally.

I am thinking of cases where the task is not launched isolated on the launch system, but for example locally from within a notebook, in which case we don't want to have the level permanently changed.

How about we change it back at the end of the task to whatever level it was before?

PR #939, opened 2026-08-03
Add new brian2 simulation features to ScanConfig

james-isbister:


james-isbister:
Please take a look at this one when you have a chance @ilkilic @mgeplf Just noticed I hadn't marked it as ready for review

obi_one/scientific/blocks/recordings/base.py
```diff
@@ -31,29 +40,24 @@ class Recording(Block, ABC):
...
+    def recording_timestep(self) -> PositiveFloat:
+        """Interval between recorded samples in milliseconds (ms)."""
+        return self._simulation_timestep
 
     def config(
         self,
         end_time: NonNegativeFloat | None = None,
         default_node_set: str = "All",
         db_client: entitysdk.client.Client | None = None,
+        simulation_timestep: PositiveFloat | None = None,
```
mgeplf: how can this be `None`? don't all simulation have timesteps?

obi_one/scientific/blocks/recordings/base.py
```diff
@@ -89,3 +93,26 @@ def config(
     @abstractmethod
     def _generate_config(self, db_client: entitysdk.client.Client | None = None) -> dict:
         pass
+
+
+class CustomDtRecording(SimulationDtRecording, ABC):
```
mgeplf: this class hierarchy seems strange to me; `CustomDtRecording` isn't a special case of `SimulationDtRecording`, if anything, it's the other way around - `CustomDtRecording` is just a Recording, and `SimulationDtRecording` is a specialization where the Dt has no choice but to the simulation one.

Also, the `ABC` isn't required, it's already inherited from `SimulationDtRecording`

obi_one/scientific/blocks/recordings/soma.py
```diff
@@ -83,3 +92,15 @@ def _generate_config(self, db_client: entitysdk.client.Client | None = None) ->
         self._end_time = self.end_time  # ty:ignore[invalid-assignment]
 
         return super()._generate_config(db_client=db_client)
+
+
+class SomaVoltageRecording(CustomDtRecording, SimulationDtSomaVoltageRecording):
```
mgeplf: same point about the inheritance hierarchy here; why does `SomaVoltageRecording` inherit from `SimulationDtSomaVoltageRecording`?

obi_one/scientific/blocks/stimuli/brian2_poisson.py
```diff
@@ -44,6 +44,9 @@ class Brian2DirectPoissonStimulus(Block):
 
     title: ClassVar[str] = "Direct Poisson Input"
 
+    # Brian2 instantiates one `PoissonInput` per target neuron, so the target has to be small.
```
mgeplf: this isn't strictly true, it uses one Poisson per contiguous range of NodeIDs; so it's true when `[0, 2, 4, 6 ...]` - each uses one Poisson, but for [0, 1, 2, 3, 4] it would be one total.

obi_one/scientific/blocks/stimuli/stimulus.py
```diff
@@ -547,12 +551,36 @@ def _single_timestamp_stimulus_config(self, offset_timestamp: NonNegativeFloat)
...
             "frequency": self.frequency,
-            "dt": self.dt,
+            "dt": self.signal_timestep,
             "represents_physical_electrode": self._represents_physical_electrode,
         }
         return stim_dict
 
 
+class SinusoidalCurrentClampSomaticStimulus(SimulationDtSinusoidalCurrentClampSomaticStimulus):
+    """A sinusoidal current injection with a fixed frequency and maximum absolute amplitude."""
```
mgeplf: not sure if you want to put bounds on the frequency; since the simulation dt interacts w/ the rate at which currents can be injected

obi_one/scientific/blocks/recordings/base.py
james-isbister: Simplified the class hierarchy without changing the names of the pre-existing classes which are used already by configs in entitycore. 

There is now a BaseRecording and with two abstract subclasses (Recording - user defined dt, and SimulationDtRecording which uses the simulation dt). From there we have the following non-abstract subclasses:

**Recording:**
SomaVoltageRecording
TimeWindowSomaVoltageRecording
IonChannelVariableRecording

**SimulationDtRecording:**
SimulationDtSomaVoltageRecording
SimulationDtTimeWindowSomaVoltageRecording

SimulationDtRecording cannot inherit as a special case of Recording, as the dt parameter can't be taken out of the schema for the subclass.

This also removes the need for the complex multi-inheritence.

obi_one/scientific/blocks/recordings/soma.py
james-isbister: See above

obi_one/scientific/blocks/recordings/base.py
mgeplf: I think this is still worth reasoning about, do you happen to know the answer?

james-isbister: Yes, still waiting to look at that one agaiin :)

obi_one/scientific/blocks/stimuli/stimulus.py
james-isbister: Added a dynamic check for the nyquist frequency, as the timestep differs between simulators, and could be user specifiable in future

obi_one/scientific/blocks/stimuli/brian2_poisson.py
james-isbister: Updated the documentation here

obi_one/scientific/blocks/recordings/base.py
james-isbister: Yes, the neuron, learning engine and brian2 configs each set the private simulation timestep. The getter raises an error if its set to None, so None should never be passed here so I've removed the option and default value

ilkilic:
One blocking issue: these new Brian2 pulse and sinusoidal inputs document frequency in Hz but the Brian2 runner combines it with dt in milliseconds without converting units. As a result, the generated waveforms run 1000× too fast. Could we fix that conversion before merging?

ilkilic (dismissed):
Thanks for the fixes, it looks good to me now. Approved.

ilkilic (approved):

PR #943, opened 2026-08-04
brian2: make spike replay filter by target ids based on node_set; fix input current unit

mgeplf:
* previously, the `node_set` was used to filter the source IDs of the spikes
* Switch to nA for input currents


obi_one/scientific/library/simulation/brian2/simulate_brian2.py
```diff
@@ -278,13 +278,60 @@ def _get_close_spikes(ids: np.ndarray, times: np.ndarray, window: float) -> np.n
...
+    syn.pre.delay = (
+        0.0 * brian2.units.ms
+        if synapse_template.params.delay is None
+        else synapse_template.params.delay.get()
+    )
+
+    # Note: brian2 doesn't allow multiple synapses between two neurons;
+    # this matters if they are supposed to have different parameters.
+    # Currently we only support single connections.
+    assert selection.flat_size == len(syn), (
```
james-isbister: > The multi-synapse assertion is dead code, and its comment is wrong. In the new _build_synapses, the comment claims "brian2 doesn't allow multiple synapses between two neurons" and guards it with assert selection.flat_size == len(syn). I checked against the locked brian2 (2.10.1): connect(i=[0,0,1], j=[1,1,2]) yields len(S) == 3 — duplicate pairs are kept, that's what brian2's multisynaptic indexing is for. Since _build_synapses always connects exactly flat_size pairs, len(syn) always equals flat_size and the assertion can never trip. So a real multi-synapse circuit (which EM-derived connectomes routinely are) will sail straight past it. If the restriction is genuinely wanted, the check has to be on the edge data — something like len(set(zip(src, tgt))) == selection.flat_size. Worth asking whether the restriction is needed at all: with duplicate pairs brian2 creates two synapses each carrying its own w, and on_pre: v += w fires for both, which sums the weights — probably the correct point-neuron behaviour already. Being an assert, it also vanishes under python -O.

james-isbister (commented):
Looks good. Just one comment on the multi synapse assertion from the overlord. At least we can change the comment in the code if its correct

obi_one/scientific/library/simulation/brian2/simulate_brian2.py
mgeplf: I'll have a look.

> Being an assert, it also vanishes under python -O.

LLMs really like to point this out, but I don't think it's useful advice, personally, I'm not sure if I ever see things run w/ `python -O`

james-isbister (approved):

PR #950, opened 2026-08-05
Return 4xx, not opaque 500s, for configs that cannot run

james-isbister:
## Problem

Three ways a config the user has to fix surfaced as an unhandled **500 with no message**, and a Sentry alert as though the server were at fault.

A scan config is a FastAPI request body, so it is validated during dependency solving, **before any endpoint code runs** — the endpoints' own `try`/`except` never sees a failure raised there. Pydantic folds only `ValueError`/`AssertionError` raised inside a validator into a `ValidationError`; anything else escapes body validation entirely, and no generic `Exception` handler is registered in `application.py`.

Reported from production:

```
POST /api/obi-one/declared/scan_config/grid-scan-coordinate-count 500 Internal Server Error
DeprecationWarning: NeuronSetReference is deprecated. Use BiophysicalNeuronSetReference, ...
```

## Changes

**`NeuronSetReference`** raised `DeprecationWarning` from its `block` accessors, which run inside the `fill_block_references_and_names` validator. Now a `ValueError`, so FastAPI returns a 422 carrying the migration message.

**A dangling block reference** — a typo, or a block deleted while something still targets it — raised `KeyError` from the same validator. Now a `ValueError`, so its already-helpful message ("Block 'X' not found in 'Y'...") reaches the caller instead of being discarded.

**The deprecated neuron sets** raised `NotImplementedError` when resolved against a circuit. That is what obi_one raises for genuinely unimplemented abstract methods, so the generate endpoints could only map it to a 500. They now raise `ConfigValidationError`, which those endpoints map to a 422. Its `details` field repeats the message in the shape the frontend reads for a non-500, so the UI shows the reason rather than a bare "An error occurred generating the simulation campaign".

Genuine server faults still return 500.

## Before / after

| Case | Before | After |
| --- | --- | --- |
| Deprecated `NeuronSetReference` | 500, body `Internal Server Error` | 422 + migration message |
| Dangling block reference | 500, body `Internal Server Error` | 422 + "Block 'X' not found in 'Y'" |
| Deprecated neuron set (e.g. `AllNeurons`) | 500, message only in `detail` | 422 + message in `message` and `details[0].msg` |

## Notes

Ruff exempts stub methods that raise `NotImplementedError` from `unused-method-argument`; the deprecated blocks lose that exemption by raising a different type, hence the per-file ignore in `pyproject.toml`.

Needs https://github.com/openbraininstitute/core-web-app PR for the UI to display the reason on the coordinate-count path — that branch is `scan-config-error-messages`.

## Testing

1999 tests pass. The 7 failures on this machine are pre-existing and dependency-related (`ultraliser` ×5, `connalysis` ×2); verified by stashing every change and re-running. New coverage pins the 422 for both a rejected config and a dangling reference, that genuine faults still return 500, and that the deprecation is wrapped by pydantic rather than escaping validation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


obi_one/scientific/blocks/neuron_sets/deprecated.py
```diff
@@ -41,25 +42,27 @@ def deprecation_error_message(self) -> str:
             "Please use an alternative neuron set instead."
         )
 
+    # These raise ConfigValidationError rather than NotImplementedError so the service can tell
+    # "this config is not runnable, the user must migrate it" apart from "this abstract method was
+    # never implemented", which is a genuine bug. NotImplementedError is used throughout obi_one
+    # for the latter, so the generate endpoints could only map it to a 500; ConfigValidationError
+    # they map to a 422 carrying the message below.
```
bilalesi: do we need all this ai generated comments (in the whole PR) ?

bilalesi (changes requested):

obi_one/scientific/blocks/neuron_sets/deprecated.py
james-isbister: Removed comments — 2b2d6d87.

bilalesi (approved):

PR #952, opened 2026-08-06
Circuit registration: draft → async validate → assets

jdcourcol:
## Summary
- Adds HTTP `/declared/circuit/register` (draft lifecycle), `/validate`, and `/generate-assets`, plus shared launch helpers.
- Extends `register_circuit` for draft/async validation and metadata parity (duplicate/species checks, contributions/publications, image overrides, etc.).
- Ships circuit validation + asset-generation launch jobs and the task gate that blocks simulation until a circuit is `active`.

## Test plan
- [ ] `make test-file FILE=tests/app/endpoints/test_circuit_registration.py`
- [ ] `make test-file FILE=tests/obi_one/utils/test_circuit_registration.py`
- [ ] `make test-file FILE=tests/obi_one/scientific/tasks/circuit_validation/test_task.py`
- [ ] `make test-file FILE=tests/app/endpoints/test_task_circuit_gate.py`
- [ ] Smoke: register a circuit → draft → validation job → assets callback

## Notes
- Stacked follow-up: circuit customization PR will target this branch.
- Original mixed branch preserved as `circuit-customization-endpoint`; full pre-split snapshot at `backup/pre-split-complete`.
- `main` has moved registration under `obi_one/db_sdk/`; this branch still uses `obi_one/utils/circuit_registration` and will need a rebase/port when landing.

Made with [Cursor](https://cursor.com)


app/endpoints/circuit_registration.py
```diff
@@ -0,0 +1,378 @@
...
+    except (ValueError, FileNotFoundError) as e:
+        raise HTTPException(status_code=422, detail=str(e)) from e
+
+    if registered is None:
+        raise HTTPException(status_code=500, detail="Circuit registration returned no entity")
+    return registered
+
+
+@router.post("/circuit/register")
+def register_circuit_endpoint(  # ruff: ignore[too-many-arguments, too-many-positional-arguments]
```
chr-pok, line 194: Just wondering, `dry_run` does not seem to be implemented in the endpoint. Is this intended? Would be nice to have for checking all dependencies before actually registering the entity.

jdcourcol: thx, I forgot about dry run.

james-isbister (approved):
Accepting for testing

chr-pok:
@jdcourcol, overall works nicely, but here some comments after I did some testing (to be addressed in separate PRs):

**Validate endpoint:**
- Seems like the validation endpoints triggers asset generation even when called standalone. I think this should be independent (unless used in the registration -> validation -> asset generation flow)
- Can the logs be made accessible in case validation fails? I suppose they are in the launch-system logs, but even the job ID is not returned. I got some of my test circuits disqualified but I don't know why.
- Maybe there should be an activity created which could keep the job ID (as we do in other tasks). Also, this would make it easier to keep track of running/completed validation tasks.
- The fixed resource allocation may not be sufficient for large circuits (to be tested).

**Asset-generation endpoint:**
- Same here: Can the logs be made accessible, especially in case the generation fails?
- Again, could create an activity to keep track of job ID and progress.
- Seems like generate_connectivity_plot_assets are skipped (because of include_visualization=False)? But the connectivity plots should always be in sync with the matrix.
- Looks like the `include_visualization` flag is causing the confusion: I think it should include only the automatically generated overview/sim designer images, but not the connectivity plots. The connectivity plots should always be (re-)generated whenever the matrix is (re-)extracted.
- And whenever not used standalone but as part of the registration -> validation -> asset generation flow, also the overview/sim designer images should be generated, unless uploaded by the user.
- Would be useful to further subdivide into two dedicated endpoints for (a) compression and (b) matrix extraction + connectivity plots since they are independent of each other and may require very different resources etc.
- Same here: The fixed allocation time (1h) will be too short, especially for the compression of large circuits.

**Registration endpoint:**
- There seems to be a wrong flow of actions: My registered circuit got “disqualified” after validation, so no assets should have been generated, but my circuit still has them.
- I think the problem originates in `_register_draft_from_uploads()` which already generated the additional assets even before validation (i.e., directly in the service, w/o using the launch-system), since:
  ```
  registered = register_circuit(
      ...
      include_visualization=not dry_run,
      skip_additional_assets=dry_run,
  )
  ```
  So at this stage, the additional assets and plots should be skipped unconditionally (to be generated only after validation on the launch-system).

chr-pok:
Found one more (while reviewing the customization PR):
-`_validate_id_mapping_files()` cannot simply remove a stale ID mapping file during validation. This would not change the actual circuit asset. I think, a validation error should be raised at this stage if the ID mapping is inconsistent.

PR #953, opened 2026-08-06
Circuit customization endpoint (stacked on registration)

jdcourcol:
## Summary
- Adds `POST /declared/circuit/customize` with staging/merge of parent circuit + uploaded overrides.
- Includes customization staging library, unit/integration tests, and docs (`circuit-customization-spec` + review follow-ups).
- Reuses the registration draft → async validate → assets pipeline from the base branch.

## Test plan
- [ ] `make test-file FILE=tests/app/endpoints/test_circuit_customization_endpoint.py`
- [ ] `make test-file FILE=tests/app/endpoints/test_circuit_customization_extra.py`
- [ ] `make test-file FILE=tests/obi_one/utils/circuit_customization/test_staging.py`
- [ ] Integration (when API available): `tests/integration/test_circuit_customization.py`
- [ ] Smoke: customize → draft circuit → validation → assets

## Notes
- Base: `circuit-registration` (merge that first, or retarget this PR to `main` after it lands).
- Diff vs base is customization-only (~13 files).

Made with [Cursor](https://cursor.com)


app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+        node_paths = _save_uploads(node_files, tmp)
+        try:
+            _validate_nodes(node_paths)
+        except ValueError as e:
+            errors.append(f"nodes: {e}")
+
+    return edge_paths, hoc_paths, mod_paths, node_paths, errors
+
+
+def _register_and_stage(
```
chr-pok: Naming is confusing. Should be better called `_stage_and_register`, i.e., following the order of the steps.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+
+    upload_sonata_circuit(db_client, registered, staged_dir)
+
+    generate_and_register_visualization_assets(
+        circuit=c,
+        config_path=merged_config,
+        output_root=staged_dir,
+        db_client=db_client,
+        registered=registered,
+    )
```
chr-pok: This whole part seems to be a duplication of the circuit registration functionality. In my understanding, we should reuse the existing functionality from the registration endpoint here, probably something like `_register_draft_from_uploads`.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+            edge_paths=edge_paths,
+            hoc_paths=hoc_paths,
+            mod_paths=mod_paths,
+            node_paths=node_paths,
+            node_sets_path=node_sets_path,
+            cfg_path=cfg_path,
+            pop_map=pop_map,
+        )
+
+    # 5. Trigger async validation task via launch-system
```
chr-pok:
```suggestion
    # 4. Trigger async validation task via launch-system
```

app/endpoints/circuit_helpers.py
chr-pok: Seems like this whole file contains mostly duplicate code. Should not be needed when re-using the existing circuit registration code.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+        )
+    except ValueError as e:
+        raise HTTPException(status_code=422, detail=str(e)) from e
+
+    # Compute metadata from the merged circuit
+    merged_config = staged_dir / "circuit_config.json"
+    if not merged_config.exists():
+        raise HTTPException(status_code=500, detail="Staged circuit is missing circuit_config.json")
+
+    try:
```
chr-pok: Seems to be a duplication of what we already have in circuit registration.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+            has_point_neurons,
+            has_electrical_cell_models,
+            has_spines,
+        ) = compute_circuit_metadata(name, merged_config)
+    except (OSError, ValueError, RuntimeError) as e:
+        raise HTTPException(
+            status_code=422, detail=f"Failed to compute circuit metadata: {e}"
+        ) from e
+
+    circuit_model = models.Circuit(
```
chr-pok: Again, duplication.

obi_one/utils/circuit_customization/staging.py
```diff
@@ -0,0 +1,346 @@
...
+    dest = circuit_dir / node_sets_path.name
+    if dest.exists() or dest.is_symlink():
+        dest.unlink()
+    shutil.copy2(node_sets_path, dest)
+    L.info("Copied nodeset file: %s", node_sets_path.name)
+
+    # If the current circuit_config doesn't reference node_sets_file, add the reference
+    try:
+        cfg = json.loads(circuit_config_path.read_text(encoding="utf-8"))
+        if not cfg.get("node_sets_file"):
```
chr-pok: I think we are missing the case when the reference exists already, but points to another filename/location, in which case we skip patching and keep pointing to the old file location.

chr-pok: Actually, not sure if we should even patch the config file at all. I see two ways in which a user can customize a node sets file:
1. The parent circuit already has a node sets and the user just replaces it. --> Must have same filename
2. The parent circuit does not have a node sets file, in which case the user needs to provide a new node sets file + a new circuit config which references it.

obi_one/utils/circuit_customization/staging.py
```diff
@@ -0,0 +1,346 @@
...
+    except Exception:  # ruff: ignore[blind-except]
+        L.warning("Could not patch circuit_config.json for nodeset file", exc_info=True)
+
+
+def _remove_stale_network_files(
+    circuit_dir: Path, circuit_config_path: Path, parent_config: dict
+) -> None:
+    """Unlink parent network files no longer referenced by the override circuit_config.
+
+    Only removes symlinks (files that came from the parent via EFS staging), never
```
chr-pok: Just wondering, can we be sure that files will always be staged as symlinks for public and private circuits?

obi_one/utils/circuit_customization/staging.py
```diff
@@ -0,0 +1,346 @@
...
+        for candidate in circuit_dir.rglob(stale_name):
+            if candidate.is_symlink():
+                candidate.unlink()
+                L.info("Removed stale network symlink: %s", candidate)
+
+
+def _network_file_names(cfg: dict) -> set[str]:
+    """Collect the bare filenames of all nodes, edges, and nodeset files in a config dict."""
+    names: set[str] = set()
+    for entry in cfg.get("networks", {}).get("nodes", []):
```
chr-pok: Why not using libsonata or SNAP for parsing the config file, wouldn't this be more robust w.r.t. changes in SONATA format? Same applies to many other locations in the code.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+    return paths
+
+
+def _validate_edge_population(path: Path, pop_name: str, pop: h5py.Group) -> None:
+    """Validate a single edge population group."""
+    for required in ("source_node_id", "target_node_id", "edge_type_id"):
+        if required not in pop:
+            msg = f"'{path.name}' population '{pop_name}': missing '{required}'"
+            raise EdgeValidationError(msg)
+    for key in pop:
```
chr-pok: I think this will skip the actual datasets under `/0`. We should probably check sub-groups recursively?

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+        ds = pop[key]
+        if hasattr(ds, "dtype") and ds.dtype.kind == "f":
+            data = ds[:]
+            if np.any(~np.isfinite(data)):
```
chr-pok: Actually, why is this even part of Level 1 validation? It reads the actual edge file content which can be huge.

obi_one/utils/circuit_customization/staging.py
```diff
@@ -0,0 +1,346 @@
...
+        mod_dir = _resolve_mod_dir(active_config, circuit_dir)
+        _copy_into(mechanism_overrides, mod_dir)
+
+    if node_sets_override:
+        _apply_node_sets_override(node_sets_override, circuit_dir, circuit_config_path)
+
+    # 5. Remove network files from the parent that the override config no longer references
+    if circuit_config_override:
+        _remove_stale_network_files(circuit_dir, circuit_config_path, parent_config)
+
```
chr-pok: Should validate the ID mapping file as well, and remove stale file if no longer consistent with customized circuit. This is currently done in `_validate_id_mapping_files()` in the async validation step, but should be done before actually registering the circuit.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+            )
+            errors.extend(
+                _run_cross_validations(hoc_paths, mod_paths, node_paths, parent_mech_names)
+            )
+
+        if errors:
+            raise HTTPException(status_code=422, detail={"validation_errors": errors})
+
+        # 3. Register, stage, and upload
+        registered = _register_and_stage(
```
chr-pok: Comment/naming is confusing and does not follow the order of steps.
```suggestion
        # 3. Stage, override, and register
        registered = _stage_and_register(
```

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+
+    for hoc_path in hoc_paths:
+        try:
+            check_mechanisms(hoc_path, available)  # ty:ignore[invalid-argument-type]
+        except ValueError as e:
+            raise HocValidationError(str(e)) from e
+
+
+def _collect_templates_from_group(
+    f: "h5py.File", group: "h5py.Group", pop_name: str, templates: set[str]
```
chr-pok: Usually we don't use this string style of type hints.
```suggestion
    f: h5py.File, group: h5py.Group, pop_name: str, templates: set[str]
```

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+) -> None:
+    """Extract model_template values from an HDF5 group into the templates set."""
+    if "model_template" not in group:
+        return
+    ds = group["model_template"]
+    if ds.dtype.kind in {"U", "S", "O"}:
```
chr-pok: Better use libsonata/SNAP to get model_template values no matter how they are stored.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+        if lib_path in f:
+            lib = f[lib_path]
+            templates.update(v.decode() if isinstance(v, bytes) else v for v in lib[:])
+
+
+def _validate_nodes_hoc_consistency(node_paths: list[Path], hoc_paths: list[Path]) -> None:
+    """Check consistency between uploaded nodes and HOC files.
+
+    Validates:
+    1. All model_templates in the nodes reference existing HOC files (uploaded or will be in parent)
```
chr-pok: I don't see where this check is happening (?)

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+    2. All uploaded HOC files are referenced by at least one node's model_template
```
chr-pok: I think this check should involve the parent's nodes as well. There could be replacements of existing hoc files but which are not referenced by the uploaded node, but by the existing ones in the parent. This could happen in case of multiple node populations where some are customized and some of them are kept from the parent I believe.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+    """
+    import h5py as h5  # ruff: ignore[import-outside-top-level]
```
chr-pok: Should not use inline imports, unless really needed.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+            msg = f"'{path.name}': all top-level keys must be strings"
+            raise NodeSetsValidationError(msg)
+        _validate_nodeset_expression(path, key, expr)
+
+
+def _validate_hoc_mechanisms(
+    hoc_paths: list[Path], mod_paths: list[Path], parent_mechanism_names: set[str] | None = None
+) -> None:
+    """Check that mechanisms used in HOC files are available (built-in or from provided MODs)."""
+    from obi_one.scientific.validations.emodels import (  # ruff: ignore[import-outside-top-level]
```
chr-pok: Should not use inline imports, unless really needed.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+                    pop = f["edges"][pop_name]
+                    _validate_edge_population(path, pop_name, pop)
+        except OSError as e:
+            msg = f"'{path.name}': not a valid HDF5 file: {e}"
+            raise EdgeValidationError(msg) from e
+
+
+def _validate_hoc(paths: list[Path]) -> None:
+    """Layer 1 validation for HOC files: check template structure (begintemplate/endtemplate)."""
+    from obi_one.scientific.validations.emodels import (  # ruff: ignore[import-outside-top-level]
```
chr-pok: Should not use inline imports, unless really needed.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+
+def _get_parent_mechanism_names(
+    db_client: entitysdk.client.Client, parent: models.Circuit
+) -> set[str]:
+    """Get MOD file stems from each population's resolved mechanisms_dir via SNAP."""
+    try:  # ruff: ignore[too-many-statements-in-try-clause]
+        from bluepysnap import Circuit as SnapCircuit  # ruff: ignore[import-outside-top-level]
+        from entitysdk.staging.circuit import (  # ruff: ignore[import-outside-top-level]
+            stage_circuit,
+        )
```
chr-pok: Should not use inline imports, unless really needed.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+            for pop_name in circuit.nodes.population_names:
+                mech_dir_str = circuit.nodes[pop_name].config.get("mechanisms_dir")
+                if not mech_dir_str:
+                    continue
+                mech_dir = Path(mech_dir_str)
+                if mech_dir.is_dir():
+                    names |= {p.stem for p in mech_dir.glob("*.mod")}
+            if names:
+                return names
+            # Fallback: scan staged directory for MOD files
```
chr-pok: May consider removing the fallback; will soon have "mechanisms_dir" in all re-extracted circuits.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+        lifecycle_status="draft",  # ty:ignore[invalid-argument-type]
+    )
+
+    try:
+        registered = db_client.register_entity(circuit_model)
+    except entitysdk.exception.EntitySDKError as e:
+        raise HTTPException(status_code=500, detail=f"Failed to register circuit: {e}") from e
+
+    # Create derivation link between the customized circuit and its parent
+    from entitysdk.types import DerivationType  # ruff: ignore[import-outside-top-level]
```
chr-pok: Should not use inline imports, unless really needed.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+
+    # 1. Fetch parent circuit
+    try:
+        parent = db_client.get_entity(entity_id=parent_circuit_id, entity_type=models.Circuit)
+    except entitysdk.exception.EntitySDKError as e:
+        raise HTTPException(
+            status_code=404, detail=f"Parent circuit {parent_circuit_id} not found: {e}"
+        ) from e
+
+    from entitysdk.types import EntityLifecycleStatus  # ruff: ignore[import-outside-top-level]
```
chr-pok: Should not use inline imports, unless really needed.

app/endpoints/circuit_customization.py
```diff
@@ -0,0 +1,728 @@
...
+            hoc_paths=hoc_paths,
+            mod_paths=mod_paths,
+            node_paths=node_paths,
+            node_sets_path=node_sets_path,
+            cfg_path=cfg_path,
+            pop_map=pop_map,
+        )
+
+    # 5. Trigger async validation task via launch-system
+    trigger_validation_task(
```
chr-pok: I think the async validation task needs an additional flag for en-/disabling additional assets generation if successful, to be used in the customization/registration -> validation -> asset generation flow.

But should not generate additional assets when used standalone just for validation.

jdcourcol: there is a generate_assets_on_success flag now. the circuit/{id}/validate defaults to false.

obi_one/utils/circuit_customization/staging.py
jdcourcol: it is using libsonata now

obi_one/utils/circuit_customization/staging.py
jdcourcol: done now

app/endpoints/circuit_customization.py
chr-pok: Potential security risk: `f.filename` could contain an abs. or rel. path, and may escape the target directory. Should use `Path(f.filename).name` instead (what we do already in circuit upload).

chr-pok (commented):
Thanks @jdcourcol, looks like all my comments have been addressed! Found one new (small) concern which I may have missed before.

app/endpoints/circuit_customization.py
jdcourcol: there is  _safe_upload_name(filename: str | None)  that does that now.

chr-pok (approved):
Thanks! Looks good to be merged so we can test once deployed in staging.

PR #961, opened 2026-08-14
Feat/mg circuit simplification task

mgeplf:


darshanmandge (approved):
Thank you for the change @mgeplf !

PR #962, opened 2026-08-14
Fix circuit config path resolution for neuronal manipulation properties

darshanmandge:
## Summary
- Fix path resolution for circuit neuronal manipulation properties endpoint: `circuit_config.json` is now staged with `copy_or_download` instead of `link_or_download`, so `$BASE_DIR` resolves to the temp directory rather than the S3 cache symlink. This fixes crashes when resolving non-default neuron sets on circuits whose config contains absolute paths.
- The above fix will prevent such errors:
 
image

- Deduplicate circuit staging logic into `_stage_circuit_for_neuron_set` helper.
- Update Ra (axial resistance) limits from `[10.0, 500.0]` to `[0.0, 1000.0]`.
- Fix broken markdown links in docstrings.
- Add tests for config copy strategy and multi-emodel intersection bounds.
- Update notebook cells for circuit manipulation properties endpoint.


obi_one/scientific/library/neuronal_manipulation_properties.py
```diff
@@ -107,29 +155,16 @@ def _resolve_neuron_set_and_get_templates(
...
-        config_path = temp_dir_path / "circuit_config.json"
-        snap_circuit = SnapCircuit(str(config_path))
-
-        # Stage all nodes.h5 files upfront so that neuron set resolution
-        # (which accesses property_names via bluepysnap) can open them.
-        for population in snap_circuit.nodes.population_names:
-            nodes_h5_resolved = Path(snap_circuit.nodes[population].h5_filepath).resolve()
-            nodes_relative = str(nodes_h5_resolved.relative_to(temp_dir_path))
-            _stage_file(db_client, circuit_id, asset_id, temp_dir_path, nodes_relative)
+        config_path, pop_to_nodes_path = _stage_circuit_for_neuron_set(
```
james-isbister: Could this be called '_stage_circuit_for_neuron_set_resolution'?

james-isbister (commented):
Looks good. Just one comment

obi_one/scientific/library/neuronal_manipulation_properties.py
darshanmandge: Done.

james-isbister (approved):

PR #963, opened 2026-08-19
fix: register SynapseParameterizationScanConfig in union and add camp…

bilalesi:
…aign properties


obi_one/scientific/tasks/synapse_parameterization/config.py
```diff
@@ -138,6 +138,14 @@ class Initialize(Block):
         },
     )
 
+    @property
```
james-isbister: Actually, the problem is that SynapseParameterizationScanConfig should have inherited from [InfoScanConfig](https://github.com/openbraininstitute/obi-one/blob/8c11d95fe986a36ee4e3f7cb02fa16fdd7a039cd/obi_one/scientific/library/info_scan_config/config.py) rather than ScanConfig

That will remove the need for these two properties

bilalesi: fixed

james-isbister (approved):

PR #966, opened 2026-08-21
EModel Optimisation Workflow

darshanmandge:
- Updated optimisation workflow


darshanmandge:
Hi @ilkilic @james-isbister,
This is ready for your initial review. I will add more tests to increase coverage once the final schema is ready. @bilalesi, you can also have a look at the FE implementation.

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/artifacts.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def write(self, coordinate_root: Path) -> None:
+        """Write the bundle below a coordinate root using only contract paths."""
+        _write_json(coordinate_root / self.params_path, self.params)
+        _write_json(coordinate_root / self.recipes_path, self.recipes)
+
+
+def _write_json(path: Path, value: Mapping[str, Any]) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(json.dumps(value, indent=4), encoding="utf-8")
```
eleftherioszisis: Use write_json from obi_one/utils/io.py

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/artifacts.py
```diff
@@ -0,0 +1,122 @@
...
+        raise ValueError(msg)
+
+    emodel = config.initialize.emodel
+    params = build_params_definition(
+        config,
+        normalized_ion_channel_models,
+        morphology_capabilities=morphology_capabilities,
+    )
+    recipes = build_optimization_recipe(emodel, mtype, morphology_filename)
+    _shared.update_pipeline_settings(
```
eleftherioszisis: Is it a common module? Name it common.py if that's the case.

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/config.py
```diff
@@ -0,0 +1,328 @@
...
+
+    contract_version: Literal["task2-config-v2"] = Field(
+        default=TASK2_CONFIG_CONTRACT_VERSION,
+        frozen=True,
+        title="Task 2 configuration contract version",
+        description="Compatibility version consumed by the launch-system optimization runner.",
+        json_schema_extra={SchemaKey.UI_ENABLED: False},
+    )
+
+    # --- Setup ---
```
eleftherioszisis:
```suggestion
```

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/config.py
```diff
@@ -0,0 +1,328 @@
...
+        title="Initialize",
+        description="E-model name and E-type entity used by the ``EModel_pipeline`` constructor.",
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.BLOCK_SINGLE,
+            SchemaKey.GROUP: InfoBlockGroup.SETUP_BLOCK_GROUP,
+            SchemaKey.GROUP_ORDER: 1,
+        },
+    )
+
+    # --- Inputs ---
```
eleftherioszisis:
```suggestion
```

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/config.py
```diff
@@ -0,0 +1,328 @@
...
+        ),
+        json_schema_extra={
+            SchemaKey.UI_ELEMENT: UIElement.BLOCK_DICTIONARY,
+            SchemaKey.GROUP: BlockGroup.INPUTS,
+            SchemaKey.GROUP_ORDER: 3,
+            SchemaKey.SINGULAR_NAME: "Custom Distance-Dependent Distribution",
+        },
+    )
+
+    # --- Settings ---
```
eleftherioszisis:
```suggestion
```

app/mappings.py
```diff
@@ -240,6 +240,18 @@
...
         ),
     ),
+    TaskType.emodel_optimization: TaskDefinition(
+        task_type=TaskType.emodel_optimization,
+        config_type=TaskConfigType.emodel_optimization__config,
+        activity_type=TaskActivityType.emodel_optimization__execution,
+        code=BuiltinCode(script=BuiltinScript.emodel_optimisation),
+        resources=ClusterResources(
+            instances=1,
+            instance_type="small",
```
ilkilic: Shouldn’t this use at least a medium/large instance, and perhaps a longer time limit?

eleftherioszisis:
I noticed the following issue while running the examples:

 **Preflight misses somatic on real SWC morphologies (e.g. notebook C060114A5)**

  I verified this on dm1_emodel_02_emodel_optimization (original PR branch) with the morphology from 02_emodel_optimization.ipynb. The failure is not introduced by the BluePyEModel
  refactor — the same preflight logic exists locally in morphology_preflight.py and fails identically.

  **Problem**

  _available_physical_sections() only inspects morphology.sections for SectionType.soma. For standard SWC files, MorphIO stores type-1 (soma) points in the separate morphology.soma
  object, not as somatic neurite sections. Preflight therefore reports no somatic capability even when the SWC clearly has soma points.

  That causes build_params_definition() to reject configs that use somatic (including default base_parameters["all"] and the notebook’s NaTg assignment in "somatic").

  **Repro (original branch)**
```
  cd obi-one
  git checkout dm1_emodel_02_emodel_optimization
  uv sync --extra emodel
  uv pip install -e .
  # 1) Inspect MorphIO on the notebook morphology (local copy of C060114A5)
  uv run python <<'PY'
  from pathlib import Path
  import morphio
  from obi_one.scientific.tasks.emodel_building.task2_emodel_optimization.morphology_preflight import preflight_morphology
  morph_path = Path("examples/data/morphologies/L5_TPC:A_MusMusculus_PrimarySomatosensory_C060114A5/C060114A5.swc")
  collection = morphio.Collection(str(morph_path.parent), extensions=[morph_path.suffix])
  m = collection.load(morph_path.stem, morphio.Option.nrn_order)
  print("morphology.soma n_points:", len(m.soma.points))
  print("morphology.soma.type:", m.soma.type)
  print("SectionType.soma sections:", sum(1 for s in m.sections if s.type == morphio.SectionType.soma))
  print("neurite section types:", {t.name: sum(1 for s in m.sections if s.type == t) for t in set(s.type for s in m.sections)})
  caps = preflight_morphology(morph_path, "replace_axon_with_taper")
  print("preflight available_physical_sections:", caps.available_physical_sections)
  PY
```
  **Expected output:**

  morphology.soma n_points: 21
  morphology.soma.type: SomaType.SOMA_CYLINDERS
  SectionType.soma sections: 0
  neurite section types: {'axon': 180, 'basal_dendrite': 74, 'apical_dendrite': 121}
  preflight available_physical_sections: ('basal', 'apical', 'axonal')   # no 'somatic'

  So: 21 soma points exist, but preflight says somatic is unavailable.

  Repro (params build / artifact step)

  Use a registered coordinate from the notebook (or any config with default base_parameters + "somatic" in mechanism_regions):

```
uv run python <<'PY'
import json
from pathlib import Path
from obi_one.scientific.tasks.emodel_building.task2_emodel_optimization.config import EModelOptimizationSingleConfig
from obi_one.scientific.tasks.emodel_building.task2_emodel_optimization.morphology_preflight import preflight_morphology
from obi_one.scientific.tasks.emodel_building.task2_emodel_optimization.parameter_builder import build_params_definition, normalize_ion_channel_model
from types import SimpleNamespace
coord_path = Path("obi-output/02_emodel_optimization/grid_scan/0/obi_one_coordinate.json")
config = EModelOptimizationSingleConfig.model_validate(json.loads(coord_path.read_text()))
morph_path = Path("examples/data/morphologies/L5_TPC:A_MusMusculus_PrimarySomatosensory_C060114A5/C060114A5.swc")
caps = preflight_morphology(morph_path, config.morphology_settings.axon_modifier)
model = SimpleNamespace(
  id="13c947c3-cb76-4a9a-91f4-146e95bd25f3", name="NaTg", nmodl_suffix="NaTg",
  is_stochastic=False, is_ljp_corrected=False, temperature_celsius=34,
  neuron_block=SimpleNamespace(range=[{"gNaTgbar": "S/cm2"}], global_=[]),
)
normalized = {"13c947c3-cb76-4a9a-91f4-146e95bd25f3": normalize_ion_channel_model(model, entity_id="13c947c3-cb76-4a9a-91f4-146e95bd25f3")}
build_params_definition(config, normalized, morphology_capabilities=caps)
  PY
```

  **Fails with:**

```
  ValueError: Configured region 'all' expands to ['apical', 'axonal', 'basal', 'somatic'],
```
  but the imported morphology has no source sections for ['somatic'].

  Same error when running full EModelOptimizationTask.execute() with the notebook coordinate — it fails after staging features/morphology/mechanisms, at _build_artifacts →
  build_params_definition.

  **Root cause in code**
```
  # morphology_preflight.py — _available_physical_sections()
  for section in morphology.sections:
      physical_name = _SECTION_TYPE_TO_PHYSICAL_NAME.get(section.type)
      ...
```
  This never reads morphology.soma. For C060114A5, MorphIO puts 21 type-1 points into morphology.soma (SOMA_CYLINDERS) and builds neurite sections only for axon/basal/apical.

  SWC topology note: the 21 soma points form a chain 1→2→…→21; axon (point 22) branches from point 1, not from point 21. That is valid SWC, but still results in zero SectionType.soma
  sections.

Is this expected behavior?

darshanmandge:
> I noticed the following issue while running the examples: ...

I will fix and check the config by running it locally with the notebook. I am also working on James's comments for the blocks after my discussion with him yesterday.

darshanmandge:
I have tested the branch locally with the notebook and also created a new block with the GUI specifications added in the docs.

tests/ui_schema/test_schema.py
```diff
@@ -177,16 +182,28 @@ def validate_scan_config_dependendent_block_components(block_schema, ref, form):
...
+            block_schemas = [{"$ref": block_ref}]
+        elif isinstance(additional_properties.get("properties"), dict):
+            block_schemas = [additional_properties]
+            direct_schema = True
+        else:
             msg = (
-                f"Validation error at {config_ref}: block_dictionary {key} must have 'oneOf'"
-                " or '$ref' in additionalProperties"
+                f"Validation error at {config_ref}: block_dictionary {key} must have 'oneOf', "
+                "'$ref', or an inline object schema in additionalProperties"
```
g-bar: This steamrolls over the spec and breaks the contract with the FE as to what a `block_dictionary` is. this is dangerous because it could break existing block dictionaries  for other workflows if they modify the schema.

In particular they must have a `oneOf`. So instead of changing the validation script, change the schema to comply with the existing spec,
if this is not possible create a new `ui_element` instead.

g-bar (changes requested):

tests/ui_schema/validate_block.py
```diff
@@ -79,12 +79,16 @@ def validate_type(schema: dict, ref: str) -> None:
...
+        except ValidationError:
+            continue
+        else:
+            return
 
-    except ValidationError:
-        msg = f"Validation error at {ref}: string_input param {param} failedto validate a string"
-        raise ValidationError(msg) from None
+    msg = f"Validation error at {ref}: string_input param {param} failed to validate a string"
+    raise ValidationError(msg) from None
```
g-bar: Same as above, this breaks the contract with the FE that a string_parem is a string not an array.

if you need this, creat a new ui_element

tests/ui_schema/validate_block.py
```diff
@@ -129,6 +133,9 @@ def determine_minimum_valid_numeric_value(schema: dict) -> float | int:
 def validate_numeric_single_and_list_types(
     schema: dict, param: str, ref: str, data_type: str, ui_element: str
 ) -> None:
+    if schema.get("type") == data_type:
+        return
```
g-bar: This skips validation, big NO NO.

tests/ui_schema/validate_block.py
```diff
@@ -152,6 +159,9 @@ def validate_numeric_single_and_list_types(
 
 
 def validate_float_param_sweep(schema: dict, param: str, ref: str) -> None:
+    if schema.get("type") == "number":
+        return
```
g-bar: Same here

g-bar:
@darshanmandge You cannot just change the validation script for a ui_element without previous discussion with the fe those form the basis of the contract between FE and `obi-one` for ui generation, without this the FE can break.

The path normally is, if you need a different schema you create a new `ui_element`, these can share design, but the schema for each `ui_element` cannot differ between uses.

tests/ui_schema/validate_block.py
```diff
@@ -657,24 +674,18 @@ def validate_model_selector_single(schema: dict, param: str, ref: str) -> None:
...
+    if boolean_branch is not None:
+        return
 
-    try:
-        validate(test_false, schema)
-    except ValidationError:
-        msg = f"Validation error at {ref}: boolean_input param {param} failed to validate False"
-        raise ValidationError(msg) from None
+    msg = f"Validation error at {ref}: boolean_input param {param} should include a boolean type"
+    raise ValidationError(msg)
```
g-bar: Same here, but I think you get the idea now ;)

tests/ui_schema/validate_block.py
```diff
@@ -897,7 +928,9 @@ def validate_block(schema: dict, ref: str) -> None:
     validate_string(schema, "description", ref)
 
     for param, param_schema in schema.get("properties", {}).items():
-        if param_schema.get(SchemaKey.UI_HIDDEN):
+        if param_schema.get(SchemaKey.UI_HIDDEN) or not param_schema.get(
+            SchemaKey.UI_ENABLED, True
+        ):
```
g-bar: Instead of this, ensure the element has the `hidden` in it.

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/registration.py
```diff
@@ -215,18 +202,49 @@ def register_output_entities(  # ruff: ignore[too-many-locals]
...
+    # inline the registration here and collect figure files recursively with rglob.
+    task_result = db_client.register_entity(
+        TaskResult(
+            name=f"EModel Optimization Result — {emodel_name}",
+            description=f"Optimisation + analysis + export for emodel '{emodel_name}'.",
+            authorized_public=authorized_public,
+            task_result_type=TaskResultType.emodel_optimization__result,
+        )
     )
+    if checkpoint_file is not None:
```
eleftherioszisis: Why checkpoint_file is optional?

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/registration.py
```diff
@@ -215,18 +202,49 @@ def register_output_entities(  # ruff: ignore[too-many-locals]
...
+            file_path=checkpoint_file,
+            file_content_type=ContentType.application_x_hdf5,
+            asset_label=AssetLabel.emodel_optimisation_checkpoint,
+        )
+    figure_files = {
+        p.relative_to(figures_dir): p
+        for p in sorted(figures_dir.rglob("*"))
+        if p.is_file()
+    }
+    if figure_files:
```
eleftherioszisis: Same here. Is this optional?

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/registration.py
```diff
@@ -215,18 +202,49 @@ def register_output_entities(  # ruff: ignore[too-many-locals]
...
+        db_client.upload_directory(
+            entity_id=task_result.id,
+            entity_type=TaskResult,
+            paths=figure_files,
+            name="analysis_figures",
+            label=AssetLabel.emodel_analysis_figures,
+            transfer_config=MultipartDirectoryUploadTransferConfig(),
+        )
+    if emodel_summary_file is not None:
```
eleftherioszisis: Same here. My understanding is that all these are needed to register a valid emodel.

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/task.py
```diff
@@ -210,6 +210,20 @@ def get_model_configuration(self, *args: Any, **kwargs: Any) -> Any:
...
+                convert_checkpoint,
+            )
+
+            checkpoint_dir = coord_root / "checkpoints"
+            for pkl_path in sorted(checkpoint_dir.rglob("*.pkl")):
+                h5_path = pkl_path.with_suffix(".h5")
+                if not h5_path.exists():
+                    L.info("Converting checkpoint %s → %s", pkl_path.name, h5_path.name)
+                    convert_checkpoint(str(pkl_path), str(h5_path))
+
```
eleftherioszisis: It would be better if this was added in setup_and_run_optimisation instead of duplicating this in two places.

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/registration.py
```diff
@@ -188,13 +176,12 @@ def register_output_entities(  # ruff: ignore[too-many-locals]
...
 
     # --- Collect file paths for helpers ---
-    # Checkpoints: BluePyEModel writes .pkl files by default (not HDF5).
-    # The entitysdk helper parameter is named hdf5_checkpoint_file but accepts
-    # any checkpoint format.
+    # Checkpoints: BluePyOpt writes .pkl files; task.py converts them to .h5
+    # via bluepyemodel.tools.checkpoint_hdf5.convert_checkpoint before registration.
     checkpoint_dir = coord_root / "checkpoints"
     checkpoint_file = None
     if checkpoint_dir.exists():
```
eleftherioszisis: Why is this optional?

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/registration.py
```diff
@@ -99,45 +98,34 @@ def parse_final_json(final_path: Path, emodel_name: str) -> dict:
...
-    # SONATA directory — needed by task3 for final export
     sonata_dir = coord_root / "export_emodels_sonata"
-    if sonata_dir.exists() and any(sonata_dir.rglob("*")):
-        db_client.upload_directory(
-            entity_id=task_result_id,  # ty:ignore[invalid-argument-type]
-            entity_type=TaskResult,
-            paths={p.relative_to(sonata_dir): p for p in sonata_dir.rglob("*") if p.is_file()},
-            name=AssetLabel.emodel_optimization_output,
-            label=AssetLabel.emodel_optimization_output,
+    if recipes_path.exists() or params_path.exists() or (
```
eleftherioszisis: Can these be optional? There are both needed, aren't they?

obi_one/scientific/tasks/emodel_building/task2_emodel_optimization/registration.py
```diff
@@ -99,45 +98,34 @@ def parse_final_json(final_path: Path, emodel_name: str) -> dict:
 
 def upload_optimization_assets(
     coord_root: Path,
-    db_client: entitysdk.Client,
+    _db_client: entitysdk.Client,
```
eleftherioszisis: If not needed it should not be passed.

tests/ui_schema/test_schema.py
```diff
@@ -176,19 +181,14 @@ def validate_scan_config_dependendent_block_components(block_schema, ref, form):
...
-        block_schemas = [{"$ref": block_ref}]
+    if schema.get("additionalProperties", {}).get("oneOf") is None:
+        msg = (
+            f"Validation error at {config_ref}: block_dictionary {key} must have 'oneOf'"
+            "in additionalProperties"
+        )
+        raise ValueError(msg)
 
-    for block_schema in block_schemas:
+    for block_schema in schema.get("additionalProperties", {}).get("oneOf"):
```
g-bar: @darshanmandge @ilkilic

In a previous commit ([diff](https://github.com/openbraininstitute/obi-one/pull/878/changes#diff-3c76af0be8e8883e184de95fb554bb0dc45291c79d46afa69612d35473242b5c)) the `block_dictionary` validator was changed so `additionalProperties` no longer needs a `oneOf`, it can nest a block directly instead.

That breaks the schema contract ([block_dictionary.md](https://github.com/openbraininstitute/obi-one/blob/main/docs/gui-definition-spec/components/block_dictionary/block_dictionary.md)). The FE reads `additionalProperties.oneOf` to get the list of block options to render, so it needs the `oneOf`.

The bigger issue: we want each `ui_element` to map to exactly one schema shape. The FE picks how to render based on the `ui_element`, so a single `ui_element` shouldn't have multiple possible shapes or internal branching. If a dictionary can have a single block type, the cleaner fix would be a new `ui_element` like `block_dictionary_single` that tells the FE not to look for a `oneOf`.

Anyway, this branch is never actually hit. In `MEModelSynapticModelPlacementScanConfig` every `block_dictionary` still produces a `oneOf`, which is why the FE didn't break. So I'm not sure what the intent was, and I've reverted the validator to the version that requires `oneOf`.

Does that sound right @ilkilic, or was there another reason for the change?

ilkilic: Yes, that sounds right. synapse_groups has only one block type but it is wrapped in a discriminated union so it still produces oneOf. No other reason for the change that I am aware of.

g-bar (approved):
The UI related changes are ready, not sure if the rest.

darshanmandge:
@eleftherioszisis, are you happy with the current state of the Optimisation Task for the launch system? The UI changes should not have affected the Task. The optimisation example notebook (generating the TaskConfig) works well and launches local optimisation correctly.

eleftherioszisis:
I have migrated the changes to the PCS script using the new function from task.py. The task executes optimisation and registers the entities successfully.

Could you please increase coverage. Other than that I believe it is good to go on my side.

darshanmandge:
> [codecov/patch](https://github.com/openbraininstitute/obi-one/pull/966/checks?check_run_id=107083722241)Successful in 1s — 90.31% of diff hit (target 82.09%)

How much you want the converage increase do you prefer? It is already above threshold.

eleftherioszisis:
> > [codecov/patch](https://github.com/openbraininstitute/obi-one/pull/966/checks?check_run_id=107083722241)Successful in 1s — 90.31% of diff hit (target 82.09%)
> 
> How much you want the converage increase do you prefer? It is already above threshold.

Ah my bad. I saw an earlier failure.

eleftherioszisis (approved):

